PR #15, opened 2017-03-03
schema length to max. 50

phips28:
I have a schema longer than 20 chars. 

Is there a reason why you limited it to 20?


timgit:
It was arbitrary.  Will merge this in

phips28:
Thx! =)

timgit:
I'll push this later today as 0.5.1.  I want to add some tests for this config setting

PR #22, opened 2017-03-24
added yarn lockfile

jordansexton:
[yarn](https://yarnpkg.com) is an npm client that allows versions of dependencies to be locked down and prevents breaks due to changes in dependencies (and most nefariously, dependencies of dependencies). Even using exact versioning doesn't fix the latter case, but yarn does, and is faster and easier than using `npm shrinkwrap`.

This commit adds a yarn.lock file (generated with `yarn install`), which allows this library to be cloned and installed with `yarn install --pure-lockfile` to guarantee that the dependencies match what's being used in tests. This is especially useful when installing PgBoss in production (which we are working on doing!)


timgit:
weird.  The last PR I did for 0.5.1 blew away the yarn lock file I had in 0.5.0.  Didn't realize it was missing until now.  Thanks!

jordansexton:
Oh, cool. I didn't check the history, sorry!

timgit:
btw, did you run into a dependency issue that prompted this?

jordansexton:
No, I was just building from source to poke around and saw there was no yarn lockfile.

timgit:
ok.  Just confirming it was a dev-time and not run-time concern with the published package

PR #23, opened 2017-03-24
accept connectionString as a config property

jordansexton:
It's currently possible to initialize with a PG connection string, or with configuration object including a database, user, password, and other params, but not both. When a string is provided, it converts to `{ connectionString: config }`; this provides no obvious way to set the other params through a config object.

This commit provides for initializing Boss with a connection string and also allowing `schema`, `poolSize`, and `uuid` to be set, e.g.

```javascript
new PgBoss({
    connectionString: 'postgres://postgres@localhost:5432/db',
    schema:           'boss',
    poolSize:         10
});
```

### DIFF
--- a/src/attorney.js
+++ b/src/attorney.js
@@ -25,6 +25,9 @@ function applyDatabaseConfig(config) {
   if(typeof config == 'string') {
     config = {connectionString: config};
   }
+  else if ('connectionString' in config) {
+    assert(typeof config.connectionString == 'string', 'configuration assert: connectionString must be a string');
+  }
   else {
     assert(config.database && config.user && 'password' in config,
       'configuration assert: not enough database settings to connect to PostgreSQL');


jordansexton:
I'm happy to write a test for this. I looked through the tests, I just wasn't really sure where to put it.

jordansexton:
This is especially useful for running PgBoss on a platform like Heroku, which injects a `DATABASE_URL` env var rather than each component. We don't use Heroku currently, but it is popular, and we use Convox with RDS and it's the same.

timgit:
Tests are always appreciated. The best home for that atm would be in configTest. I have a connectionString test that needs to be moved there methinks

timgit:
Closing this as it will be bundled into 2.0 here pretty soon.  You can see a preview in the 2.0.0-beta3 release.

PR #32, opened 2017-05-09
Next job with LIMIT teamSize (Performance Boost)

phips28:
changed behaviour of teamSize > 1: 
* do not start multiple workers, just one!
* fetch jobs with a `LIMIT teamSize` to get multiple jobs with one query

Pros:
+ saves connections to db
+ less queries
+ less db work (-CPU)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pg-boss",
-  "version": "1.1.0",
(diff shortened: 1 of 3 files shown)


timgit:
Closing this as it's been manually merged into 2.0.  Thx

PR #34, opened 2017-08-24
Add TypeScript type definitions

druska:
Thanks for the awesome queueing framework! I've added TypeScript type definitions. If you don't want to be responsible for maintaining this, we can add them to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) instead. Integrated is preferable for TypeScript users.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,5 +52,6 @@
   "bugs": {
     "url": "https://github.com/timgit/pg-boss/issues"
   },
-  "homepage": "https://github.com/timgit/pg-boss#readme"
+  "homepage": "https://github.com/timgit/pg-boss#readme",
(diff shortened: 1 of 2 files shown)


timgit:
@druska, thanks for the PR.  I've yet to use this in a Typescript project yet, but I can see how handy this would be.  Did you hand-create this, or did you use a tool to auto-gen most of it?  If I were to take over maintenance, I'd like to automate it as much as possible.  That is, something I could use outside of a full-blown rewrite in TS. 😱

druska:
@timgit if the project is written in plain JavaScript, it's something you have to maintain yourself unfortunately.

PR #36, opened 2017-09-05
Fix TypeScript in NPM

druska:
I realize the `./src/types.d.ts` file was stripped when published to NPM. I believe this will fix it, but I don't know your deploy method.

Thanks!


timgit:
well, that's embarassing, lol.  Yeah, /src is in my .npmignore. 😁

druska:
Works great now, thanks 👍

PR #38, opened 2017-10-16
Respect custom schema along with connectionString

VitalyOstanin:
Custom schema (and poolSize) ignored when connectionString is used.


timgit:
Thanks for the PR

Issue #29, opened 2017-05-08
No retry for failed jobs


phips28:
I am using `done(error)` if an error occurs. 
retrylimit = 2
But it seems to ignore the retry settings for a failed job.

Is this on purpose?

timgit:
Yes, this is intentional and also subjective.  I originally designed retries only for the expiration use case and not for explicit failures.  This could be changed, though.   Can I add your name to the supporters column?

phips28:
ok got it.

We are also using SQS for job queing, and they retry both, expired and failed. 
And we have jobs where the initial attempt fails and a second is successfully.
Errors could be: database connection lost, network timeout or API failures to 3rd parties, ...

So I would say also retry the failed once. Would that be ok for you?

ofc, where is this supporter column? 🙂

elmigranto:
So does this mean we should implement retry logic on client-side? Say I have some webhooks to notify, and want to try each notification up to 3 times with 5 seconds interval. How would I go about that, keep stuff in memory and repeat HTTP requests on my end, and either `done()` or `fail()` after 15 seconds and 3 attempts? (Because judging from source, `fail()`ed job won't be picked up again?)

timgit:
Hey there @elmigranto.  Yes, until I implement this enhancement, you will need to handle this using other techniques.  I wouldn't recommend using in-memory state unless you're ok with losing that job. 

One way in which you could build this using the existing api would be to add a subscription via [`onFail()`](https://github.com/timgit/pg-boss/blob/master/docs/usage.md#onfailname--options-handler) and just republish the job after incrementing a retry counter in the payload.  Then, on each subsequent failure, you can check the retry count and decide if you want to re-publish or not.  You can use the [`startIn`](https://github.com/timgit/pg-boss/blob/master/docs/configuration.md#delayed-jobs) publish option for your delay like the following.

```js
{ startIn : '15 seconds' }
```

terehov:
+1
Yep, definitely something we would need :-) Any updates here?

timgit:
This has become more annoying in my own projects lately as my orchestrations have grown, and so has the amount of expire, fail and complete handlers.  This is going to be quite disruptive, no doubt.

Right now, my thoughts on the new api are 
consolidate api to `onComplete()` only, which has `request` and `response` as usual, with added an `result` prop or something like that with `state` (success or failed), then expired could just be a "reason" for the failure.

remove: `onFail()` and `onExpire()` subs

[...]

PR #40, opened 2017-12-01
Retry for failed jobs (#29)

chrisjensen:
This PR adds support for automatic retry of failed jobs (#29)  if the error object returned via `done(err)` has a `shouldRetry` property that is truthy.
(allowing the job to differentiate between a soft fail and a hard fail)

The jobs are retried using exponential backoff. The delay until the next retry is stored in a new column `retryIn`.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -9,6 +9,7 @@ pg-boss can be customized using configuration options when an instance is create
     - [Database options](#database-options)
     - [Job fetch options](#job-fetch-options)
     - [Job expiration options](#job-expiration-options)
+    - [Job retry options](#job-retry-options)
     - [Job archive options](#job-archive-options)
 - [Publish Options](#publish-options)
     - [Delayed jobs](#delayed-jobs)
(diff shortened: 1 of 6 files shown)


timgit:
👍 Thanks for this!  Reviewing...

timgit:
I owe you a response on this.  :)  I don't think this PR is compatible with what I'm wanting to do with #29. My plans for failure handling are to consolidate it with expiration so expiration can simply become a failure type.  

Also, I have a philosophical disagreement on forcing the the job handler (or `fail()` api) to control this via something like a `shouldRetry` property on an error object.  My preference for configuring things like retries, expiration, throttling, etc. is to use the `publish()` api with options so no downstream logic is required in handlers.

However, there's also a good chance I'm just not considering your specific use case as well.  😁

chrisjensen:
Sorry for the slow reply over Christmas period. The reason I'm using a flag on the exception is we have different exceptional circumstances that come up, some should retry, some should definitely not, so a flag on publish isn't nuanced enough for our needs.
I opted for shouldRetry because it's least breaking - in existing implementations the flag would be missing so no behaviour change.
(It also means the code handling the exception doesn't need to know how to do job re-queueing so it's easier to separate concerns)

Happy to revise this patch for other ways to resolve this that allow the type of failure to determine if the job is retried.

timgit:
Chris, no worries.  😁 

If the retry itself is conditional, I'm persuaded that it should be handled either in the handler itself (just do another publish() with `job.name` and `job.data`, for example), or via `onFail()`/`onExpire()` if the publish logic needs to be kept out of the handler for logic consolidation purposes.  I have a couple use cases like this, and typically colocate a `onFail()` and `publish()` call at the origin/orchestrator.

timgit:
Chris, I just published a beta for 3.0 that addresses retry on fail.  If you have a chance and you're still keen on this, please give it a try and let me know your thoughts.

Thanks!

chrisjensen:
Thanks Tim, sorry I didn't follow up. We're under the pump a bit at the moment, but hopefully in a few weeks I may have time to come back and check this out.
For the moment we went along the lines of your suggestion: we've wrapped our jobs in a handler that will fail the job if its a hard fail, but let it expire if it's a soft failure so that the expire retry regime takes over.

Issue #45, opened 2018-01-01
Allow the job handler of subscribe() to process all fetched jobs at once


haoliangyu:
Currently the job handler function of `subscribe()` is called for each fetched job. It would be sweet if there is an option to allow the handler to receive an array of all fetched jobs. In my use case, doing a batch processing with all fetched jobs could help to significantly reduce system workload.

``` javascript
boss
  .subscribe(
    'my-queue', 
    { batchSize: 50 }, 
    (jobs) => batchProcess(jobs)
  );
```

This is the expected usage. Though it could be achieved by periodically calling `fetch()`, a native support is better (and implementing it looks simple after reading the source code).

timgit:
I agree.  Currently, the docs use `teamSize` for the behavior to assume a single job per invocation, but since internally it currently supports both `teamSize` and `batchSize`, it would be easy to just split these options apart and have them behave differently, such as in your your example.

haoliangyu:
If that's the decision, I would be happy to write a PR for it.

timgit:
Sounds great!  Thanks

PR #46, opened 2018-01-06
use batchSize for batch subscription

haoliangyu:
This is the implementation of batch subscription (PR #45)

@timgit Please review the code and I will update the documentation once the code is approved.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -122,7 +122,7 @@ If the CREATE privilege is not granted (so sad), you can still use the static fu
 
 All job monitoring will be stopped and all subscriptions on this instance will be removed. Basically, it's the opposite of `start()`. Even though `start()` may create new database objects during initialization, `stop()` will never remove anything from the database.  
 
-**If you need to uninstall pg-boss from a database, just run the following command.** 
+**If you need to uninstall pg-boss from a database, just run the following command.**
 
 ```sql
 DROP SCHEMA $1 CASCADE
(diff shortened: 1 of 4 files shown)


timgit:
Thanks for this. I think the tests are failing because teamSize or batchSize are both optional.  But I think we're on the same page in regards to intent.  Also, I'm hesitant to do batch completion on them as well.  Do you have a strong opinion on that?

haoliangyu:
As for the batch completion, it does have the risk of not completing all jobs because some rows are locked for other operations, as mentioned in the documentation. Is that your concern? I personally don't have a strong opinion or answer on how to do it. Maybe just not provide the complete function and instruct the user to complete jobs manually?

Also, I do have a concern about updating the meaning of `teamSize` and `batchSize` as they are the same in the current version. In my use case, I have been using `batchSize` but many other people may be using `teamSize`. Therefore this PR should be part of the next major update because of the breaking change. So it is a right time to do it? I am fine with deferring the PR. BTW, my initial thought was to introduce a flag like `batchProcessing=true/false`.

timgit:
Yeah, the [subscribe config docs](https://github.com/timgit/pg-boss/blob/master/docs/configuration.md#subscribe-options) show them both.  I forgot that I was so verbose there, lol.  Ok, yeah, that's unfortunate for those who are shy of majors.  👍 

I have a list of things that merit a major, regardless.  Seems like a good time here in a month or 2.

haoliangyu:
@timgit 

The code is updated to 
* to use `teamSize` by default unless `batchSize`is specified, which satisfies current tests
* add an example at the documentation

PR #61, opened 2018-02-21
2.5

timgit:


timgit:
[whatevs](https://trello.com/c/88lE3REA/1-whatevs)

phips28:
@timgit 2.5.0 broke prev version compatibility. :(
if you have breaking changes, you should always upgrade to a major version 3.0.0!!

Now I had to fix my version to "pg-boss": "2.4.4"

phips28:
after some investigation, I found it was an error in the migrate.

`...start()`

1) it created the `archive` table, and ALTER to add archivedOn column
2) it failed to update `version` to 7

at next start, it tried to migrate again, because version is still 6 in database, but archive tables was already created.
1) CREATE IF NOT EXISTS works, but the ALTER add archivedOn throws an error:
`column "archivedon" of relation "archive" already exists`

now I had to resolve this by hand to set the version to 7.

Solution: do the migration `commands` within a tx (BEGIN & COMMIT)

```
BEGIN;
CREATE TABLE
ALTER TABLE
UPDATE version
COMMIT;
```

It would rollback if there is any error until COMMIT. Therefore db is clean at the next server start and it can try again to migrate.
If you need more info, let me know ;)

timgit:
Phil, thanks for posting. I have migration tests that should have caught this. I’ll figure this out and write a test to repro.

PR #66, opened 2018-03-06
Make JobDoneCallback params optional

robgraeber:
Verified this works locally. This basically lets you call `job.done()` rather than `job.done(null, null)`  in typescript.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -72,8 +72,7 @@ declare namespace PgBoss {
   }
 
   interface JobDoneCallback<T> {
-    (err: Error): void;
-    (err: null | undefined, data: T): void;
+    (err?: Error | null, data?: T): void;
   }
 
   interface Job<T = object> {


types.d.ts
```diff
@@ -73,7 +73,7 @@ declare namespace PgBoss {
 
   interface JobDoneCallback<T> {
     (err: Error): void;
-    (err: null | undefined, data: T): void;
+    (err?: null | undefined, data?: T): void;
```
timgit: @bradleyayers , I'll defer to you on this since I think you may have more of a vested interest in this than I.

bradleyayers: This would let pass a null or undefined error without data, is that supported? e.g.

```ts
done(null);
```

Another option is to add an option that takes no arguments:

```ts
interface JobDoneCallback<T> {
  (): void;
  (err: Error): void;
  (err: null | undefined, data: T): void;
}
```

timgit: It won't throw, but there would no point in passing a null.

bradleyayers: Alright, well if you're happy for that to be API then @robgraeber's PR looks good!

robgraeber: @bradleyayers  Hmm thanks for pointing that out. I think usually we want less overloading than more. I noticed we could also condense this further to 1 declaration and get rid of the `undefined` type for optional parameters:
```typescript
interface JobDoneCallback<T> {
  (err?: Error | null, data?: T): void;
}
```

robgraeber: Cool updated.

Issue #70, opened 2018-04-07
Transactional migrations


timgit:
@phips28, I'm creating this issue to track what you reported in #61.  Since pg-boss uses pg's pool.query, implementing this change may require special attention from all users that pass in an existing database connection since it may or may not support transactions.

From https://node-postgres.com/features/transactions:
![image](https://user-images.githubusercontent.com/656974/38449638-0abfc7fa-39d7-11e8-8983-aa534a455fe3.png)

Another altenative I could try is squashing all migration scripts into a single multiple command statement.  The docs don't seem to encourage that, though.  There's probably other options as well.  

My assumption at this point is that even if callers import a db prop during construction, the majority of them will finally resolve to the same pg driver in node.  However, if they're highly abstracted, like strongloop's loopback connector, they may never have that level of control over the driver.

phips28:
I created a PR for that:

https://github.com/timgit/pg-boss/pull/73

PR #73, opened 2018-04-07
Transactional migration (#70)

phips28:
Issue: https://github.com/timgit/pg-boss/issues/70

Migration: use pg transaction to avoid inconsistency. In case one query fails, or the node process restarts during the query is running, it will automatically rollback. 

Param `poolSize`: also use `max` as valid config parameter (its the underlying pg-pool param) 

Tests: changed timeouts to work with my remote database; unfortunately the `speedTest.js` doesnt work with my remote database 👎  but works on travis 👍


timgit:
@phips28 Thanks for this!  🙇 

btw, I downgraded bluebird to a dev dependency, which actually introduces a use case that Travis didn't catch since it would only show up with the --production arg or NODE_ENV var.

Also, all the commands don't need an extra semicolon, do they?  `join(';')` should take care of that, right?

phips28:
> Also, all the commands don't need an extra semicolon, do they? join(';') should take care of that, right?

Yes thats true, I first added the `;` in every query, but then I realized that the migration.js would need the `;` too at the end adn I was too lazy to edit all. So I used the join(';').

I removed it

src/contractor.js
```diff
@@ -2,6 +2,7 @@ const assert = require('assert');
 const plans = require('./plans');
 const migrations = require('./migrations');
 const schemaVersion = require('../version.json').schema;
+const Promise = require('bluebird');
```
timgit: Remove this

src/contractor.js
```diff
@@ -46,10 +47,13 @@ class Contractor {
...
 
   create(){
-    let promises = plans.create(this.config.schema).map(command => () => this.db.executeSql(command));
-
-    return this.promiseEach(promises)
-      .then(() => this.db.executeSql(plans.insertVersion(this.config.schema), [schemaVersion]));
+    // use transaction, in case one query fails, it will automatically rollback to avoid inconsistency
+    let queryInTransaction = `
+    BEGIN;
+    ${plans.create(this.config.schema).join(';')}
```
timgit: Add semicolon after the join() to terminate last command

src/contractor.js
```diff
@@ -46,10 +47,13 @@ class Contractor {
...
+    ${plans.insertVersion(this.config.schema).replace('$1', `'${schemaVersion}'`)}
```
timgit: Add semicolon after this as well to terminate command

src/contractor.js
```diff
@@ -82,23 +86,21 @@ class Contractor {
...
       return Promise.reject(new Error(errorMessage));
     }
-
-    let promises = migration.commands.map(command => () => this.db.executeSql(command));
-
-    return this.promiseEach(promises)
+    // use transaction, in case one query fails, it will automatically rollback to avoid inconsistency
+    let queryInTransaction = `
+    BEGIN;
+    ${migration.commands.join(';')}
```
timgit: Add semicolon after the join() to terminate last command

test/databaseTest.js
```diff
@@ -108,7 +104,47 @@ describe('database', function(){
...
     function countConnections(db) {
-      return db.executeSql('SELECT count(*) as connections FROM pg_stat_activity')
+      return db.executeSql('SELECT count(*) as connections FROM pg_stat_activity WHERE application_name=$1',
+        [boss.db.config.application_name])
+        .then(result => parseFloat(result.rows[0].connections));
+    }
+
+  });
+
+  it('connection count does not exceed configured pool size with `max`', function(finished){
```
timgit: I'd prefer to move this into a config test (does poolSize == max) instead of duplicating the entire connection pooling test

phips28: added new tests, do you like it that way? should we remove the pooling test?

timgit: Thanks!  I'll go ahead and merge and take a stab at the migration failure rolllback test

PR #75, opened 2018-05-08
Retry failed jobs

timclipsham:
This PR adds support for automatic retry of failed jobs (#29) using the existing `retryLimit` mechanism. It works in a similar way to expiration: when enabled it will on a specified interval check for failed jobs that have `retryCount < retryLimit`, and set their state to `retry`.

It's backwards compatible (i.e. a non-breaking change, 2.6.x), that uses a new constructor configuration option `failedCheckInterval` to control how frequently to check. If it isn't specified, pgboss behaves as it did before (i.e. only retries expired jobs).

I realise you have larger plans to address this problem in a new major version, however I'm keen to see if we can find a way to merge an interim solution into 2.x.x. No feelings will be hurt if you're not comfortable with that – keen to hear your honest feedback 👍.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -11,6 +11,7 @@ pg-boss can be customized using configuration options when an instance is create
     - [Job fetch options](#job-fetch-options)
     - [Job expiration options](#job-expiration-options)
     - [Job archive options](#job-archive-options)
+    - [Job retry options](#job-retry-options)
 - [Publish Options](#publish-options)
     - [Delayed jobs](#delayed-jobs)
     - [Unique jobs](#unique-jobs)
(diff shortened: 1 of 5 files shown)


timgit:
Tim, thanks for the PR! 👍  

I really like the simpliciity of your approach here, an opt-in, behind the scenes monitoring technique. 

Ideally if pg-boss advertises "failure retries", I want that be a strong guarantee. There's a couple issues that give me pause with this, the primary one being a potential race condition between the failure monitor and the archiver. The other issue is how to deal with an inconsistent behavior based on the conditional contructor option.  For example, starting an instance multiple times with and without this setting, may drop jobs into the archive too soon.

However, this does get closer to the prize. How would you feel if I published it under a beta tag?

timclipsham:
Sorry for the slow reply. A beta tag sounds great.

I also noticed your 3.0 PR – I'm looking forward to what's coming next!

PR #100, opened 2019-01-09
Add RETURNING 1 to insertJob plan

ChrisYammine:
Hey :wave: first of all thanks for this awesome library.
Turns out some of the logic that determines whether or not to return an `id` on publish requires that INSERT have some output.

I'm using `typeorm`(https://github.com/typeorm/typeorm/) and while I've gotten everything else to work when using the typeorm connection, this small hurdle is blocking me.

I've confirmed that this works after editing the plan locally.

Think we can roll this in?

### DIFF
--- a/src/plans.js
+++ b/src/plans.js
@@ -334,7 +334,8 @@ function insertJob(schema) {
       $12
     )
     ON CONFLICT DO NOTHING
-  `;
+    RETURNING 1
+  `; // returning 1 here so we can actually return id on publish
 }
 
 function purge(schema) {


timgit:
Hey there!  Thanks for the PR.  The change looks innocent enough, and I may actually just switch it over to the inserted id instead of 1 while I'm at it.  By the way, with typeorm, are you saying that typeorm won't allow you run an insert statement against pg without a `returning`?  I didn't see any issues posted in their repo, and it does sound like quite an odd requirement since it's just executing arbitrary SQL.  Can you tell me a bit more about why this fixed your issue?

ChrisYammine:
:wave: Hey
So when using a typeorm's connection `query` fun to execute SQL it appears as though `INSERT`s with no `RETURNING` statement simply returns an empty array (even though Postgres spits out something like `INSERT oid count`)

I imagine Sequelize/other ORMs correctly bubble up the inserted row count and therefore make it possible to use `pg-boss` without specifying a `RETURNING` for job insert

PR #101, opened 2019-01-14
Document retry parameters

knan-nrk:
Retry parameters were slightly underdocumented. Fix that.


knan-nrk:
Sounds better, agreed.

timgit:
Thanks!

PR #103, opened 2019-01-18
Add static methods to exported class type

ChrisYammine:
Adding `getConstructionPlans` and `getMigrationPlans` to the exported class type so I can use it in my test helpers that are `.ts` :smile:


timgit:
Thanks

PR #141, opened 2020-02-01
Add missing JobMonitorOptions to ConstructorOptions

bhsiao123:
Thank you for this wonderful library!

Seems like `monitorStateIntervalSeconds` and `monitorStateIntervalMinutes` are missing as ConstructorOptions in `types.d.ts` even though the docs mention they are there: https://github.com/timgit/pg-boss/blob/master/docs/configuration.md.

Let me know if PR looks reasonable, if anything is missing, or perhaps if I misunderstood anything.


timgit:
Thanks!

PR #142, opened 2020-02-03
multi-master and maintenance dogfooding

### LINKED ISSUE #106: Non-continuous upkeep of jobs table
I am looking into integrating this awesome lib (thanks, Tim and everyone involved!) into yet another project, but have some concerns on a way I'm planning to setup table monitoring.

**The Problem**

No good way to select master server which would be running `pgboss.start()`. All our instances are the same and we would prefer to keep it that way. Obviously, running N-1 "extra" monitoring queries and everything related to that is not ideal, and coming up with "master-selecting" protocol and monitoring that makes me dizzy :)

**Solution (?)**

Have a periodic cronjob fire via internal HTTP call that's guaranteed to hit just one node, run `pgboss.start()` on it, wait for the thing to do its job and `pgboss.stop()` it after. 

**Concerns**

- Looking over the codebase and docs, this looks okay, but would be neat to get your opinion?
- Is there a better way to make sure old jobs are archived, and expirations are propogated other than `setTimeout(someMinutes, guessItIsProbablyDoneByNow)`? My understanding is that `await pgboss.start()` only takes care of schema?
- Other than that, my take is `pgboss.connect()` would take care of all the actual scheduling/processing which is good to go on all the nodes at the same time without any additional coordination?
- Some other APIs I am missing, any general tips on the topic?

<hr />

Once again, Tim, thanks for working on pg-boss and making it available to everyone
♥️ @timgit

### LINKED ISSUE #146: Readiness handling, exposing isReady
pg-boss rejects normal requests (such as `publish()` or `subscribe()`) if pgBoss isn't ready yet:
https://github.com/timgit/pg-boss/blob/0dbc2d2628d96df343b3f6f3e1cd4afe44bd9dd2/src/index.js#L59-L84

This is inconvenient during integration testing because we have to wait for several seconds before starting commands or else we don't have expected results.

I think it makes more sense to check that pg-boss is *started* and *await* if it is not ready yet.

Also, it would be useful to expose `isReady()` in pg-boss API so that it could be used in readiness checks logic.

### LINKED ISSUE #148: Decouple creation of tables with monitoring/archiving
Currently, `start()` is responsible for both creation of pg-boss's tables in the database, and monitoring/archiving jobs. When starting multiple actors using docker-compose (alongside a database), we cannot be sure which actor will hit the database first. So, I think it makes sense to make `connect()` to also create the tables if they are absent yet. Perhaps, via opt-in boolean parameter `connect(createTables)`.

### LINKED ISSUE #149: Default job retries
Hi,

First of all let me start by saying you did an awesome job building this. 

I have a feature request concerning job retries. Currently it seems that it is not possible to specify a default value for job retries through the constructor and it is only possible by passing them through the publish method. 

I believe it would be a great feature addition to be able to set the retries to a sensible default instead of defaulting it to off. 

Thanks

### LINKED ISSUE #151: Problems with onComplete() documentation (and behavior?)
https://github.com/timgit/pg-boss/blob/master/docs/usage.md#oncompletename--options-handler

1) `onComplete()` handler apparently doesn't make the *completion* job completed if it just normally exits. The code example in the doc contains a non-existent `finished()` function call which kind of indicates that: it only works if I replace `finished()` with `boss.complete(job.id)`

2) Instead of example
```
{
    "request": {
        "id": "26a608d0-79bf-11e8-8391-653981c16efd",
        "name": "onCompleteFtw",
        "data": {
            "token": "trivial"
        }
    },
    "response": {
        "message": "so verbose",
        "code": "1234"
    },
    "failed": false,
    "state": "completed",
    "createdOn": "2018-06-26T23:04:12.9392-05:00",
    "startedOn": "2018-06-26T23:04:12.945533-05:00",
    "completedOn": "2018-06-26T23:04:12.949092-05:00",
    "retryCount": 0
}
```

I think it would be clearer to show that the job passed into handler is actually:
```
{
    "id": ...
    "name": ...
    "data": {
	    "request": {
	        "id": "26a608d0-79bf-11e8-8391-653981c16efd",
	        "name": "onCompleteFtw",
	        "data": {
	            "token": "trivial"
	        }
	    },
	    "response": {
	        "message": "so verbose",
	        "code": "1234"
	    },
	    "failed": false,
	    "state": "completed",
	    "createdOn": "2018-06-26T23:04:12.9392-05:00",
	    "startedOn": "2018-06-26T23:04:12.945533-05:00",
	    "completedOn": "2018-06-26T23:04:12.949092-05:00",
	    "retryCount": 0
	}
}
```

timgit:
Resolves #106 
Resolves #146 
Resolves #148
Resolves #149 
Resolves #151


.editorconfig
```diff
@@ -0,0 +1,9 @@
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+[*.js]
+charset = utf-8
+indent_style = space
+indent_size = 2
```
elmigranto: No final new line 🤣

timgit: haha.  Yeah, I added standard linting to keep whitespace PRs to a minimum.

moltar:
🥳

PR #156, opened 2020-03-19
WIP: Convert project to TypeScript

moltar:
Getting this WIP PR in to just get started.

If anyone wants to help - feel free.


tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "outDir": "lib",
+    "allowJs": true,
+    "target": "es6",
```
Shinigami92: We could use `target: es2017` because thats the best fit to support at least node 8

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
+  "compilerOptions": {
+    "outDir": "lib",
+    "allowJs": true,
```
Shinigami92: Is this only temporary while wip?

tsconfig.json
```diff
@@ -0,0 +1,12 @@
...
+    "target": "es6",
+    "module": "CommonJS",
+    "esModuleInterop": true,
+  },
```
Shinigami92: We should add `"rootDir": "src", "sourceMap": false and "strict": true` within `compilerOptions`

Shinigami92:
ToDo
- [ ] boss.js
- [ ] index.js
- [ ] manager.js
- [ ] File `types.d.ts` should be removed

tsconfig.json
moltar: Sure we can change it, I just left any TS optimizations until the end.

tsconfig.json
moltar: Are you referring to `allowJs`? If so, then yes. It's for WIP so that TS compiler copies over `*.js` files into `lib` for testing.

tsconfig.json
moltar: Yeah, as I mentioned above, I'm going to tighten up TS config to be strict after a first pass.

moltar:
This is my first time converting a project, so I am not sure if this is the best strategy or not, but this is what I decided:

* I created a dependency graph in `docs/dependency-graph.svg`, and I am starting to convert from leaf nodes up to parent nodes. Not strictly, as sometimes it is easier to do 2 files at a time, but this is a general guideline.
* I have loosened up TS config to make the initial first pass easy to get everything to `.ts` files.
* Once all of the files are converted, I will replace it with a tighter config. I have a habit of using [tsconfigs](https://github.com/mightyiam/tsconfigs) in my projects and was planning on utilizing that too. But I am open for reproducing a config also.

moltar:
FYI I am currently working on `boss.js` and have touched on `attorney.js` also.

moltar:
@Shinigami92 let me know if you want to help on this today, as I am will be stepping away from this in about 5 minutes for today. If you'd like to pick this up, please let me know and I'll commit and push my WIP.

Shinigami92:
I think I will let you working on this and later on I will review it and help with suggestions

moltar:
* [ ] Fix all `TODO` items marked in the code

moltar:
@Shinigami92 do you have any thoughts on b6c700a?

Shinigami92:
I'll checkout the branch when you're done and review it then

You can build the code and look in the transpiled JavaScript code and see if it is different in any way than the original code

---

Also, but off-topic, why is `bluebird` used? I thought this was supposed to create `Promise` / `async/await` from legacy code. But is this really still necessary in this project today?
It's used in `manager.js` and `worker.js` we may use native Promise for this, maybe in another PR?

moltar:
> You can build the code and look in the transpiled JavaScript code and see if it is different in any way than the original code

Yeah, did that. No obvious differences. The major difference is that transpiled version uses `__awaiter` custom Promise implementation. But then I also tried to transpile to `es2017` which has native promise support and that didn't have any effect.

> Also, but off-topic, why is bluebird used?

It's used for promise utility methods, like `delay` and `map`.

> we may use native Promise for this, maybe in another PR?

Yes, my plan is to leave any implementation refactoring for later.

First, I just want to translate the code with minimal changes. Then we can discuss actual structural changes, if necessary.

I'd prefer to use the `p-*` family of modules, instead of `bluebird` also.

moltar:
@timgit:

The `fetch` in `manager` has three possible return values:

```ts
    return jobs.length === 0 ? null
      : jobs.length === 1 && !batchSize ? jobs[0]
        : jobs
```

* `Job` (single job)
* `Job[]` (array of jobs)
* `null`

Yet, in `watch` method, the `sendItBruh` function only handles `null` (via early return) and `Job[]` (array of jobs), but not a single `Job`.

Am I reading it wrong, or is it a bug?

Thanks!

timgit:
Great function name, amirite?  `sendItBruh()` is always passed an array of jobs since the batchSize argument is forced in fetch():

```js
 const workerConfig = {
      name,
      fetch: () => this.fetch(name, options.batchSize || options.teamSize || 1),
      onFetch: jobs => sendItBruh(jobs),
      onError,
      interval: options.newJobCheckInterval
    }
```

moltar:
Thanks @timgit!

moltar:
@Shinigami92 Having trouble porting the following idiom in a type-safe way:

https://github.com/timgit/pg-boss/blob/5931c7d9149fa07d7f0449b63ac24841b24b9e29/src/index.js#L62

Any tips?

Shinigami92:
Will take a look / checkou the repo

Shinigami92:
Ok, just weak type definitions for now, without going too deeply into the details

```ts
function promoteFunction (obj: Manager | Boss, func: Function) {
  this[func.name] = (...args: unknown[]) => {
    if (!this.isReady) return Promise.reject(notReadyErrorMessage)
    return func.apply(obj, args)
  }
}
```

---

`obj: Manager | Boss` cause this are the only two that are called in

https://github.com/ScaleLeap/pg-boss/blob/12d8208502d0c0cfa07834da05c4927c87a28564/src/index.ts#L42

and

https://github.com/ScaleLeap/pg-boss/blob/12d8208502d0c0cfa07834da05c4927c87a28564/src/index.ts#L48

---

`func: Function` cause in

https://github.com/ScaleLeap/pg-boss/blob/12d8208502d0c0cfa07834da05c4927c87a28564/src/boss.ts#L46

you already only have `Function` 
and in 

https://github.com/ScaleLeap/pg-boss/blob/12d8208502d0c0cfa07834da05c4927c87a28564/src/manager.ts#L38

you have to many different callables... so make it weak for now

---

`...args: unknown[]` cause we dont know any of the argument types here

moltar:
@Shinigami92 Thanks for looking into this.

Sorry, I was not clear enough.

I meant to ask how to actually re-implement this whole thing.

That function is used to hang methods from Boss and Manager classes on the main `PgBoss` class. But that doesn't actually make those methods visible on the `PgBoss` class.

If you were to do:

```ts
const pgBoss = new PgBoss({...})

pgBoss.fetch() // this would fail to type check
```

Shinigami92:
Why is converting to TypeScript currently a problem?
Shouldn't we outsource refactorings / reimplementations to a later PR?
Then we could tackle each task individually

I'm a little afraid that you want to do too much in one big PR

moltar:
> Why is converting to TypeScript currently a problem?

Because those methods will not be available on the `PgBoss` instance.

It will not be possible to actually use it in TypeScript, as no methods will be visible.

Shinigami92:
Ah now I see 🤔 
Maybe it's possible to add some `declare`s?

Shinigami92:
Okay.... Don't know if this is really working XD but...

1. Add to Boss.ts

```ts
export interface BossFunctions {
  expire: typeof Boss.prototype.expire
  archive: typeof Boss.prototype.archive
  purge: typeof Boss.prototype.purge
  countStates: typeof Boss.prototype.countStates
  getQueueNames: typeof Boss.prototype.getQueueNames
}
```

2. `class Boss extends EventEmitter implements BossFunctions`
3. `class PgBoss extends EventEmitter implements BossFunctions`
4. Add into PgPass (e.g. below constuctor)
```ts
  expire: typeof Boss.prototype.expire
  archive: typeof Boss.prototype.archive
  purge: typeof Boss.prototype.purge
  countStates: typeof Boss.prototype.countStates
  getQueueNames: typeof Boss.prototype.getQueueNames
```

5. profit $$$

image

Same for Manager...

timgit:
Closing this out for now due to staleness.  We can reopen later if desired

Issue #166, opened 2020-05-15
Getting job metadata like startafter on subscribe?


kevboh:
Hi! We're in the processing of instrumenting our pg-boss instance, and one of the metrics we're interested in is the span of time between a job's `startafter` column and the time it was actually dequeued and initiated. It looks like only id, name, and data are returned for subscribed jobs—is there any way to get other data out beyond manually selecting for it on `pgboss.job` once a job is dequeued?

timgit:
All of this data is in the payload of an onComplete() subscription.  Is that an option?

kevboh:
Unfortunately it's not, purely due to the way our tracing is configured—we need that info up-front so it can sent to our observation service alongside the rest of the "job has begun" event.

I've added this in a fork as it's something we needed pretty urgently, but if you think this is worthwhile I can PR either specifically this change or a more general config option can opt in to extra columns when fetching the next job.

timgit:
I like this idea, and definitely think it's worth a general config option. Can you post a link to your fork here if it's public?

kevboh:
Here's the only change in the fork: https://github.com/postlight/pg-boss/commit/1d53ea47dec635da9a2beda210d892b780f8b05b

Thinking about it over night, I think the API I'd prefer is either a single boolean `extendedSubscription` (or similar, I'm sure you'd have a better name) that opts in to the entire job row for `subscribe`, or a more fine-grained approach were you can pass in an array of column names you want, it makes sure they're valid, and then appends them to the query. I'd be happy to PR either approach.

kevboh:
I suppose it's also worth noting that we're doing this to be able to calculate `startedon - startafter`, which is something pg-boss itself could calculate and return in the API, but I think the other approaches are probably more useful for others.

kevboh:
Just to follow up—I can probably submit this PR this week. Just let me know how you'd prefer the API to look like.

timgit:
For now I think that it would be more simple to return all columns using a boolean config option added to fetch and subscribe. Maybe something like `includeMetadata`  What do you think?

kevboh:
Yeah, I’m into it. I hope to be able to PR that soon.

PR #168, opened 2020-05-20
feat: allow opt-in to full job rows on subscribe/fetch

kevboh:
As discussed in #166, this change allows a user to specify a boolean, `includeMetadata`, in both the `subscribe` options object and the `fetch` function. When `true`, those functions will return an extended version of the `Job` type that includes all columns in `pgboss.job`.

Includes tests and types.

### DIFF
--- a/src/manager.js
+++ b/src/manager.js
@@ -221,10 +221,10 @@ class Manager extends EventEmitter {
     return this.createJob(name, data, options, singletonOffset)
   }
 
-  async fetch (name, batchSize, includeMetadata) {
-    const values = Attorney.checkFetchArgs(name, batchSize, includeMetadata)
+  async fetch (name, batchSize, options = {}) {
+    const values = Attorney.checkFetchArgs(name, batchSize, options)
     const result = await this.db.executeSql(
-      this.nextJobCommand(includeMetadata || false),
+      this.nextJobCommand(options.includeMetadata || false),
       [values.name, batchSize || 1]
     )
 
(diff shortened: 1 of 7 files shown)


timgit:
This looks fabulous.  Thanks!  :100: The only change request I have is to put the `includeMetadata` in `fetch()` and `fetchCompleted()` inside an `options` object so it matches `subscribe` and makes it future proof for new options (or even to consolidate batchSize into it in a later release).

kevboh:
@timgit done. I considered overloading it such that you could pass only a second argument, `options`, and include `batchSize` in it to allow a deprecation path for batchSize outside options, but that felt a little presumptive (and messy).

src/manager.js
```diff
@@ -100,7 +100,7 @@ class Manager extends EventEmitter {
 
     const workerConfig = {
       name,
-      fetch: () => this.fetch(name, options.batchSize || options.teamSize || 1),
+      fetch: () => this.fetch(name, options.batchSize || options.teamSize || 1, options.includeMetadata || false),
```
timgit, line 103: this line needs the new options

kevboh: My bad, I'm so used to TypeScript catching this sort of thing for me!

timgit:
@kevboh This has been merged and published via `pg-boss@4.3.0-beta` so you can kick the tires.  If you're happy with how it's working (i'm not using this yet atm), let me know and I'll kill the beta tag.  Thanks!

kevboh:
@timgit seems to work on my end! Thank you.

PR #169, opened 2020-05-21
Add missing `all` property to MonitorStates interface

brianmcd:
As mentioned [here](https://github.com/timgit/pg-boss/blob/master/docs/usage.md#monitor-states), `MonitorStates` has an `all` property that's missing from the type definitions.


timgit:
Thanks!

PR #171, opened 2020-06-06
Update subscribe handler types to reflect removal of done arg and promise-returning handlers.

GriffinSchneider:
It looks like, since [3.0](https://github.com/timgit/pg-boss/blob/master/CHANGELOG.md#tada-300-tada), `SubscribeHandler`s no longer receive the `done` callback as a separate argument, and those handlers can optionally return a Promise rather than calling `done`. This updates the types to reflect those changes.


timgit:
Thanks!

Issue #89, opened 2018-09-15
ssl with certificates


goock:
Hi,
My SSL connection requires certificate at client side therefore simple true/false does not work for me.
This is my ssl parameter:
```
      ssl: {
        rejectUnauthorized : false,
        ca   : fs.readFileSync("./client-certs/server-ca.pem").toString(),
        key  : fs.readFileSync("./client-certs/client-key.pem").toString(),
        cert : fs.readFileSync("./client-certs/client-cert.pem").toString()
      },
```
I modify the ssl parameter as follow:
```
      ssl: typeof poolConfig.ssl === 'boolean' ? !!poolConfig.ssl : poolConfig.ssl
```
in ```pg-boss/lib/db.js.```

Would be possible to modify the official repo?

Thanks,
Marcin

asafh:
Hi @timgit, thanks for this.  
The above issue is resolved in runtime but is not reflected in the types.d.ts file.  

I wouldn't mind creating a pull request for this - are you ok with adding a dev-dependency on @types/pg and simply having the DatabaseOptions type be an the import of PoolConfig?

Cheers

timgit:
dev dependency is fine. Thanks

on2air:
hi @asafh / @timgit  - any progress on the type modification for ssl?  I'm hacking it with
 `ssl?: boolean|any`
 to get a quick fix, but hoping to avoid in future. thx

asafh:
I forgot about it but now that've reminded me, I've created the PR.
@timgit looks ok?
Cheers

timgit:
Yes. Thanks

timgit:
This is merged and published in 4.3.3

asafh:
Hi @timgit , I realize now the solution is slightly wrong. Since devDependencies of pg-boss would not be installed, a Typescript project using it would raise an issue when it reaches the import:
``` 
node_modules/pg-boss/types.d.ts(2,26): error TS7016: Could not find a declaration file for module 'pg'. '/home/node/some_app/node_modules/pg/lib/index.js' implicitly has an 'any' type.  
Try `npm install @types/pg` if it exists or add a new declaration (.d.ts) file containing `declare module 'pg';
````

I think the right approach is to change it to a normal dependency, is that ok or would you suggest otherwise?
[moved my comment from the PR to here.]

**Comment by timgit** [comment]
How about we punt to the any type for now as originally suggested?

```
ssl?: boolean|any
```

asafh:
[...]

PR #173, opened 2020-06-16
Importing types from @types/pg

asafh:
This PR includes the fix suggested in issue #89 to have the more correct typings for the ssl parameter (used in pg's PoolConfig).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "pg-boss",
-  "version": "4.2.0",
+  "version": "4.3.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


asafh:
Hi @timgit , I realize now the solution is slightly wrong. Since devDependencies of pg-boss would not be installed, a Typescript project using it would raise an issue when it reaches the import:
``` 
node_modules/pg-boss/types.d.ts(2,26): error TS7016: Could not find a declaration file for module 'pg'. '/home/node/some_app/node_modules/pg/lib/index.js' implicitly has an 'any' type.  
Try `npm install @types/pg` if it exists or add a new declaration (.d.ts) file containing `declare module 'pg';
````

I think the right approach is to change it to a normal dependency, is that ok or would you suggest otherwise?

cmanou:
@timgit Just noticed that this is  a breaking change for typescript as it removes the other Database options in the constructor ie: `schema` and `db.executeSql`

timgit:
This will be fixed in the next release.  There's a thread on this in #89

Issue #175, opened 2020-06-30
db parameter can not be used anymore in typescript code


mlegenhausen:
#173 introduced a bug where the `db` Parameter can not be specified anymore in the constructor. When you look in the `types.d.ts` file the `Db` interface is not used anywhere.

Following code in the `types.d.ts` solves this issue

```ts
type DatabaseOptions = PoolConfig | { db: Db };
```

timgit:
Included in 4.3.4

PR #176, opened 2020-07-01
Missing db parameter readded to DatabaseOptions type #175

mlegenhausen:
See #175


timgit:
Thanks!  I can go ahead and release another patch for this, but the fix is already included in the cron branch and released in pg-boss@5.0.0-beta2. I will be removing the beta on this next week

PR #187, opened 2020-08-19
Continue adding to team batch up to teamSize

chrisjensen:
Creates a queue of `teamSize` and resolves the promise to the worker as soon as a job completes so as to maintain the queue at capacity.
This uses p-queue to implement the queuing and signalling when a job is complete.

Adds a new option `teamMinimumFetch`. Once promises from a batch start resolving, it will wait until at least that many have resolved before fetching a new batch (default value: 1)

Addresses the issue raised in the [discussion of multiple subscriptions](https://github.com/timgit/pg-boss/blob/master/CHANGELOG.md#changes-3)

~~This works, but a number of the tests now throw unhandled rejection errors because of a race condition with `pgboss.stop()` in the tests. I can replicate these errors by adding some console.log statements into the `watch()` function on master without my changes so I think it is coming down to just a very tight deadline before pgboss shuts everything down.
I'm not sure there's much of a solution here except maybe adding a `nextTick()` inside pgboss.stop()~~

~~I'd appreciate a steer on the best way to resolve the race.~~

Turns out this wasn't a race, but was missing [this catch](https://github.com/timgit/pg-boss/blob/master/src/manager.js#L93), and so errors from the closed connection were not being caught in the tests, adding a catch for that has fixed the tests.


chrisjensen:
Here's a POC for the solution above (waiting for a teamMinimumFetch before returning to the worker loop.
It's broken a test, but I thought I'd check this is what you had in mind before continuing.

I haven't yet added the non-promise handling you suggested above. I'll do that next

chrisjensen:
Ok. I've fixed the error. Turns out it was simply a missing `.catch()` and have reworked it so non-promise handling is as expected. (ie the queue resolves the job immediately, and will fetch a new team every interval)

I added a timeout to the queue equal to the default expires, but I'm having second thoughts about that and thinking to just keep the behaviour for non-resolving promises consistent with before this patch (ie that promise never resolves, then that job is never replaced, effectively reducing the teamSize by 1)

src/manager.js
```diff
@@ -70,6 +71,10 @@ class Manager extends EventEmitter {
   async watch (name, options, callback) {
     options.newJobCheckInterval = options.newJobCheckInterval || this.config.newJobCheckInterval
 
+    const teamQueue = options.batchSize ? null : new PQueue({ concurrency: options.teamConcurrency || 1, timeout: 15 * 60 * 1000 })
+    const teamSize = options.teamSize || 1
+    const queueSize = () => teamQueue.size - teamQueue.pending
```
timgit, line 76: Thinking about this (for probably too long, too).  Following is an example I'm using to reason about the flow.

1. subscribe() with teamSize = 10 and concurrency = 2, teamMinimumFetch = undefined
2. First load starts 2 promises immediately
  * queue.size = 8
  * queue.pending = 2
  * queueSize() = 6
  * teamSize - queueSize() = 4
3. First promise is resolved, on('next') fires.
  * queue.size = 7
  * queue.pending = 2
  * queueSize() = 5
  * teamSize - queueSize() = 5
4. Fetch 5 and load into the queue
  * queue.size = 12
  * queue.pending = 2
  * queueSize() = 10
  * teamSize - queueSize() = 0

Based on this, I think this might load more jobs into the queue that we want.  What if we were to use a simple "if the queue is <= teamSize / 2 then fetch the difference" instead of a config setting?  

Also, can you tell me why you're subtracting the pending promises from the size and not adding them together to get the size?

timgit:
> Addresses the issue raised in the discussion of multiple subscriptions

This issue was just an unnecessary limitation I removed from the previous version regarding multiple subscriptions for the same queue in the same instance.

src/manager.js
chrisjensen: ah, this is a bad copy and paste from when it was inline in the previous version, it was: 
`numToFetch = options.teamSize - teamQueue.size - teamQueue.pending`

it should be teamQueue.size + teamQueue.pending

Fixing that should mean it'll maintain the queue at teamSize.

Would setting it to teamSize / 2 be desirable for all cases?
If there were folks that were running their teamConcurrency at greater than teamSize / 2, it would mean the queue would be spending some unintended idle time.

src/manager.js
```diff
@@ -70,6 +71,10 @@ class Manager extends EventEmitter {
   async watch (name, options, callback) {
     options.newJobCheckInterval = options.newJobCheckInterval || this.config.newJobCheckInterval
 
+    const teamQueue = options.batchSize ? null : new PQueue({ concurrency: options.teamConcurrency || 1, timeout: 15 * 60 * 1000 })
```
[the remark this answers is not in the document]
chrisjensen: I'm not following why we'd use `Promise.race()`?
The timeout setting in p-queue "finishes" the job after that timeout, triggering the `next` event and so replenishing the queue.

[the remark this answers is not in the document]
chrisjensen: Sorry, I'm following now. Promise.race() on a per job basis since publish can have different per job timeout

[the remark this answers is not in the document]
chrisjensen: I've updated the branch to use Promise.race based on `job.expiresin`

src/manager.js
timgit: Yeah, I agree that it shouldn't be allowed to stay under the teamConcurrency value.  When I first read the logic, I didn't realize the `off()` would relieve some pressure as these events could be ignored during the next job fetch and we'd catch up later if jobs were completing quickly.  I think it would probably be just fine the teamMinimumFetch config property for now.

chrisjensen: Ok, just want to check I'm following:
When you say leave out teamMinimumFetch and also remove the teamSize/2 we just have it return to the loop after at least 1 job returns?

timgit: Yes

src/manager.js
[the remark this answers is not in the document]
timgit: After reviewing this a bit more, this will require jumping over 1 last hurdle.  Jobs are retried on explicit failure or expiration from `active` state, but the end state of `expired` is only triggered by an `active` job being marked as such, which means we could most likely get away with kicking the promise from the PQueue in the instance, but not necessarily take action on the promise resolution (resolve or reject, whichever we choose). The maintenance interval also has the ability to be configured, which makes any attempt at "making the decision" probably a bit too high a price to pay in effort and/or length of time.

So, perhaps we just "magic string" a constant and look for that in the failure handler to bypass the explicit failure and let the maintenance handler catch it?  I'm open to other ideas you may have.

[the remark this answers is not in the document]
chrisjensen: I've added this in, so to confirm. If a job completes or fails, normal failure or complete handlers are called as expected.
If the timeout triggers, then the job is abandoned and the promise will fall out of the team queue creating more space, but the job will continue running.
There's a bit of a gray zone of what to do with the job if it fails or completes _after_ the timeout.
I've added a `.catch(noop)` so that the job can't raise an unhandled promise rejection if it fails after the timeout.

src/manager.js
chrisjensen: Ok, i've removed teamMinimumFetch and it will always fetch if the queue drops below the team size

chrisjensen:
Sorry it's taken me a very long time to get back to this.
I've made the updates discussed above if you're still open to the patch

chrisjensen:
Oh, i see the codebase has moved on quite a bit (unsurprisingly 😅)
I've opened a new pull request for an implementation against the latest

PR #230, opened 2021-03-03
Add installation instructions to README

fiznool:
I guess this is pretty obvious to most, but as a newcomer I wasn't sure whether the package name was `pg-boss` or `pgboss`.


timgit:
Thanks!

PR #232, opened 2021-03-05
Update types

fiznool:
A few types were missing, so I added them in.

Technically speaking, since the main pg-boss object extends from a node `EventEmitter` it also has access to `emit`, `once`, `removeAllListeners`, etc. But I figured they aren't really needed here.

Oh, and the `package.lock` file was updated when I ran `npm install` since its version didn't match up with that in the main `package.json`.


timgit:
thanks!

PR #248, opened 2021-05-31
Fix .stop() types

stnwk:
Currently this causes a typescript error:

image

This PR fixes that by allowing `.stop()` to be called without any argument, truly using fallback/default options.


timgit:
Thanks!

PR #249, opened 2021-06-01
create schema if not exists

mamannn:
I have to create the tables in public schema, the current method failed due to the already existing public schema

### DIFF
--- a/src/plans.js
+++ b/src/plans.js
@@ -115,7 +115,7 @@ function create (schema, version) {
 
 function createSchema (schema) {
   return `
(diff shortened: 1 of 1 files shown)


timgit:
This is the first request I've had to reuse an existing schema.  This has the caveat of making installation unpredictable, for example, on a name collision failure. Uninstall also becomes a manual process, and the schema will have to be reviewed in detail in order to make sure all created objects were removed.  These comments and step will be needed in the docs.

timgit:
Closes #252

PR #259, opened 2021-07-02
Graceful stop docs & improved types

fiznool:
Thanks for landing the graceful stop feature in v6! It's something that will really help us in production.

I couldn't find any docs on this, so I took a stab at adding them. I must admit that I wasn't sure about the finer details, e.g. how the promise resolution works for `stop()` in graceful mode, so I would very much appreciate your insight here if anything isn't correct.

I also made a slight tweak to the types to be more consistent with the rest of the package, namely making the fields optional directly in the type, rather than using the `Partial` utility type.

Thanks again for all your hard work on this lib, it's working wonderfully for us. ❤️

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -307,11 +307,17 @@ Additionally, all schema operations, both first-time provisioning and migrations
 
 One example of how this is useful would be including `start()` inside the bootstrapping of a pod in a ReplicaSet in Kubernetes. Being able to scale up your job processing using a container orchestration tool like k8s is becoming more and more popular, and pg-boss can be dropped into this system with no additional logic, fear, or special configuration.
 
-## `stop()`
+## `stop(options)`
+
+**Arguments**
+
+- `options`: optional object ([stop options](configuration.md#stop-options))
 
 **returns: Promise**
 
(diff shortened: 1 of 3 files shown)


docs/usage.md
```diff
@@ -307,12 +307,18 @@ Additionally, all schema operations, both first-time provisioning and migrations
...
+
+**Arguments**
+
+- `options`: optional object ([stop options](configuration.md#stop-options))
 
 **returns: Promise**
 
 All job monitoring will be stopped and all subscriptions on this instance will be removed. Basically, it's the opposite of `start()`. Even though `start()` may create new database objects during initialization, `stop()` will never remove anything from the database.
 
+By default, calling `stop()` without any arguments will gracefully wait for all workers to finish processing active jobs before closing the internal connection pool and stopping maintenance operations. This behaviour can be configured using the stop options object. In graceful stop mode, the promise returned by `stop()` will be resolved once all active jobs have finished, or will be rejected if the timeout is reached before jobs have finished processing.
```
timgit, line 320:
```suggestion
By default, calling `stop()` without any arguments will gracefully wait for all workers to finish processing active jobs before closing the internal connection pool and stopping maintenance operations. This behaviour can be configured using the stop options object. In graceful stop mode, the promise returned by `stop()` will still be resolved immediately.  If monitoring for the end of the stop is needed, add a listener to the `stopped` event.
```

timgit (changes requested):
Looks great. Thanks. I added a change request to the stop() docs

PR #272, opened 2021-09-19
Publish subscribe

tcoats:
This PR builds on #271
I've added a subscription table and subscribe, unsubscribe methods to populate.
The publish method takes an event name (new concept) and maps through the subscription table to queue names, adding a job to each queue.

Documentation included, tests added.

Feedback welcome.

#270


tcoats:
Thinking about this more - ideally it would be a transaction. Jobs are added to all queues or none.

ilijaNL:
What is the progress on this. Since this feature can be used to create event driven application and potentially replace rabbitmq I would to contribute if necessary.

timgit:
We're waiting on the package maintainer. ;).  My goal is to merge this into it's own feature branch so I can get more hands on experience with the API and settle on semantics.  It will be the most major change made since creation to semantics, so I don't want to make this change too quickly.  I still really appreciate the effort that went into this and still look forward to collaborating on it

ilijaNL:
> We're waiting on the package maintainer. ;). My goal is to merge this into it's own feature branch so I can get more hands on experience with the API and settle on semantics. It will be the most major change made since creation to semantics, so I don't want to make this change too quickly. I still really appreciate the effort that went into this and still look forward to collaborating on it

Okaay, let me know if you need some help with testing/developing

tcoats:
@ilijaNL Thank you for your interest! We've been using this code in production for a few months and it's working well for our use case. Because the subscribed queues are the same job queues behind the scenes all existing singleton and job processing flow control options work.

I hear you around stale job queues. Our preference is to keep the queues populated by events for all time as we are publishing business events - things like new product. It's important that those new products make it into their destination systems. However I can see that other events may have time sensitive information that is not important to receive later. The publish method has the same flow control options so you could publish events with built in expiry. That's configured on publish, not on the subscribing queue so it wouldn't let you have different flow control configured per subscribing queue.

ilijaNL:
@tcoats thanks for reply. Glad to hear it works perfectly for you! I understand that all options for publishing can be used, retry, signleton etc. However I wonder how you cleanup the queue table (aka removing the stale queue entries from the queue table)

tcoats:
@ilijaNL When you're talking about a stale entry are you thinking about an entry in the job table? This is where the expiry option will trigger the maintenance job to remove expired jobs. Or are you thinking about stale event subscriptions? I'm not 100% sure I understand what you'd like to achieve.

Multiple workers will naturally execute things in parallel... if you want the sequence of events preserved you'd need to have a single worker I think.

ilijaNL:
@tcoats stale event subscription

tcoats:
We consider subscriptions permanent, not temporary. For us the system (queue) is always interested in an event it's subscribed to. If it is no longer interested it can unsubscribe. You could trigger the unsubscribe via a scheduled job or some other orchestrated time based event.

Do you have an example of a workflow that involves a queue that is subscribed to an event and that subscription should become stale? I'd like to understand your use case so it could be considered or if there is a way to configure Boss that meets the same objectives.

ilijaNL:
> We consider subscriptions permanent, not temporary. For us the system (queue) is always interested in an event it's subscribed to. If it is no longer interested it can unsubscribe. You could trigger the unsubscribe via a scheduled job or some other orchestrated time based event.
> 
> Do you have an example of a workflow that involves a queue that is subscribed to an event and that subscription should become stale? I'd like to understand your use case so it could be considered or if there is a way to configure Boss that meets the same objectives.

Consider we have a service billing which subscribes to UserJoined event, after some business logic, the billing service doesnt care anymore about the UserJoined event and so should not be subscribed anymore. The problem arises when you are removing the worker for job with name `event.billing.user_joined`, however since the entry `event.billing.user_joined` still exists in the subscribes table, the event will be created for this queue. In the end you must remove the subscribe row manually. Consider having many subscribers and manage those. If I look at RabbitMQ, the have a expires time on an queue, which will be removed after some time if inactive.

tcoats:
@ilijaNL We have a queue per service, rather than per event. So if the requirement changed for a billing service we don't drop it's queue, we'd remove the subscription. Do you have a job queue per event type?

ilijaNL:
@tcoats  I am considering it yes, however I wonder if it is best practise. If you make a queue per service, you need somehow to route the events to correct handler? Could you elaborate on your implementation

tcoats:
@ilijaNL I believe the queue is gone if there are no more jobs on it so if you removed the subscription and had processed all jobs you’d be okay.

We have an [internal memory event bus](https://github.com/y5labs/seacreature/blob/master/lib/hub.js) for in process event handling. However a Map or switch statement would also work.

ilijaNL:
> @ilijaNL I believe the queue is gone if there are no more jobs on it so if you removed the subscription and had processed all jobs you’d be okay.
> 
> We have an [internal memory event bus](https://github.com/y5labs/seacreature/blob/master/lib/hub.js) for in process event handling. However a Map or switch statement would also work.

Thanks for the feedback.


@timgit What needs to be done to make this PR be merged?

timgit:
I need to merge this into a new tracking branch.  Since there are conflicts now with master, I'll just pick a slightly older commit as the base commit for this and then merge.  After a review and some feedback cycles, I expect there may be follow-on PRs against this new branch.  I can start publishing alpha-tagged releases as soon as the merge is completed if desired. I expect it may take a few weeks for it to land in master.

tcoats:
@timgit please let me know if there is anything I can do. Happy to re-work these changes too if we work out a better way to implement this feature.

ilijaNL:
> @timgit please let me know if there is anything I can do. Happy to re-work these changes too if we work out a better way to implement this feature.

I am also in. I think this setup can be good to remove rabbitmq from our system and switch to pg-boss to do pub subs as well

Issue #277, opened 2021-10-03
onError listener does not emit error name & stack


phips28:
While debugging the https://github.com/timgit/pg-boss/issues/274 issue, I found out that the `this.boss.on('error', (error) => {..` listener does not show the full error (message, stack).

Listener:
```javascript
this.boss.on('error', (error) => {
	console.error('PgBossQueue: on error', error);
});
```

Error Log:
```
PgBossQueue: on error {
  length: 123,
  name: 'error',
  severity: 'ERROR',
  code: '42601',
  detail: undefined,
  hint: undefined,
  position: undefined,
  internalPosition: undefined,
  internalQuery: undefined,
  where: undefined,
  schema: undefined,
  table: undefined,
  column: undefined,
  dataType: undefined,
  constraint: undefined,
  file: 'postgres.c',
  line: '1400',
  routine: 'exec_parse_message',
  queue: 'xxxx',
  worker: 'c8a14fd2-0e80-4190-82e5-8946fb56297e'
}
```

---

I then printed the error in the `onError` function in `manager.js`, and that showed the full error. 
The problem I see: an Error object has no iterator: `TypeError: Found non-callable @@iterator` (But this error is not visible with current spread usage in emit, only when you add it to the console.log(...error)) -> So I added message and stack directly.
```diff
    const onError = error => {
+      console.log('onError', error);

-      this.emit(events.error, { ...error, queue: name, worker: id })
+      this.emit(events.error, { ...error, message: error.message, stack: error.stack, queue: name, worker: id })
    }
```

PR #278, opened 2021-10-03
Fix: onError emit full error message and stack

phips28:
Fixes: https://github.com/timgit/pg-boss/issues/277

I don't know if it introduces any regressions? But I don't think so, as you basically tried to spread the error object, and therefore expected `message` & `stack` to be there.


timgit:
thanks!

PR #291, opened 2021-12-07
v7 - pub/sub + teamRefill

### LINKED ISSUE #270: Publish Subscribe support

timgit:
Closes #270


tcoats:
@timgit a short note to say - thank you for version 7. We've been using this in several projects and it's excellent. The pub sub functionality has been invaluable and gives us exactly what we want.

timgit:
Thanks for the note @tcoats. Glad it's working well

PR #295, opened 2021-12-21
Types: Extend core events.EventEmitter

jhermsmeier:
This extends the PgBoss class declaration with the core EventEmitter, to enable usage of other EventEmitter methods, such as `.removeListener()`, and also updates the `on()` and `off()` return values accordingly.

It does bring in an implicit peer-dependency on `@types/node` for TypeScript users though, which I'm not quite sure about how to deal with best. Another option could be adding it as a (peer-)dependency in the `package.json`, but that would impact non-TypeScript users, and TypeScript users would most likely already have the core types present anyway.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "pg-boss",
       "version": "7.0.1",
       "license": "MIT",
       "dependencies": {
@@ -16,6 +17,7 @@
         "uuid": "^8.3.2"
       },
       "devDependencies": {
+        "@types/node": "^17.0.2",
         "coveralls": "^3.1.0",
         "luxon": "^2.0.1",
         "mocha": "^9.0.1",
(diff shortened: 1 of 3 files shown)


timgit:
> It does bring in an implicit peer-dependency on @types/node for TypeScript users though, which I'm not quite sure about how to deal with best.

The risk is that the consumer wouldn't have already imported @types/node and does `npm i --production` (or NODE_ENV=production), so it would be missing?  If that's the case, I also don't know what the normally acceptable solution would be.

> Another option could be adding it as a (peer-)dependency in the package.json, but that would impact non-TypeScript users, 

By impact, you mean it just makes the install size increase?  I vote "this is fine" on that if so.

> and TypeScript users would most likely already have the core types present anyway.

This I think addresses your first point about risk. What's the percentage of typescript apps not importing types?

jhermsmeier:
> By impact, you mean it just makes the install size increase? I vote "this is fine" on that if so.

Yes, exactly – the presence of an unnecessary dependency.

> This I think addresses your first point about risk. What's the percentage of typescript apps not importing types?

I think it'd be the exception that there's a typescript app running on node that doesn't import the node core types.

> The risk is that the consumer wouldn't have already imported @types/node and does `npm i --production` (or NODE_ENV=production), so it would be missing? If that's the case, I also don't know what the normally acceptable solution would be.

Yup, but then, with typescript it wouldn't even build, so the missing types would immediately become apparent, I think.

timgit:
I think using devDependencies like you have in the PR is the best option for round 1.  I've searched around github for other TS npm packages and it seems appropriate from what I've read. For example, the got package had quite a thread going earlier in the year about this, but it ended up being something very specific to their package.

PR #299, opened 2022-01-14
Add a type to MonitorStates.queue

amc6:
This adds a more complete type to `MonitorStates.queue`. The top level `MonitorStates` type shares most of the fields in common with each entry in `MonitorStates.queue` So `MonitorStates` ends up both extending the new type `MonitorState` as well as having an extra field `queue` which is an object with string keys and `MonitorState` values.


timgit:
Thanks!

PR #328, opened 2022-04-15
Ability to notify the workers to skip waiting

ilijaNL:
Add a method which notifies the worker that something has been changed and therefore canceling or skipping the loopDelay.

This gives ability reduce the jobs overal latency.


ilijaNL:
Any possible review? @timgit

timgit:
This has been merged and deployed in version 7.3.0.  Thanks!

PR #329, opened 2022-06-12
Support singleton queue behavior for any singletonKey

adamhamlin:
This updates the `job_singleton_queue` index to drive the special singleton queue behavior for any singletonKey that _starts with_ `__pgboss__singleton_queue`--not just equal to it. In effect, this allows multiple singleton queues per job name (see discussion: https://github.com/timgit/pg-boss/discussions/285).

I included the new boolean option `useSingletonQueue` to prepend the passed singletonKey with `__pgboss__singleton_queue`. Otherwise, it has no effect.

This should not impact existing behavior UNLESS someone already had a singletonKey that is prepended with `__pgboss__singleton_queue` but not equal to `__pgboss__singleton_queue`. That seems highly unlikely.

FYI: I wasn't sure about the appropriate versioning and I had some issues running the CI tests locally. I'll add some comments about that.

### DIFF
--- a/test/multiMasterTest.js
+++ b/test/multiMasterTest.js
@@ -20,7 +20,7 @@ describe('multi-master', function () {
     try {
       await pMap(instances, i => i.start())
     } catch (err) {
-      assert(false)
+      assert(false, err.message)
     } finally {
       await pMap(instances, i => i.stop({ graceful: false }))
     }
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -547,6 +547,15 @@ Available in constructor as a default, or overridden in send.
 
   This can be used in conjunction with throttling explained below.
 
+  * **useSingletonQueue** boolean
+
+  When used in conjunction with singletonKey, only allows 1 job (within the same name) to be queued with the same singletonKey.
+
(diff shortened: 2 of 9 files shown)


test/multiMasterTest.js
```diff
@@ -20,7 +20,7 @@ describe('multi-master', function () {
     try {
       await pMap(instances, i => i.start())
     } catch (err) {
-      assert(false)
+      assert(false, err.message)
```
adamhamlin, line 23: Sort of commenting in a random spot, but couldn't get a successful run of the entire test suite with `npm test` locally (I could get successful test runs of individual test files using `npx mocha`). I think I was using a valid environment looking at the CI configurations: Postgres 9.5 and Node 14.

It always failed with the same cause (`sorry, too many clients already`), but the specific tests that failed were not consistent. But always something like this:
Screen Shot 2022-06-12 at 12 05 34 PM

What am I missing? FYI I was running against the official docker image `postgres:9.5`

timgit:
@adamhamlin I'm merging this into a new tracking branch to run tests locally. Thanks!

PR #341, opened 2022-08-02
Add transaction support on createJob, insert & fetch

klesgidis:
## Summary
Enables user to be able to pass a `db` interface that implements the `executeSql` function

With this changes users are now able to pass their own DB which enables them to add a job inside a transaction


timgit:
Thanks!

SokratisVidros:
@klesgidis ❤️ this is exactly what we were looking for!

fragkakis:
@klesgidis ❤️

PR #358, opened 2022-12-29
Add support for destroy option

### LINKED ISSUE #318: Feature Request: Close connection pool
Hey there! Really grateful that you created and are continuing to support this lib.

I noticed that `PgBoss.stop` does not seem to close the `pg.Pool` instance, even when using the internal construction (i.e. not BYODB). I believe any socket handles will eventually be cleaned up, but it does create some noise in our tests which report "leaked handles." 
```
Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   1 passed, 1 total
Time:        4.727 s
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

Furthermore, it also creates slowdown when shutting down our application as we need to wait for the pool resources to automatically clean up on the pool's idle timeout.

I was playing around, and it looks like the following addition alleviated the problem for me:
```diff
diff --git a/src/boss.js b/src/boss.js
index cb85eac..4116c9a 100644
--- a/src/boss.js
+++ b/src/boss.js
@@ -175,6 +175,10 @@ class Boss extends EventEmitter {
         await this.manager.offWork(queues.MONITOR_STATES)
       }
 
+      if (this.db.isOurs && this.db.opened) {
+        await this.db.close()
+      }
+
       this.stopped = true
     }
   }
```

Let me know if you would be open to including something like this - I'm unclear if there are other ramifications here to graceful shutdown.

abelsoares:
This PR adds support for `destroy` as an option for the `stop()` method.  
If the database connection is handled by `pb-boss`, this option will make it possible to close the `pg` connection pool.  

Closes #318

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -438,6 +438,9 @@ By default, calling `stop()` without any arguments will gracefully wait for all 
 
 * `options`: object
 
+  * `destroy`, bool
+    Default: `false`. If `true` and the database connection is managed by pg-boss, it will destroy the connection pool.
+
   * `graceful`, bool
 
     Default: `true`. If `true`, the PgBoss instance will wait for any workers that are currently processing jobs to finish, up to the specified timeout. During this period, new jobs will not be processed, but active jobs will be allowed to finish.
--- a/src/index.js
+++ b/src/index.js
@@ -126,7 +126,7 @@ class PgBoss extends EventEmitter {
       this.emit(events.stopped)
     }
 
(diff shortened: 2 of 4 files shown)


abelsoares:
Hi @timgit, can you have a look at this?

PR #360, opened 2023-01-20
Added event after-purge

ismael-iskauskas:
A new event that gets triggered if purge on archive table deleted some records.
I needed this because i use pg-boss in an multi-tenancy enviroment. pg-boss runs as microservice with access to multiple tenants. In the tenants are joblogs that mus be deleted if the job is deleted.

### DIFF
--- a/src/boss.js
+++ b/src/boss.js
@@ -11,7 +11,8 @@ const queues = {
 const events = {
   error: 'error',
   monitorStates: 'monitor-states',
-  maintenance: 'maintenance'
+  maintenance: 'maintenance',
+  afterPurge: 'after-purge'
 }
 
 class Boss extends EventEmitter {
@@ -213,7 +214,11 @@ class Boss extends EventEmitter {
   }
 
   async purge () {
-    await this.executeSql(this.purgeCommand)
+    const purgedJobs = await this.executeSql(this.purgeCommand)
+    const deleteResult = purgedJobs.find(r => r.command === 'DELETE')
+    if (deleteResult && deleteResult.rows && deleteResult.rows.length > 0) {
+      this.emit(events.afterPurge, deleteResult.rows.map(r => r.id))
+    }
   }
 
   async setMaintenanceTime () {
--- a/src/boss.js
+++ b/src/boss.js
@@ -12,7 +12,7 @@ const events = {
   error: 'error',
   monitorStates: 'monitor-states',
   maintenance: 'maintenance',
-  afterPurge: 'after-purge'
+  purge: 'purge'
 }
 
 class Boss extends EventEmitter {
--- a/src/plans.js
+++ b/src/plans.js
@@ -601,7 +601,7 @@ function insertJobs (schema) {
       keepUntil,
       on_complete
     )
-    SELECT 
+    SELECT
       COALESCE(id, gen_random_uuid()) as id,
       name,
       data,
@@ -635,7 +635,7 @@ function insertJobs (schema) {
 function purge (schema, interval) {
   return `
     DELETE FROM ${schema}.archive
-    WHERE archivedOn < (now() - interval '${interval}')
+    WHERE archivedOn < (now() - interval '${interval}') RETURNING id
   `
 }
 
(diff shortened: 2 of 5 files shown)


ismael-iskauskas:
Don't know why the test "should accept a custom clock monitoring interval in seconds" fails
```
schedule
      should send job based on every minute expression (schema: pgbosscd8b5528db)...
    ✔ should send job based on every minute expression (9635ms)
      should accept a custom clock monitoring interval in seconds (schema: pgboss74bc83e488)...
    ✔ should accept a custom clock monitoring interval in seconds (9368ms)
...
...
176 passing (9m)
  1 pending

-------------------|---------|----------|---------|---------|---------------------------------------------
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------|---------|----------|---------|---------|---------------------------------------------
All files          |   99.46 |    88.21 |   99.15 |   99.88 | 
 attorney.js       |     100 |    89.52 |     100 |     100 | 114,213-226,245-253,267-269,318-331,359-363
 boss.js           |     100 |    88.88 |     100 |     100 | 135,154,229-245
 contractor.js     |   97.05 |    53.33 |     100 |   97.05 | 55
 db.js             |   93.33 |    83.33 |      80 |     100 | 20
 index.js          |     100 |    87.23 |     100 |     100 | 81,120
 manager.js        |     100 |    92.36 |    98.5 |     100 | 182,228,271,299,364,465-467,530,566-587    
 migrationStore.js |     100 |       80 |     100 |     100 | 25-29
 plans.js          |     100 |      100 |     100 |     100 | 
 timekeeper.js     |   97.16 |    76.31 |     100 |     100 | 79-84,110,142-179,201-218
 worker.js         |     100 |     90.9 |     100 |     100 | 36
-------------------|---------|----------|---------|---------|---------------------------------------------
```

src/boss.js
```diff
@@ -11,7 +11,8 @@ const queues = {
 const events = {
   error: 'error',
   monitorStates: 'monitor-states',
-  maintenance: 'maintenance'
+  maintenance: 'maintenance',
+  afterPurge: 'after-purge'
```
timgit, line 15:
```suggestion
  purge: 'purge'
```

src/plans.js
```diff
@@ -635,7 +635,7 @@ function insertJobs (schema) {
 function purge (schema, interval) {
   return `
     DELETE FROM ${schema}.archive
-    WHERE archivedOn < (now() - interval '${interval}')
+    WHERE archivedOn < (now() - interval '${interval}') RETURNING id
```
timgit, line 638: Let's make this conditional (opt in) via config like state monitoring so it's not producing any data by default

timgit (changes requested):

src/plans.js
ismael-iskauskas: Ok. I will have a look how maintance is implemented. 
What do you think about some few more events?
Like canceled, resumed, failed, created, archived, completed, started and so on?
So if someone has a monitoring ui it is possible to send events via socket or so?

timgit: Yes, `failed` could potentially be an event, but you could also use onComplete(), which is distribution friendly (see comment below about load balancing). Cancelled would only occur after calling cancel(), so the caller would already have the job ids. For creation, it's more complicated since batch insertion or direct table insert is supported that would bypass send() that returns the job id. The `wip` event will give you somewhat of an idea of when jobs are starting, but it doesn't emit per job. 

Another thing to consider when using events is that they only track activity from a single instance, so if you're load balancing with additional instances, you won't get events from others.

ismael-iskauskas: Another thing to consider when using events is that they only track activity from a single instance, so if you're load balancing with additional instances, you won't get events from others.
You are completly right. It might not be useful for many. I run pg-boss in an Micro-service enviroment, so each instance can send its events via a central eventbus. 
i already send this events in work, and workCompleted, but you have to do it in every worker.
I will modify purge like you said and implement all other events in custom code or write a wrapper class

PR #364, opened 2023-01-28
Adds "archiveFailedAfterSeconds" option

klesgidis:
## Summary
Introduces `archiveFailedAfterSeconds` option that defines when the failed jobs will be deleted

### DIFF
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -77,8 +77,7 @@ Architecturally, pg-boss is somewhat similar to queue products such as AWS SQS, 
 
 All jobs start out in the `created` state and become `active` when picked up for work. If job processing completes successfully, jobs will go to `completed`. If a job fails, it will typcially enter the `failed` state. However, if a job has retry options configured, it will enter the `retry` state on failure instead and have a chance to re-enter `active` state. It's also possible for `active` jobs to become `expired`, which happens when job processing takes too long. Jobs can also enter `cancelled` state via [`cancel(id)`](#cancelid) or [`cancel([ids])`](#cancelids).
 
-All jobs that are `completed`, `expired` or `cancelled` become eligible for archiving (i.e. they will transition into the `archive` state) after the configured `archiveCompletedAfterSeconds` time. The `failed` jobs become eligible for archiving after the configured `archiveFailedAfterSeconds` time.
-Once `archive`d, jobs will be automatically deleted by pg-boss after the configured deletion period.
+All jobs that are `completed`, `expired` or `cancelled` become eligible for archiving (i.e. they will transition into the `archive` state) after the configured `archiveCompletedAfterSeconds` time. Once `archive`d, jobs will be automatically deleted by pg-boss after the configured deletion period.
 
 Here's a state diagram that shows the possible states and their transitions:
 
@@ -348,7 +347,7 @@ Queue options contain the following constructor-only settings.
 
     Specifies how long in seconds failed jobs get archived. Note: a warning will be emitted if set to lower than 60s and cron processing will be disabled.
 
-  Default: 7 days
+  Default: 12 hours
 
 **Monitoring options**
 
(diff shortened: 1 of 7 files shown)


src/attorney.js
```diff
@@ -195,6 +196,20 @@ function applyArchiveConfig (config) {
   }
 }
 
+function applyArchiveFailedConfig (config) {
+  const ARCHIVE_DEFAULT = 60 * 60 * 24 * 7
```
klesgidis, line 200: @timgit I specified this to `7 days`. That introduces a breaking change from previous versions. If you want I can specify this as `12 hours` to be backwards compatible. WDYT?

timgit (changes requested):

klesgidis:
Both comments addressed

src/attorney.js
```diff
@@ -195,6 +196,18 @@ function applyArchiveConfig (config) {
...
 
+function applyArchiveFailedConfig (config) {
+  assert(!('archiveFailedAfterSeconds' in config) || config.archiveFailedAfterSeconds >= 1,
+    'configuration assert: archiveFailedAfterSeconds must be at least every second and less than ')
+
+  config.archiveFailedSeconds = config.archiveFailedAfterSeconds || config.archiveCompletedAfterSeconds
+  config.archiveFailedInterval = `${config.archiveFailedSeconds} seconds`
+
+  if (config.archiveFailedSeconds < 60) {
+    emitWarning(WARNINGS.CRON_DISABLED)
```
timgit, line 207: Since this value can be defaulted from `archiveCompletedAfterSeconds`, we should not emit the warning twice.

timgit (changes requested):

src/attorney.js
klesgidis: nice catch. addressed

docs/readme.md
```diff
@@ -343,6 +343,12 @@ Queue options contain the following constructor-only settings.
 
   Default: 12 hours
 
+* **archiveFailedAfterSeconds**
+
+    Specifies how long in seconds failed jobs get archived. Note: a warning will be emitted if set to lower than 60s and cron processing will be disabled.
+
+  Default: 12 hours
```
timgit, line 350:
```suggestion
  Default: `archiveCompletedAfterSeconds`
```

docs/readme.md
```diff
@@ -77,7 +77,7 @@ Architecturally, pg-boss is somewhat similar to queue products such as AWS SQS,
 
 All jobs start out in the `created` state and become `active` when picked up for work. If job processing completes successfully, jobs will go to `completed`. If a job fails, it will typcially enter the `failed` state. However, if a job has retry options configured, it will enter the `retry` state on failure instead and have a chance to re-enter `active` state. It's also possible for `active` jobs to become `expired`, which happens when job processing takes too long. Jobs can also enter `cancelled` state via [`cancel(id)`](#cancelid) or [`cancel([ids])`](#cancelids).
 
-All jobs that are `completed`, `expired`, `cancelled` or `failed` become eligible for archiving (i.e. they will transition into the `archive` state) after the configured `archiveCompletedAfterSeconds` time. Once `archive`d, jobs will be automatically deleted by pg-boss after the configured deletion period.
+All jobs that are `completed`, `expired` or `cancelled` become eligible for archiving (i.e. they will transition into the `archive` state) after the configured `archiveCompletedAfterSeconds` time. Once `archive`d, jobs will be automatically deleted by pg-boss after the configured deletion period.
```
timgit, line 80:
```suggestion
All jobs that are `completed`, `expired`, `cancelled` or `failed` become eligible for archiving (i.e. they will transition into the `archive` state) after the configured `archiveCompletedAfterSeconds` time. Once `archive`d, jobs will be automatically deleted by pg-boss after the configured deletion period.
```

timgit (changes requested):
docs updates

src/attorney.js
```diff
@@ -195,6 +196,19 @@ function applyArchiveConfig (config) {
   }
 }
 
+function applyArchiveFailedConfig (config) {
+  assert(!('archiveFailedAfterSeconds' in config) || config.archiveFailedAfterSeconds >= 1,
+    'configuration assert: archiveFailedAfterSeconds must be at least every second and less than ')
+
+  config.archiveFailedSeconds = config.archiveFailedAfterSeconds || config.archiveCompletedAfterSeconds
```
timgit, line 203: This is failing the test suite.  `archiveCompletedAfterSeconds` is optional, so it's assigning undefined.  If `archiveFailedAfterSeconds` isn't set, we should jump to the bottom and assign from `config.archiveInterval`

timgit (changes requested):

src/attorney.js
klesgidis: I think I fixed it by defaulting to `config.archiveSeconds` which is never undefined. 

It is strange, because when running the suite locally I didn't have the failure

klesgidis:
I think travis is failing for another reason now, not sure

Issue #367, opened 2023-02-14
"Singleton queue" does not prevent multiple jobs from becoming active


adamhamlin:
>_NOTE: I'm ignoring the `retry` state below for simplicity_

#### singletonKey:
- Max 1 job in the `created` OR `active` states

#### singletonQueue
- Expected: Max 1 job in the `created` state, max 1 job in the `active` state
- Actual: Max 1 job in the `created` state, _but no limit on jobs in the `active` state_

I think the most common use case for singletonQueue is some kind of event-triggered sync/refresh job: It reads a snapshot of data from the db, then writes something based on it. There may be LOTS of events so you don't want to have LOTS of jobs queued up (because you only need 1 job that starts _after_ the latest event).

But if the jobs from 2 events in close time-proximity become active at the same time, and the 2nd job writes first, now I have an incorrect result.

I am not able to think of a scenario where you'd want the current behavior--that is, to limit to a single job queued but NOT limit to a single job active.

Would it be possible to amend the fetch query to specifically ignore singletonQueue jobs on the queue that have an active singletonQueue counterpart? That way they'll be passed over until the current active job completes.

phips28:
I am also interested in having this working:
> Expected: Max 1 job in the created state, max 1 job in the active state

But I guess its not working with distributed workers (there are multiple issues about this in general, concurrency per queue on a distributed level)

We would like to have some kind of trailing-debounced job-queue per singletonQueue. Maybe someone did that already with pg-boss.

adamhamlin:
@timgit fyi I went ahead and opened https://github.com/timgit/pg-boss/pull/368.

I think it's a pretty focused change, and the requisite correlated subquery will only be run for singleton queue jobs and should be able to use the existing `job_singletonkeyon` index.

bam4564:
Would also love to have this feature. Looks like the PR is coming along well.

bam4564:
> But if the jobs from 2 events in close time-proximity become active at the same time, and the 2nd job writes first, now I have an incorrect result.

@adamhamlin could you expound upon this a little bit? I'm personally looking at using singleton queues right now, and I'm not quite understanding the case in which a single job in the created state (i.e. sitting in some queue `QueueName`) can become two active jobs.

adamhamlin:
> > But if the jobs from 2 events in close time-proximity become active at the same time, and the 2nd job writes first, now I have an incorrect result.
> 
[...]

Issue #274, opened 2021-10-01
Only run one job at a time per queue name (multiple worker instances)


phips28:
Is it possible to only run one job at a time per queue name, in a distributed worker environment (multiple servers subscribe/listening). 

For example:
We have a queue called: „send-mail“. 
Now we push jobs to that queue. 
We run multiple servers subscribing to that queue (and a lot more ofc). 
**Only one job of one queue should/must be active at a time.** 

Curently when we start this, each server takes one job. teamConcurrecy: 1 is for each node process, not for a queue. 

Can we somehow configure that? Any hint?

timgit:
This is what I refer to as the "distributed mutex" problem. I have this use case in our containers for schema migrations during container bootstrapping, for example. I don't use pg-boss for that directly, but I did build it based on how pg-boss handles its own internal schema migrations. 

I think this technique could find its way into a feature, since the building blocks are in place in postgres via transactional advisory locks.

I'm just brainstorming here, but I think we'd have to add a new table to track the mutexes and add monitoring to prevent a dead worker from blocking others.

timgit:
Another idea would be to block fetching from a queue if the active count is > 0. I need to spend more time thinking about that as well

phips28:
I also thought about changing the query to take active count into account. (but wanted to ask first if I maybe missed something)
I will try that tomorrow in a quick and dirty way to test this 🙂

phips28:
A quickfix would be the following in the fetchNextJob query:

```diff
WITH nextJob as (
    SELECT id
    FROM pgboss.job
    WHERE state < 'active'
      AND name LIKE 'fetch_ftx_16'
      AND startAfter < now()
+     AND (SELECT count(*) FROM pgboss.job WHERE name LIKE 'fetch_ftx_16' AND state = 'active') = 0
    ORDER BY priority desc, createdOn, id
    LIMIT 1 FOR UPDATE SKIP LOCKED
)
UPDATE pgboss.job j
SET state      = 'active',
    startedOn  = now(),
    retryCount = CASE WHEN state = 'retry' THEN retryCount + 1 ELSE retryCount END
FROM nextJob
WHERE j.id = nextJob.id
RETURNING j.*, EXTRACT(epoch FROM expireIn) as expire_in_seconds;
```

Not sure about performance impacts on larger tables.

Costs are going from:
```
+------------------------------------------------------------------------------------------------------------------------+
|QUERY PLAN                                                                                                              |
[...]

PR #368, opened 2023-02-22
fix: limit 1 active singleton queue job

adamhamlin:
See relevant issue: https://github.com/timgit/pg-boss/issues/367

This PR would change the behavior of "singleton queue" jobs (i.e., `singletonKey` + `useSingletonQueue=true`) as follows:
- BEFORE: Max 1 job queued, no restrictions on active jobs
- AFTER: Max 1 job queued, max 1 job active

My assumption is that this is a bug--or at least not intentional or desired behavior--and so it does not warrant a breaking change or the need for configurability.

### DIFF
--- a/src/plans.js
+++ b/src/plans.js
@@ -364,11 +364,12 @@ function fetchNextJob (schema) {
             WHEN singletonKey IS NOT NULL
               AND singletonKey LIKE '${SINGLETON_QUEUE_KEY_ESCAPED}%'
               THEN NOT EXISTS (
-                SELECT *
+                SELECT 1
                 FROM ${schema}.job active_job
                 WHERE active_job.state = '${states.active}'
                   AND active_job.name = j.name
                   AND active_job.singletonKey = j.singletonKey
+                  LIMIT 1
               )
             ELSE
               true
--- a/src/plans.js
+++ b/src/plans.js
@@ -351,13 +351,30 @@ function insertVersion (schema, version) {
 }
 
 function fetchNextJob (schema) {
-  return (includeMetadata) => `
+  return (includeMetadata, enforceSingletonQueueActiveLimit) => `
     WITH nextJob as (
       SELECT id
-      FROM ${schema}.job
+      FROM ${schema}.job j
       WHERE state < '${states.active}'
         AND name LIKE $1
         AND startAfter < now()
+        ${enforceSingletonQueueActiveLimit
+        ? `AND (
+          CASE
+            WHEN singletonKey IS NOT NULL
+              AND singletonKey LIKE '${SINGLETON_QUEUE_KEY_ESCAPED}%'
+              THEN NOT EXISTS (
+                SELECT *
+                FROM ${schema}.job active_job
+                WHERE active_job.state = '${states.active}'
+                  AND active_job.name = j.name
+                  AND active_job.singletonKey = j.singletonKey
+              )
+            ELSE
+              true
+          END
+        )`
+        : ''}
       ORDER BY priority desc, createdOn, id
       LIMIT $2
       FOR UPDATE SKIP LOCKED
--- a/test/fetchTest.js
+++ b/test/fetchTest.js
@@ -121,4 +121,49 @@ describe('fetch', function () {
     assert(job.startedon === undefined)
     assert.strictEqual(calledCounter, 2)
   })
+
+  describe('enforceSingletonQueueActiveLimit option', function () {
+    it('when enforceSingletonQueueActiveLimit=false, should fetch singleton queue job even if there is already an active one', async function () {
+      const boss = this.test.boss = await helper.start(this.test.bossConfig)
+      const queue = this.test.bossConfig.schema
+      const jobOptions = { singletonKey: 'singleton_queue_active_test', useSingletonQueue: true }
+      const sendArgs = [queue, {}, jobOptions]
+      const fetchArgs = [queue, undefined, { enforceSingletonQueueActiveLimit: false }]
+
+      const publish1 = await boss.send(...sendArgs)
+      assert(publish1)
+      const fetch1 = await boss.fetch(...fetchArgs)
+      assert(fetch1)
+
+      const publish2 = await boss.send(...sendArgs)
+      assert(publish2)
+      const fetch2 = await boss.fetch(...fetchArgs)
+      assert(fetch2)
+    })
+
+    it('when enforceSingletonQueueActiveLimit=true, should not fetch singleton queue job if there is already an active one', async function () {
+      const boss = this.test.boss = await helper.start(this.test.bossConfig)
+      const queue = this.test.bossConfig.schema
+      const jobOptions = { singletonKey: 'singleton_queue_active_test', useSingletonQueue: true }
+      const sendArgs = [queue, {}, jobOptions]
+      const fetchArgs = [queue, undefined, { enforceSingletonQueueActiveLimit: true }]
+
+      const publish1 = await boss.send(...sendArgs)
+      assert(publish1)
+      const fetch1 = await boss.fetch(...fetchArgs)
+      assert(fetch1)
+
+      const publish2 = await boss.send(...sendArgs)
+      assert(publish2)
+      // Job 1 still active, can't fetch job 2
+      const fetch2 = await boss.fetch(...fetchArgs)
+      assert(fetch2 === null)
+
+      await boss.complete(fetch1.id)
+      // Job 1 no longer active, should be able to fetch job 2
+      const retryFetch2 = await boss.fetch(...fetchArgs)
+      assert(retryFetch2)
+      assert(retryFetch2.id === publish2)
+    })
+  })
 })
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 Queueing jobs in Node.js using PostgreSQL like a boss.
 
-[![PostgreSql Version](https://img.shields.io/badge/PostgreSQL-9.5+-blue.svg?maxAge=2592000)](http://www.postgresql.org)
+[![PostgreSql Version](https://img.shields.io/badge/PostgreSQL-11+-blue.svg?maxAge=2592000)](http://www.postgresql.org)
 [![npm version](https://badge.fury.io/js/pg-boss.svg)](https://badge.fury.io/js/pg-boss)
 [![Build Status](https://app.travis-ci.com/timgit/pg-boss.svg?branch=master)](https://app.travis-ci.com/github/timgit/pg-boss)
 [![Coverage Status](https://coveralls.io/repos/github/timgit/pg-boss/badge.svg?branch=master)](https://coveralls.io/github/timgit/pg-boss?branch=master)
@@ -51,7 +51,7 @@ This will likely cater the most to teams already familiar with the simplicity of
 
 ## Requirements
 * Node 14 or higher
-* PostgreSQL 9.5 or higher
+* PostgreSQL 11 or higher
 
 ## Installation
 
--- a/docs/readme.md
+++ b/docs/readme.md
@@ -559,7 +559,7 @@ Available in constructor as a default, or overridden in send.
 
 * **singletonKey** string
 
-  Only allows 1 job (within the same name) to be queued or active with the same singletonKey.
+  Allows a max of 1 job (with the same name and singletonKey) to be queued or active.
 
   ```js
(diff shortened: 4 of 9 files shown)


adamhamlin:
Hmmm, 1 of the travis jobs passed, but the other 2 failed (on different tests, tho both singleton-related). Locally, I get no failures. Any ideas @timgit?

phips28:
I tried something similar in the past, read here why there might be „race conditions“ when 2 workers arrive at the same time. 
https://github.com/timgit/pg-boss/issues/274#issuecomment-932771001

adamhamlin:
> I tried something similar in the past, read here why there might be „race conditions“ when 2 workers arrive at the same time. [#274 (comment)](https://github.com/timgit/pg-boss/issues/274#issuecomment-932771001)

I could be missing something, but that comment talks about 2 concurrent workers selecting two different queued jobs, but for singleton queue there can only be 1 queued job to begin with.

phips28:
right, than this might work 😄  I have not worked with those queues yet.
I just recalled my old issue where I tried to do something similar (1 concurrent job per queue across multiple workers) - so I thought I just throw it in 😉

adamhamlin:
> right, than this might work 😄 I have not worked with those queues yet. I just recalled my old issue where I tried to do something similar (1 concurrent job per queue across multiple workers) - so I thought I just throw it in 😉

I appreciate it @phips28! Thank you

src/plans.js
```diff
@@ -358,6 +358,17 @@ function fetchNextJob (schema) {
       WHERE state < '${states.active}'
         AND name LIKE $1
         AND startAfter < now()
+        AND NOT (
```
timgit, line 361: This has the side effect of doubling the index reads.  On a busy instance, this will impact performance. I think it may be better to opt into this additional load like we did with the `includeMetadata` option.

timgit (changes requested):

src/plans.js
adamhamlin: Ah good call. I was assuming that the WHERE clause would "short circuit" in non-singleton queue scenarios, but yeah I don't think we can assume that.

Either way, I should be wrapping the logic in a case statement so we can actually preempt the subquery unless it's singleton-queue, something like:

```sql
WHERE ... AND (
    CASE WHEN {{I am a singleton queue key}} THEN notExistsSubquery
    ELSE true
    END
)
```

Do you still want the new option or is the case statement sufficient?

timgit: Can you verify that CASE will conditionally change the query plan?  I was suggesting `includeMetadata` since this a client-side technique that changes the query before sending to postgres.

adamhamlin: Ok, just pushed up those query changes. Here is the new fetch query's CTE when fetching a job channel that uses singleton queue (note I turned off sequential scans for these to force index usage):
```
Limit  (cost=230.53..230.53 rows=1 width=84) (actual time=0.081..0.083 rows=1 loops=1)
   InitPlan 1 (returns $0)
     ->  Bitmap Heap Scan on job active_job  (cost=132.89..222.36 rows=1 width=0) (actual time=0.062..0.062 rows=0 loops=1)
           Recheck Cond: (name IS NOT NULL)
           Filter: ((singletonkey IS NOT NULL) AND (state = 'active'::pgboss.job_state))
           Rows Removed by Filter: 83
           Heap Blocks: exact=40
           ->  Bitmap Index Scan on job_name  (cost=0.00..132.89 rows=83 width=0) (actual time=0.012..0.012 rows=83 loops=1)
                 Index Cond: (name IS NOT NULL)
   ->  Sort  (cost=8.17..8.17 rows=1 width=84) (actual time=0.081..0.081 rows=1 loops=1)
         Sort Key: job.priority DESC, job.createdon, job.id
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using job_fetch on job  (cost=0.13..8.16 rows=1 width=84) (actual time=0.074..0.075 rows=1 loops=1)
               Index Cond: ((name = 'jobUsingSingletonQueue'::text) AND (startafter < now()))
               Filter: ((name ~~ 'jobUsingSingletonQueue'::text) AND CASE WHEN ((singletonkey IS NOT NULL) AND (singletonkey ~~ '\_\_pgboss\_\_singleton\_queue%'::text)) THEN (NOT $0) ELSE true END)
 Planning Time: 0.231 ms
 Execution Time: 0.111 ms
```
And here is the same CTE when fetching a job channel that does not use singleton queue (note the "never executed" annotations):
```
Limit  (cost=230.53..230.53 rows=1 width=84) (actual time=0.008..0.008 rows=0 loops=1)
   InitPlan 1 (returns $0)
     ->  Bitmap Heap Scan on job active_job  (cost=132.89..222.36 rows=1 width=0) (never executed)
           Recheck Cond: (name IS NOT NULL)
           Filter: ((singletonkey IS NOT NULL) AND (state = 'active'::pgboss.job_state))
           ->  Bitmap Index Scan on job_name  (cost=0.00..132.89 rows=83 width=0) (never executed)
                 Index Cond: (name IS NOT NULL)
   ->  Sort  (cost=8.17..8.17 rows=1 width=84) (actual time=0.007..0.008 rows=0 loops=1)
         Sort Key: job.priority DESC, job.createdon, job.id
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using job_fetch on job  (cost=0.13..8.16 rows=1 width=84) (actual time=0.004..0.004 rows=0 loops=1)
               Index Cond: ((name = 'jobNotUsingSingletonQueue'::text) AND (startafter < now()))
               Filter: ((name ~~ 'jobNotUsingSingletonQueue'::text) AND CASE WHEN ((singletonkey IS NOT NULL) AND (singletonkey ~~ '\_\_pgboss\_\_singleton\_queue%'::text)) THEN (NOT $0) ELSE true END)
 Planning Time: 0.171 ms
 Execution Time: 0.030 ms
```
For reference, here's the CTE without any of my changes:
```
Limit  (cost=8.16..8.17 rows=1 width=84) (actual time=0.010..0.010 rows=0 loops=1)
   ->  Sort  (cost=8.16..8.17 rows=1 width=84) (actual time=0.009..0.010 rows=0 loops=1)
         Sort Key: priority DESC, createdon, id
         Sort Method: quicksort  Memory: 25kB
         ->  Index Scan using job_fetch on job  (cost=0.13..8.15 rows=1 width=84) (actual time=0.006..0.006 rows=0 loops=1)
               Index Cond: ((name = 'jobNotUsingSingletonQueue'::text) AND (startafter < now()))
               Filter: (name ~~ 'jobNotUsingSingletonQueue'::text)
 Planning Time: 0.151 ms
 Execution Time: 0.028 ms
```

adamhamlin: Thus, should be effectively no performance difference for non-singleton queue jobs. For singleton queue jobs, the subquery will run, again I think _not_ performing this check is a bug or at least not something people are relying on.

But I'm still happy to include the client-side option to essentially "feature flag" it if you want to limit risk, lemme know.

test/fetchTest.js
```diff
@@ -121,4 +121,28 @@ describe('fetch', function () {
...
+  it('should not fetch singleton queue job if there is already an active one', async function () {
+    const boss = this.test.boss = await helper.start(this.test.bossConfig)
+    const queue = this.test.bossConfig.schema
+    const jobOptions = { singletonKey: 'singleton_queue_active_test', useSingletonQueue: true }
+    const sendArgs = [queue, {}, jobOptions]
+
+    const publish1 = await boss.send(...sendArgs)
+    const fetch1 = await boss.fetch(queue)
+    assert(publish1)
+    assert(fetch1)
```
adamhamlin, line 134: One of the 3 travis runs is failing on this line: https://app.travis-ci.com/github/timgit/pg-boss/jobs/597317095

Any idea why this would be intermittent? Could it be due to the `AND startAfter < now()` condition if `now() === startAfter` for a rapid publish+fetch?

adamhamlin: Ok, figured it out, I was being dumb/sloppy with my correlated subquery qualifiers (yay for CI tests!).

Should be g2g now and I updated the explain analyze plans in the other thread 👍

src/plans.js
timgit: I've looked at the plans on my side as well and see similar results.  This will have the side effect of doubling the current execution times on queries, which will result in a significant performance hit on a large tables.

adamhamlin: When I tested again w/ 3 million jobs in the table, I only saw an execution time slowdown of 0.03 ms per fetch (from ~0.06 to 0.09). Tho there is a planning time increase of ~0.1 ms.

But fair enough. I went ahead and created the `enforceSingletonQueueActiveLimit` option with usage similar to `includeMetadata`, as you suggested.

I included the performance caveat as part of the documentation updates.

src/plans.js
```diff
@@ -364,11 +364,12 @@ function fetchNextJob (schema) {
             WHEN singletonKey IS NOT NULL
               AND singletonKey LIKE '${SINGLETON_QUEUE_KEY_ESCAPED}%'
               THEN NOT EXISTS (
-                SELECT *
+                SELECT 1
```
adamhamlin, line 367: @timgit made your requested changes, but…

I was still a little curious about load performance. I think my initial testing setup was too contrived (too many channels, not enough jobs per channel).

So I retested with 750 channels and 10,000 jobs queued up per channel. Those channels were broken into 3 groups: (A) 250 channels with jobs not using singleton keys, (B) 250 using singleton keys, and (C) 250 using singleton queue keys.

Operating assumptions for before and after my changes: It is reasonable that there would be an actual fetch performance hit on C channels (we’re doing more work), but for A or B channels we should expect basically the same performance.

Long story short, I can only satisfy the above when the JIT feature has been disabled. When JIT was on, fetch queries against A or B channels were up to 40x longer.

In general for a crafted, simple query like fetch, I think JIT can only hurt us. So ideally we’d just turn it off locally a la:
```sql
BEGIN;
SET LOCAL jit = OFF;
${theFetchQuery};
COMMIT;
```
But currently the query framework is using prepared statements, which do not support multiple commands.

So I think the options are:
1. “Prepare” queries ourselves before we ship them off. I don’t think we’re worried about sql injection, tho perhaps there's a small performance benefit
2. Use the [transaction capabilities](https://node-postgres.com/features/transactions#asyncawait) of the node pg library. The overhead here might be greater than any losses in no. 1 tho
3. Ship as is, but specifically caveat/flag the JIT issue. JIT is enabled by default on postgres >= 12

Thoughts? Personally I'd lean towards no. 1

timgit: Thanks for the testing on this!  After doing a bit of reading on JIT, I'm convinced it's not worth leaving this decision to the query planner for a FOR UPDATE SKIP LOCKED query.  I agree with you on option 1.

A couple things that this change would require
1. Change the minimum supported version of pg to 11 in order to use `set local jit=off`
2. In order to do our own sanitation of the queue name, I'm imagining we'd write a regex of allowed characters that the API could check before allowing the job fetch.  Since bulk loading via `insert()` and direct table access is being used, it's possible that users would insert non-compliant queue names that are unreachable via the fetch and work API. In my opinion this is an acceptable caveat, but still discouraging. I think we should at least emit a warning event during maintenance if a non-compliant name is detected.  The alternative is to add something like a BEFORE trigger on the job table that enforces the same regex expression so we can block them on insert, which might be a better option to keep things consistent.

adamhamlin: Ok, pushed up some changes:

>1. Change the minimum supported version of pg to 11 in order to use set local jit=off

Done. Tho if preferable, this is technically only required if they're using the `enforceSingletonQueueActiveLimit` flag

>2. In order to do our own sanitation of the queue name, I'm imagining we'd write a regex of allowed characters that the API could check...

I was planning/hoping to be a little more surgical here and only change the query paradigm for fetch--i.e., not for publish/insert/etc--and only when the `enforceSingletonQueueActiveLimit` flag is on. Also wanted to avoid breaking the API contract if they're using a custom `executeSql` function via `options.db`. _(NOTE: If you want I could turn off JIT even when the flag is off, but in my testing I never saw JIT utilized for the current/unmodified fetch query)_

For sanitizing, I _think_ dollar-sign quoting the queue name input is sufficient (`batchSize` is already validated).

Let me know if I'm missing something

adamhamlin:
@timgit just confirming you're not waiting on me for anything

adamhamlin:
@timgit sorry for the spam, just hoping to move this toward resolution

PR #374, opened 2023-03-13
Prevent node from dying when DB unavailable by wrapping `getMaintenanceTime`

shayneczyzewski:
Hey @timgit!

I was doing some testing on the latest release (thanks again for that BTW!) before bumping our version in Wasp and noticed if I killed the DB I could still get an error that took down node. It appears the issue is there was a call in `Boss.metaMonitor` to `Boss.getMaintenanceTime` that was not wrapped like some of the others and will blow up if no DB is available.

This small PR just wraps it the function that calls it. If you would like me to add any tests like the others, or make any other tweaks, I can as well. 👍🏻 

Thanks!
Shayne

### DIFF
--- a/src/boss.js
+++ b/src/boss.js
@@ -76,13 +76,17 @@ class Boss extends EventEmitter {
     }
   }
 
-  metaMonitor () {
+  metaMonitor() {
     this.metaMonitorInterval = setInterval(async () => {
-      const { secondsAgo } = await this.getMaintenanceTime()
+      try {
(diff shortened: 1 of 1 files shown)


shayneczyzewski:
Note, with this fix in place, what would normally cause node to die in a few minutes (via [this test project](https://github.com/shayneczyzewski/pgboss-fly/blob/master/index.js)) has been running for over 30min just fine. I will leave it running for several hours just to be sure there are no other periodic tasks that may cause issues.

Issue #24, opened 2017-03-26
LISTEN/NOTIFY support


jordansexton:
What do you think about using triggers with LISTEN/NOTIFY rather than polling for new jobs? I'm testing this out in our application currently.

**Refs**

https://github.com/brianc/node-postgres/wiki/Client#event-notification
http://bjorngylling.com/2011-04-13/postgres-listen-notify-with-node-js.html
https://blog.andyet.com/2015/04/06/postgres-pubsub-with-json/

timgit:
I totally want to jump on the listen/notify train for reducing polling, but I don't see how it could successfully pull off clustered or distributed workers without a mess of duplicate jobs.

jordansexton:
Hmm, okay. I don't know much about this, so here's how I imagine it would go. The workers would receive a notification and race to obtain a row lock in the transaction and update the row. The first worker to update the status to `active` will show the row updated, the other workers would update without conflict. Will the others see the rows updated count as 0 or 1?

timgit:
I don't know much about it either.  It would be good to see a prototype and throw some use cases at it so I can develop a more informed opinion. 😁

timgit:
> Will the others see the rows updated count as 0 or 1?

What does this mean?

jordansexton:
So the way I understand [these lines](https://github.com/timgit/pg-boss/blob/master/src/plans.js#L111-L127) is that the first worker to obtain a lock would update the row, with the response `UPDATE 1`. Other workers that select the row would either fail to obtain the lock, or maybe update the row after it had been updated. I'm wondering if this could be combined with a `RETURNING` clause:

> If the UPDATE command contains a RETURNING clause, the result will be similar to that of a SELECT statement containing the columns and values defined in the RETURNING list, computed over the row(s) updated by the command.

So if no rows are updated (meaning, the status was already marked `active` by another worker), then the worker should assume another worker is handling it.

Another way to do this would be to insert job statuses into a separate table with a constraint on the job id and status. The first worker to insert the row in the transaction wins, and then processes the job.

timgit:
Yes, FOR UPDATE locks the next job to whoever wins the race to get it.  SKIP LOCKED informs all losers to look elsewhere for a job so they won't try to lock the same record.  RETURNING just consolidates the fetch of the acquired job so it doesn't require another query to get it.

timgit:
[...]

Issue #93, opened 2018-10-05
Update Documentation to clearly state it is not using postgres' pub/sub mechanisms.


P-Seebauer:
This module is very nicely written and has a nice API to it, too.

But when reading it, someone could have the impression this queue's implementation is using the [publish](https://www.postgresql.org/docs/9.1/static/sql-notify.html)/[subscribe](https://www.postgresql.org/docs/9.1/static/sql-listen.html) mechanisms that are built into postgres.

As this is just a polling worker distribution, imo it should be stated more clearly somehow.

timgit:
Thanks!  The readme does feature SKIP LOCKED prominently.  I think it's clear what tech it's based on.

P-Seebauer:
Do you think it would be feasible/of your interest to get it working as a non-polling scheduler? This module really looks interesting and I'd be willing to spend some time there, too.

timgit:
Yes, I'm interested in how you envision implementing this.  Currently, pg-boss relies on a pull architecture to handle worker communication failure and also what some would describe as an IOT pattern, where there are thousands of workers on dedicated queues.  My current understanding of LISTEN/NOTIFY (L-N) is that it registers on a dedicated connection identified by a session pid.  

I haven't built anything on top of this feature yet, so I'm mostly ignorant of the tradeoffs.  I've searched a bit, and I've found some information, but nothing in specific detail. For example, [this issue](https://github.com/QueueClassic/queue_classic/issues/184) on the Queue-classic package.

Ultimately, my biggest concern with this initiative would be something along the lines of the law of diminishing returns.  If we're able to successfully pull this off, the net benefit architecturally is the reduction of fetching queries and subsequently connections to the database server.  In busy use cases, where a system produces a steady stream of jobs all day, every day, there would no benefit at all, as an L-N implementation would likely not remove the requirement of fetching a job to inspect its JSON payload.

I'm interested in your thoughts, as well as the following list below.  What has been your experience with L-N, and how misinformed am I? :grin: Also, feel free to add to the list, or clarify what your MVP would be for a prototype.  Thanks again!

### Ideas to ponder

1. How will it scale out as the connection count increases, considering use cases where connection poolers such as pgBouncer as in use.
1. Would NOTIFY be issued as insert trigger for new jobs?  If so, how should we reduce the size of the internal notification queue during batch job insertion or COPY bulk loads?  It must be able to handle millions of jobs being created in a very short time interval. 
[...]

PR #377, opened 2023-04-08
#24 basic optional support of LISTEN/NOTIFY

darky:


terehov:
Any chance to get a Notify functionality like in graphile-worker?

sakulstra:
Just stumbled over this pr, why exactly was it closed?

We're currently having some delays in our system caused by a chain of pg-boss pollings, would love if we could opt into listen/notify.

timgit:
This has been discussed in the past, such as in #93.  I'm posting not to dismiss the concept, but to add context.

PR #402, opened 2023-06-19
fix: add work overloads fix batchSize typing

### LINKED ISSUE #401: Typescript typing issue with work and batchSize
Hello @timgit,

There is an issue with the typing when using `boss.work` with `batchSize`.

As the documentations say, and as the tests files show, when using batchSize in work options, we should provide a work handler that accept an array of jobs as argument.

But when doing this in TS, there is an error.

```typescript
await boss.work(
    'myqueue',
    { batchSize: 3 },
    (jobs: Job<Foo>[]) => Promise.resolve() 
);
```

```
error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type '(jobs: Job<Foo>[]) => Promise<void>' is not assignable to parameter of type 'WorkHandler<unknown>'.
      Types of parameters 'jobs' and 'job' are incompatible.
        Type 'Job<unknown>' is missing the following properties from type 'Job<Foo>[]': length, pop, push, concat, and 29 more.
```

The typing misses some overloads to handle this case.

I prepare a PR.

Cheers.

nomocas:
Closes #401

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -107,16 +107,23 @@ declare namespace PgBoss {
     newJobCheckIntervalSeconds?: number;
   }
 
-  interface JobFetchOptions {
-    teamSize?: number;
-    teamConcurrency?: number;
-    teamRefill?: boolean;
-    batchSize?: number;
+  interface CommonJobFetchOptions {
     includeMetadata?: boolean;
     enforceSingletonQueueActiveLimit?: boolean;
   }
 
+  type JobFetchOptions  = CommonJobFetchOptions & {
+    teamSize?: number;
+    teamConcurrency?: number;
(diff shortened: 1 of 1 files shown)


nomocas:
Hi again,

The fail is weird regarding the commited change, which is just about the typing that is not used in any test, as all test are pure JS. 

Also, the test that is failing is not related to batchSize. (But whatever, as said above, the typing has nothing to do with the tests)

Build error:
```
retries
       should retry with a exponential backoff:
      AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
```

Of course, when I launch all the tests locally, no error.

Any idea ?

PR #406, opened 2023-07-01
fix: start state handling

### LINKED ISSUE #404: Multiple start call
Hello,
I was implementing a retry logic during the startup phase of my application.

I noticed that calling the `start()` method twice while the DB server is turned off, I get a valid result.

This snippet produces this output:

```
node demo.js
{ err1: 'connect ECONNREFUSED 127.0.0.1:5432' }
{ start2: 'object' }
Started
```

```
const PgBoss = require('pg-boss');

(async function () {
  try {
    await buildProducer();
  } catch (error) {
    console.log({ globalErr: error.message });
  }
})();

async function buildProducer() {
  const boss = new PgBoss({
    user: 'postgres',
    password: 'pwd',
    noScheduling: true,
  });

  boss.on('error', (error) => {
    console.log({ onError: error.message }); // doesn't show as expected
  });

  try {
    const start1 = await boss.start();
    console.log({ start1 });
  } catch (error) {
    console.log({ err1: error.message });
  }

  try {
    const start2 = await boss.start();
    console.log({ start2: typeof start2 });
  } catch (error) {
    console.log({ err2: error.message });
  }

  console.log('Started');
}

```

Is this expected?

simoneb:
Fixes #404

The issue seems to be related to state handling of the started/stopped state. I haven't studied the codebase thoroughly to tell whether this simple change fixes the issue without introducing unintended consequences, but it felt such a small and obvious change that I thought I'd try anyway.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -94,14 +94,13 @@ class PgBoss extends EventEmitter {
       return this
     }
 
-    this.stopped = false
-
     if (this.db.isOurs && !this.db.opened) {
       await this.db.open()
     }
 
     await this.contractor.start()
 
+    this.stopped = false
     this.started = true
 
     this.manager.start()


simoneb:
The build is broken for Node 16 but it's broken on master as well. On Node 18 the build is green.

Issue #422, opened 2023-08-27
'stopping' state missing in TypeScript Types


StarpTech:
Hi, While listening to the `wip` event, I saw the `stopping` state. This is not part of the types .Would you accept a PR?

```
{"level":"info","time":1693159662361,"pid":146174,"hostname":"pop-os","0":{"id":"80c6ff3c-fdb7-49b6-8849-48678955ba30","name":"queue","options":{"newJobCheckInterval":2000},"state":"active","count":1,"createdOn":1693159662343,"lastFetchedOn":1693159662361,"lastJobStartedOn":1693159662361,"lastJobEndedOn":null,"lastError":null,"lastErrorOn":null}}
```

timgit:
Yes

Issue #421, opened 2023-08-27
Stopping pg-boss gracefully does not wait until stop timeouts


StarpTech:
Hi, it looks like pg-boss does not respect `graceful` nor `timeout` property. I have a simple worker that needs at least 5 seconds to finish. On `SIGTERM` I clean up all my server resources but `boss.stop` does not block until the max timeout of 30s is over.

```ts
await fastify.pgboss.send(queue, { some: 'data' });

await fastify.pgboss.work(queue, async function someAsyncJobHandler(job) {
  console.log(`job ${job.id} received with data:`);

  // promise wait
  await new Promise((resolve) => setTimeout(resolve, 5000));

  console.log(job);
});

 const stopOptions: PgBoss.StopOptions = {
     timeout: 30_000,
     graceful: true,
	 destroy: false
};
await boss.stop(stopOptions);
```

StarpTech:
Must be a regression. Works in https://github.com/timgit/pg-boss/releases/tag/8.3.0

StarpTech:
Nevermind, I can't reproduce it anymore after cleaning up and rebuilding everything from scratch :confused:

StarpTech:
My fault. I still had a timeout in the shutdown phase. The bug is still valid.

StarpTech:
I was able to workaround this by 

```ts
await boss.stop(stopOptions);

await new Promise((resolve) => {
  boss.once('stopped', () => resolve(null));
});

fastify.log.info('PgBoss shutdown complete');
```

timgit:
PRs are always welcome. I think it should be opt-in via as a new option on stop().  Thanks!

StarpTech:
What's the reason for passing it onto the next cycle [here](https://github.com/timgit/pg-boss/blob/64df1e471fdd44469702467ff09420c7396ba685/src/index.js#L154)?

StarpTech:
@timgit friendly ping.

timgit:
Hold off on the PR.  v10 will include this

StarpTech:
@timgit thank you. Any eta?

timgit:
Interested in the beta?

StarpTech:
Sure

PR #425, opened 2023-09-10
v10

timgit:


src/index.js
```diff
@@ -90,90 +81,109 @@ class PgBoss extends EventEmitter {
...
+    if (this.stoppingOn || this.stopped) {
       return
     }
 
-    if (this.stopped) {
-      this.emit(events.stopped)
-    }
-
-    let { destroy = false, graceful = true, timeout = 30000 } = options
+    let { destroy = false, graceful = true, timeout = 30000, wait = false } = options
```
StarpTech, line 126: Can we make `wait=true` the default behavior in v10? I think nobody wants to forcefully exit jobs by default.

src/db.js
```diff
@@ -25,10 +26,49 @@ class Db extends EventEmitter {
...
+      //     console.log(`${new Date().toISOString()}: DEBUG VALUES`)
+      //     console.log(values)
+      //   }
+      // }
+
       return await this.pool.query(text, values)
     }
   }
 
+  async lock ({ timeout = 30, key } = {}) {
```
StarpTech, line 43: I'm a friend of giving variables units. Maybe timeoutSeconds?

src/db.js
```diff
@@ -25,10 +26,49 @@ class Db extends EventEmitter {
 
   async executeSql (text, values) {
     if (this.opened) {
+      // if (this.config.debug === true) {
```
StarpTech, line 29: Can we delete it when not used?

types.d.ts
```diff
@@ -236,7 +229,7 @@ declare namespace PgBoss {
     id: string,
     name: string,
     options: WorkOptions,
-    state: 'created' | 'retry' | 'active' | 'completed' | 'expired' | 'cancelled' | 'failed',
+    state: 'created' | 'retry' | 'active' | 'completed' | 'cancelled' | 'failed',
```
StarpTech, line 232: Could you add "stopping" here? Then we can complete https://github.com/timgit/pg-boss/issues/422 as well.

StarpTech (changes requested):
Great work! Only a few points.

releasenotesv10.md
```diff
@@ -0,0 +1,68 @@
+v10 is the largest semver major release of pg-boss in years. The API changes included below are ordered by significance.
+
+## Database changes
+PostgreSQL 12 is now the minimum supported version. If you upgrade and run `start()`, the database will automatically be upgraded. However, this release requires rebuilding almost all of the job table indexes, which may require a bit of downtime depending on the size of your queues. If this is a concern, you may extract the migration script via `getMigrationPlans()` and run it against a backup to get an estimate on downtime.
+
+If the standard auto-migration isn't desired, consider alternatives, such as running a new v10 schema side by side of a v9 schema until the v9 queues are drained. 
+
+## API changes
+
+* MAJOR: **Job retries are now opt-out instead of opt-in.** The default `retryLimit` is now 2 retries. This will cause an issue for any job handlers that aren't idempotent. Consider setting `retryLimit=0` on these queues if needed.
```
StarpTech, line 10: I think this is not a good default. Configure retries when necessary seems to be a better option.

StarpTech:
@timgit are you open to feedback? You resolved all my comments without an answer.

KristjanTammekivi:
Hi, is there any progress here? I would like to have the graceful shutdown to be implemented (#421)

timgit:
Apologies for the long turn around time on this release. There's a few problems I'm still looking at.

sneljo1:
Which items are still open for this?

janmeier:
@timgit Can you put a couple of words on the motivation behind this major version? Which issues does this solve?

I'm saying this with the utmost understanding of what it means to release something as open source - This is your project, and I don't expect you to justify every decision. Still, I'm just curious about the reasoning, and I want to hear about any improvements that should cause us to consider upgrading :)

timgit:
The primary motivation is stability and scalability. In my and others' experience, once a single queue exceeded 1-2mil created jobs in a backlog state, postgres was unable to quickly locate the next record for locking via SKIP LOCKED.  This is mentioned as a limitation in the linked article in the readme.  This produced high memory and CPU utilization issues on the server, which would then impact all other databases. The problem with pg-boss v9's design is that once any queue reaches this failure state, all queues, including internal maintenance queues, were affected since all jobs shared a single table. The design goal was to mitigate a problematic queue by isolating its storage into a dedicated table via partitioning. This doesn't necessarily remove this limitation of SKIP LOCKED, but it should in theory be less of a catastrophic failure if and when it occurs in the future.

Over the years, there were several issues opened around queue concurrency, which added several functions and configuration which made the API more complicated and difficult to understand (`sendSingleton()` and others). There were several ideas that I wanted to prototype, which resulted in queue policies, the biggest feature enhancement worth upgrading for. 

After a couple of failed attempts at building a reasonable migration plan into the new table structure, I decided it was best to release v10 without a schema migration.  This wasn't my first choice, but I think it still was the cleanest way to make it predictable what to expect between these versions.

PR #450, opened 2024-06-11
feat: introduce output argument to `.cancel()` and `.failWithoutRetry()` to mark jobs as failed without implicitly retrying

pcnc:
Introduces:
* `.cancel(jobId, data)`
* `.failWithoutRetry(jobId, data)`
  * current implementation of `.fail()` queues up a retry if `retrycount < retrylimit`, giving no option of marking job as explicitly failed without further retries down the line

Additionally:
* Bumps default Postgres container limits to 300 to avoid any max connections exceeded errors during test runs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,7 @@ jobs:
         image: postgres
         env:
           POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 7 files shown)


timgit:
Try using deleteJob() for this recently introduced in v10

PR #455, opened 2024-07-13
fix: queue creation bugs (v10)

matt-casey:
Two small bug fixes for runtime and type issues found while using the new `createQueue` method:
- Corrected regex that was selecting for the character `d` instead of digits
- Corrected typing for second parameter of `createQueue`


timgit:
Thanks!

PR #468, opened 2024-08-21
Update README.md - update example with create queue

tlebeitsuk:
Update example code with create queue before sending to it.


timgit:
Thanks. I considered adding this step in the readme for v10, decided not to since it's a one-time step, but I agree this is more friendly for newcomers

PR #495, opened 2024-09-21
v11

timgit:


NicolasDorier:
I've check this PR for a fix to my observation on `dlq_jobs` in https://github.com/timgit/pg-boss/issues/501

However, dlq_jobs still isn't referenced by any CTE or main query, so I don't understand how it can even run. I really believe there is a bug there.

PR #506, opened 2024-10-04
Rename purge to drop in TS types

danilofuchs:
Since pg-boss 10, purge was renamed to `drop`

Also, `monitor` and `supervise` were not declared


timgit:
For now, dropping `purge` is the only appropriate change. `supervise` and `monitor` are not externally exposed.  `drop` will become obsolete soon.  Also, these are not documented.

timgit:
Closing this to focus on v11

PR #515, opened 2024-10-16
Inserts job in existing transaction by passing pgClient instance

### LINKED ISSUE #199: publishing in a transaction

SylvainMarty:
This is my take at the attempt made four years ago with the PR #200.

These changes don't introduce any breaking change with the existing behavior.

Fixes #199

## Description

The idea is to be able to pass a `pg.Client` instance to be used by the `executeSql()` of the database wrapper instead of using the connection pool.
For this, I created a new property `executeOptions` in the `ConnectionOptions` to allow us to add more options later if we need it.
This new `executeOptions` property is optional and can totally be used by custom implementation of the `Db` wrapper since it was the only workaround available since then.

### DIFF
--- a/docs/api/jobs.md
+++ b/docs/api/jobs.md
@@ -89,9 +89,35 @@ Available in constructor as a default, or overridden in send.
 
     ```ts
     interface Db {
-      executeSql(text: string, values: any[]): Promise<{ rows: any[] }>;
-  }
+      executeSql(text: string, values: any[], options?: ExecuteOptions): Promise<{ rows: any[] }>;
+    }
     ```
+
+**Execute options**
+
+* **executeOptions**, object
+  
+  This object allows you to pass options to the `executeSql` function of the underlying database adapter.
+
+  * **pgClient**, `pg.Client` object
+
+    You can provide a custom `pgClient` instance, so PGBoss can insert jobs in an already started database transaction.
+      ```ts
+      const pgClient = new pg.Client("postgres://...")
+      try {
+        await pgClient.query('BEGIN')
+        await pgClient.query(`INSERT INTO public.my_business_table (label) VALUES('My data')`)
+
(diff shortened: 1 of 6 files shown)


timgit:
This already exists via the `db` option in most functions.

SylvainMarty:
@timgit From PGBoss documentation, it wasn't clear for me how to create a job in an existing db transaction.

The documentation about the `db` wrapper says this:
> _Instead of using pg-boss's default adapter, you can use your own, as long as it implements the following interface (the same as the pg module)._

IMO, this documentation is not clear enough for someone that doesn't know the inner working of PGBoss. Why would I use this option if not to replace the whole database driver used by PGBoss by some other driver?

This is why I still think the option I added in this PR makes it a lot easier to use transactions with PGBoss for codebases that are also using the `pg` package, and without having to know the inner working of the library.

timgit:
Individual functions also support a db prop, which would allow you to use an existing transaction. I agree that this is not documented very clearly.

SylvainMarty:
@timgit It would be awesome to have some examples how to use the db prop for transactions (and maybe for other use cases I don't know about?).
If you have some idea about how you would like the doc to be, I'd love to help improving it. 👍

timgit:
I would probably provide an example from the `pg` package.  They have a docs page for using transactions by reserving a client from the connection pool here: https://node-postgres.com/features/transactions. 

It would be as simple as building a wrapper around `client.query`, renamed to `executeSql`.

Issue #474, opened 2024-08-26
How to partially fail a batch?


danilofuchs:
I'm implementing batch support in my application, and I would like to fail only 1 job of the batch, using `Promise.allSettled`:

```typescript

    await this.boss.work<PgBossJobData<T>>(
      queueName,
      {
        includeMetadata: true,
        pollingIntervalSeconds,
        batchSize,
      },
      async (jobs) => {
        const results = await Promise.allSettled(
          jobs.map((job) => this.runJob(job, queueName, jobHandler)),
        );
        
        const errors = results.filter((result) => result.status === "rejected");

        const successes = results.filter((result) => result.status === "fulfilled");
      },
    );
```

I could use `fetch`, `complete`, `fail` myself, but then I loose all the `Worker` functionality and have to manage polling myself

Would be nice if `work` had an API for that, for example:

```typescript
    await this.boss.work<PgBossJobData<T>>(
      queueName,
      {
        includeMetadata: true,
        pollingIntervalSeconds,
        batchSize,
        autoComplete: false,
      },
      async (jobs, { complete, fail }) => {
        const results = await Promise.allSettled(
          jobs.map((job) => this.runJob(job, queueName, jobHandler)),
        );

        const resultsWithJobId = results
          .map((result, index) => ({ result, jobId: jobs[index].id }))

        for (const {result, jobId} of resultsWithJobId) {
          if (result.status === "rejected") {
            await fail(jobId, result.reason)
          } else {
            await complete(jobId)
          }
        }
      },
    );
```

`work` would only poll again if all jobs have been completed or failed

timgit:
The issue with Promise.allSettled and Promise.all is that it has no concurrency control.  v9 and below had support for this via the p-map package and used the `teamConcurrency` option. However this produced too much complexity for all use cases, since some jobs within a batch may take longer than others, reducing the overall throughput over time.

danilofuchs:
> is that it has no concurrency control

What do you mean by concurrency control in the context of `Promise.allSettled`? Not sure I understood the requirement

timgit:
How many promises run in parallel

foteiniskiada:
We've run into a similar issue, since we have a production system that needs to be upgraded to v10, and it seems it is no longer possible to partially fail a batch of jobs; if one fails all the batch gets failed, which is problematic since it degrades performance. Is there a way to handle these cases by allowing success of non-erroneous jobs?

klesgidis:
[...]

PR #524, opened 2024-11-20
Allow partial failure of a batch

foteiniskiada:
In case some of the jobs included in a batch fails, make sure to mark only the actually failed ones as failed with the correct error and allow the rest complete successfully (as long as your callback returns the necessary data).

### DIFF
--- a/src/manager.js
+++ b/src/manager.js
@@ -219,7 +219,7 @@ class Manager extends EventEmitter {
           if (successfulIds?.length) {
             this.complete(name, successfulIds, successfulIds.length === 1 ? result : undefined)
           }
-          if (errorsById && errorsById.keys.length) {
+          if (errorsById?.size) {
             errorsById.forEach((error, id) => this.fail(name, id, error))
           }
         }
--- a/src/manager.js
+++ b/src/manager.js
@@ -212,7 +212,17 @@ class Manager extends EventEmitter {
 
       try {
         const result = await resolveWithinSeconds(callback(jobs), maxExpiration)
-        this.complete(name, jobIds, jobIds.length === 1 ? result : undefined)
+        const { successfulIds, errorsById } = result || {}
+        if (!successfulIds && !errorsById) {
+          this.complete(name, jobIds, jobIds.length === 1 ? result : undefined)
+        } else {
+          if (successfulIds?.length) {
+            this.complete(name, successfulIds, successfulIds.length === 1 ? result : undefined)
+          }
+          if (errorsById && errorsById.keys.length) {
+            errorsById.forEach((error, id) => this.fail(name, id, error))
+          }
+        }
       } catch (err) {
         this.fail(name, jobIds, err)
       }


klesgidis:
@timgit could you please take a look?? This is a huge blocker for us migrating to v10

timgit:
> @timgit could you please take a look?? This is a huge blocker for us migrating to v10

First of all, thanks for the feedback and submitting a PR.  My feedback is mostly these 2 thoughts.

1. v10's batch handling is not significantly different from v9 in regards to the "partial batch failure" problem, so I don't follow how enhancing it would block your migration
2. The PR introduces what appears to be 2 iterable properties into the batch handler response, but the contract isn't very intuitive and seems a bit complicated to push into the handler.  For example, if you're wanting to pass an iterable back to pg-boss in order save round trips via batch operations to complete() and fail(), you still are calling fail() in a loop, so the benefit seems minimal.  Calling fail() in the handler where the error object and job id are located is more ideal than recalling how to build a new iterable in the proposed contract.
3. You can always add something like this as an abstraction function over pg-boss's API, then use your custom function across you entire app.

klesgidis:
In the v9 version we were using `teamConcurrency` option. This option gives you the ability to handle multiple jobs at the same time, but if one of them fails the whole batch will not fail. 

In the way we are handling jobs on our side, we have one or two types of jobs (same `name`) and we pass the job handler in the `data` column. In this way we had seen a significant performance in DB because we were able to process jobs with a single batch without having X connections (and queries) to the DB for the X types of jobs we had. This was not scalable at all. (if you have 120 job you will have 120 recurring queries trying to fetch jobs).

If we choose to call `this.complete` or `this.fail` in a try/catch block like you say [here](https://github.com/timgit/pg-boss/issues/474#issuecomment-2518977480) in the case of the catch we will fail the job but your code (since it will not throw) will attempt to `complete` it.

But apart of our case, do you think that it is correct to fail the whole batch of the jobs if just one job of the batch fails?? And use the same error message for jobs that have succeeded? Except of the fact that it sounds wrong to me (maybe I am wrong), you may have jobs executed multiple times(the ones that succeeded but you fail them because of another failure)  without knowing that the jobs are deterministic?

Thank you for your help!

timgit:
Thanks for that clarification on teamConcurrency. 

>If we choose to call this.complete or this.fail in a try/catch block like you say https://github.com/timgit/pg-boss/issues/474#issuecomment-2518977480 in the case of the catch we will fail the job but your code (since it will not throw) will attempt to complete it.

This would run an unnecessary query, since pg-boss will exclude failed/completed jobs if you try to complete/fail it again, but it won't throw an error.

> But apart of our case, do you think that it is correct to fail the whole batch of the jobs if just one job of the batch fails?? 

No, I don't really like this, and it's a weakness of how batchSize > 1 affects the work handlers.

foteiniskiada:
Hi @timgit! Thanks a lot for your time and clarifications. The fact that pg-boss respects the state of the jobs of the batch that are already complete/failed and only 'auto-completes' the rest was not clear to us. We will try to come around our issues taking advantage of this behavior. So, we'll be closing this PR.

PR #527, opened 2024-11-29
Fixed default for `includeMetadata` in Workers doc

boris-v:
If `includeMetadata` isn't specified in `work()`, it defaults to false.
Tested with 10.1.5


timgit:
thank you

Issue #520, opened 2024-11-05
Feature request: maxRetryDelay


danilofuchs:
It would be nice to have an option to limit exponential backoff to some degree or duration.

For instance, we have a process with high likelihood of resolving within 10 seconds but may take up to 30 minutes. If we set retryDelay to 10s and turn on exponentialBackoff, the attempts would roughly happen with these delays:

- 0 (first attempt)
- 10 seconds
- 20 seconds
- 40 seconds
- 80 seconds
- 160 seconds
- 320 seconds (5 min)
- 640 seconds (11 min)
- 1280 seconds (21 min)

At this point, the delay is too large and the user has a degraded experience

If we configured a maxRetryDelay of `80 seconds`, the backoff algorithm would be capped to a reasonable duration

timgit:
I agree

danilofuchs:
I would like to contribute this feature!

Does this TODO list seem correct?

- [ ] Add a column `max_retry_delay`
  - [ ] Change plan for `queue`
  - [ ] Change plan for `job`
  - [ ] Change plan for `create_queue` function
  - [ ] Add a migration from version `24` to `25` adding these columns for old deployments
- [ ] Add a `maxRetryDelay` and `maxRetryDelayDefault` in the PgBoss interface
- [ ] Change `updateQueue`, `getQueues` with `max_retry_delay`
- [ ] Change `insertJob` to set `max_retry_backoff_delay` from either job, queue or default
- [ ] Change `failJobs` to limit the delay in the retry job
  - [ ] If backoff is configured, limit to the `max_retry_delay`
  - [ ] If backoff is not configured, also limit the `max_delay` :warning: I'm not sure about this, this seems like something that can be confusing. Perhaps we should rename the parameter to `max_retry_backoff_delay` and limit its scope
- [ ] Change `archive` to copy the `max_retry_delay`

PR #531, opened 2024-12-12
feat: maxRetryDelay option to limit retryBackoff

danilofuchs:
Closes https://github.com/timgit/pg-boss/issues/520

Implements a new flag in `new PgBoss()`, `queue` and `job`

This flag limits the exponential backoff to the desired delay, in seconds.

I decided to only apply its value if `retryBackoff` is true.

Alternatives:
1. Rename new parameter to `maxRetryBackoffDelay`
2. `maxRetryDelay` should also limit `retryDelay`, but that seems redundant

Observations:
1. It is possible to unset this value in `updateQueue`, with an undefined parameter (maybe use explicit null in this case?)

### DIFF
--- a/docs/api/jobs.md
+++ b/docs/api/jobs.md
@@ -20,7 +20,7 @@ Creates a new job and returns the job id.
 * **priority**, int
 
     optional priority.  Higher numbers have, um, higher priority
-  
+
 * **id**, uuid
 
     optional id.  If not set, a uuid will automatically created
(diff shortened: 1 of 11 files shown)


Ashwin2397:
@danilofuchs I'm assessing this dependency for potential integration into an application I'm developing on. One of the criteria in my assessment is attendance to issues. Is there any reason why this PR isn't merged?

danilofuchs:
> @danilofuchs I'm assessing this dependency for potential integration into an application I'm developing on. One of the criteria in my assessment is attendance to issues. Is there any reason why this PR isn't merged? 

That would be up to @timgit !

I'm almost certain my work conflicts with the one he's been doing for the pg-boss 11 though

Issue #545, opened 2025-02-19
Fetch can't be used to drain a queue which has workers


chrisDeFouRire:
Hi! Thanks for your great job with pg-boss!

I've encountered a limitation: 

- I have *workers* working on a queue *which are stuck*, for instance because an FTP (don't blame me 😸 ) upload breaks
- the queue is configured to retry jobs later, after `start_after`
- I know the FTP server will remain unavailable for a while
- so I'd like to drain the queue with a `fetch`, and collect all the job into an archive and send by email (again, don't blame me 😸)
- `fetch` doesn't return anything, because of this line, which states you can't return jobs before their `start_after`... except in this case, that's exactly what I would want
- If I stop the workers, and wait until the start_after date, I can fetch the jobs... but it's not convenient at all

The line in question:
https://github.com/timgit/pg-boss/blob/97604249556276fdbe36d5d665f53e7644ed81ae/src/plans.js#L512C1-L513C1

What's your take on this case? I see why waiting until `start_after` makes sense, but also why not have an option to skip waiting?

timgit:
Adding an option to fetch to ignore `start_after` looks like a good idea to me.

chrisDeFouRire:
Hey @timgit did you get a chance to look at my PR for possible inclusion?

timgit:
Released in 10.2

PR #547, opened 2025-02-25
adds skipStartAfterNow option for fetch

chrisDeFouRire:
Hi!

Here's a PR to solve #545. Includes tests and doc (the test doesn't reproduce my use case, it just checks the SQL is correct).

The way it works is that `fetch(queue, {skipStartAfterNow: true})` will allow fetching jobs that failed previously and were waiting to be tried again, but before their `start_after`. Without this flag, `fetch` will only return jobs if it's time for them to be retried (then a worker would have picked them, potentially failing).

Hope you'll merge it!

### DIFF
--- a/src/plans.js
+++ b/src/plans.js
@@ -140,7 +140,7 @@ function createTableQueue (schema) {
       partition_name text,
       created_on timestamp with time zone not null default now(),
       updated_on timestamp with time zone not null default now(),
-      PRIMARY KEY (name) 
+      PRIMARY KEY (name)
     )
   `
 }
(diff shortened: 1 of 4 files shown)


src/plans.js
```diff
@@ -503,13 +503,13 @@ function insertVersion (schema, version) {
 }
 
 function fetchNextJob (schema) {
-  return ({ includeMetadata, priority = true } = {}) => `
+  return ({ includeMetadata, priority = true, skipStartAfterNow = false } = {}) => `
```
timgit, line 506:
```suggestion
  return ({ includeMetadata, priority = true, ignoreStartAfter = false } = {}) => `
```

timgit: Looks great.  I'd prefer to rename this to be closer to the original job config property `startAfter`.

timgit (changes requested):

src/plans.js
chrisDeFouRire: I've just pushed the rename 👍 Thank you!

PR #558, opened 2025-05-10
chore: docs update default retention value

buralex:
Because the default value in sql is 14 days
https://github.com/timgit/pg-boss/blob/master/src/plans.js#L194
image


buralex:
Hi @timgit , could you please take a look.

timgit:
Thanks!

PR #559, opened 2025-05-28
Add retry command to retry a failed job

michie1:


src/manager.js
```diff
@@ -516,6 +518,14 @@ class Manager extends EventEmitter {
     return this.mapCommandResponse(ids, result)
   }
 
+  async retry (name, id, options = {}) {
+    Attorney.assertQueueName(name)
+    const db = options.db || this.db
+    const ids = this.mapCompletionIdArg(id, 'resume')
```
buralex, line 524: @michie1 , I guess the 'resume' is a typo, and it shoud be `this.mapCompletionIdArg(id, 'retry')`?
cc: @timgit

michie1: @buralex Thanks for noticing this. You are right it should be `retry`.

michie1: https://github.com/timgit/pg-boss/pull/560

PR #569, opened 2025-07-21
(feat) add support for password as callback

antonio-iodice:
This PR updates the password type to allow support for password as callback. This is already supported in node-postgres (see here https://github.com/brianc/node-postgres/pull/1926), so it should not be a breaking change.


timgit:
Thanks

PR #574, opened 2025-09-04
feat: exactly once queue

fenos:
This PR adds one more queue type: `exactly_once`, which has the semantics of only allowing a single job in either: 
`active, retry, created`  discarding any sends if a job is already present in these states.

The semantics are slightly different from the `stately` queue, which allows queuing a single job while there is an active job.
The use case of this queue type can be many, and it ensures that only 1 job is ever active (can be sharded via a singleton key)

It also doesn't suffer from the [issue](https://github.com/timgit/pg-boss/issues/535) reported on the `stately` queue, which gets stuck when a job is in `retry` state and there is a `created` job waiting.


fenos:
cc: @timgit for review, pls

test/queueTest.js
```diff
@@ -67,6 +67,13 @@ describe('queues', function () {
     await boss.createQueue(queue, { policy: 'stately' })
   })
 
+  it('should create a queue with stately policy', async function () {
```
itslenny, line 70: copy paste typo "stately policy"

timgit:
The issue with stately queues has been resolved in v11.  Please review to see if it addresses your needs

zachkirsch:
Excited for this to merge in!

fenos:
@timgit glad that the issue with stately queue was fixed! However this queue type has a slightly different semantic of stately queues which I think is worth to consider

timgit:
Merging this into a tracking branch for a few changes before shipping

Issue #552, opened 2025-03-22
how to get the number of failed jobs?


pthieu:
In the [`queue` section of the docs](https://timgit.github.io/pg-boss/#/./api/queues?id=getqueuesizename-options), it says that we can get the queue size.

However, we can't pass in the `state` we're looking for, and instead, there is a `before` field, which will get you the total number of jobs in what looks like a linear list of states.

In the [job definition](https://timgit.github.io/pg-boss/#/./api/jobs), the states are listed as:
```
state: 'created' | 'retry' | 'active' | 'completed' | 'cancelled' | 'failed';
```

This means we can probably make a query for each of the states and do some math to get the number for each state *except* `failed`, because there's no state after it to specify in the `before` field.

Is there a way to get the number of failed jobs or is there another way to get queue numbers?

timgit:
You can write your own query against the table if you want.  There is also an event `monitor-states` that could be used, but this may not exist by the time v11 is released

pthieu:
@timgit does pgboss expose a method to query the db through its connection?
I'm using Drizzle, so I'd either have to hack around their raw query method or generate a schema for pgboss and use that.

timgit:
Seems like a good thing to add to the API. It's not there yet.

timgit:
I revisited this and remembered I expose this is already exposed via `getDb()`

stalniy:
Hey @timgit 

`getDb()` is not in types file. I can make a PR for this if you approve this change

timgit:
Approved.

PR #575, opened 2025-09-10
fix: expose getDb in types (issue #552)

danilofuchs:
Closes https://github.com/timgit/pg-boss/issues/552

Exposes `getDb()` to allow making queries in the same connection, without requiring a custom `Db` implementation

@timgit the Typescript definition for `Db` does not declare it as extending `EventEmitter`, nor includes `open()` and `close()`. Adding these to the typings may be a breaking change for people implementing custom `Db`s in the `options` argument

It's possible to expose a different interface for this specific implementation, and only depend on the base interface with `executeSql()` externally


timgit:
Thanks

Issue #546, opened 2025-02-20
Migrate to TypeScript - would PR(s) be welcomed?


murbanowicz:
As in title - I'd be happy to help with that and that would make typing valid as generated from code etc. Would make it easier to maintain and also use.

timgit:
As long as this happens after I merge the v11 branch. :)

timgit:
v11 is out.  I made changes to the types as best as I could, but I'm always aware that I could have missed some

bcomnes:
Seems as thought the name is required in the Queue type, which updateQueue and createQueue expect as their second argument, making a redundant name field.

bcomnes:
Fix: https://github.com/timgit/pg-boss/pull/600

PR #585, opened 2025-09-28
refactor: Switch to esm setup

boredland:
as a leap forward towards https://github.com/timgit/pg-boss/issues/546, I figured switching to esm and building the cjs version for backwards compatibility could be a good first step.

I did my best not to change anything relevant by accident, but sadly had to move away from standard.js to biome, because I didn't find a way to make standard (or standard-ts) support import attributes. I have an idea how to move back, that I'll try next to get back to it, just didn't want to spend to much time initially on some formatting setup.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 *.tgz
 *.log
 coverage
+dist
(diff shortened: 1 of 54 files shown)


boredland:
Just realized how bad my timing way. Will re-do :(

PR #588, opened 2025-09-28
refactor: Switch to esm setup

boredland:
contributes to https://github.com/timgit/pg-boss/issues/546

- uses tsup to build esm/cjs artifacts
- uses ts-standard to replace standard.js (imo it is a bit too strict and would recommend as an alternative)

This should IMO prepare everything to switch over to typescript in the near future, but incrementally.


bcomnes:
If looking to upgrade standard I really recommend switching to https://github.com/neostandard/neostandard instead.
ts-standard was a community contributed typescript addition to standard but it didn't keep in theme with the main project and has since became its own project. 

neostandard provides support for the latest eslint version, full eslint lsp support, and includes support for typescript out of the box, and is committed to the overal design and intention of the original standard project.  I'm working with feross to reconcile neostandard as the next version of standard before the end of the year, but its available to use now under the new name. 

In terms of cjs/esm, I don't really have a preference other than to recommend against dual package, as its not needed at this point. A single cjs or esm codebase works in either environment.

bcomnes:
This is what a neostandard migration would look like: https://github.com/timgit/pg-boss/pull/601 Should remain mostly out of conflict with the esm conversion done here. Biome is also an option but a bigger change overall.

boredland:
Should probably use "tsdown", as "tsup" seems deprecated. But will only touch this again, if some general feedback was there, as conflicts are already accumulating.

bcomnes:
I’ve been really happy with a type stripped workflow now that node supports it. It lets you develop without a build step, in typescript or js syntax (if that’s what’s desired here) and then when you publish, the artifacts are built, and then cleaned after publish.  https://github.com/bcomnes/ts-template/ is my personal template that uses tsc directly.

boredland:
> I’ve been really happy with a type stripped workflow now that node supports it. It lets you develop without a build step, in typescript or js syntax (if that’s what’s desired here) and then when you publish, the artifacts are built, and then cleaned after publish. https://github.com/bcomnes/ts-template/ is my personal template that uses tsc directly.

For my personal projects I do the same, but:

- no cjs support, which IMO makes it unnecessarily hard for some projects to upgrade
- no tree-shaking

In fact, having this as an option is the primary reason I added file-endings to all the imports. So we should still be able to switch it over to esm-only and a type-stripping approach in the future.

bcomnes:
Whatever direction this ends up going, I’d be sad to see pg-boss end up doing dual package (just pick one and keep it simple) or ship heavily compiled artifacts that are annoying to debug. Pg-boss being one of the last deps I can drop into in place in node_modules and debug is a real joy still.

boredland:
Sourcemaps exist and can be turned on with a flag in tsdown. But in principle I agree, as esm-only would be fine for me and the right path for typescript adoption anyway.

boredland:
no reaction.

Issue #593, opened 2025-10-01
deleteAllJobs('queue-name') deletes all jobs from all queues instead of the mentioned name


echamudi:
I was trying to delete all jobs from a queue, but pgboss decided to delete all rows.

```ts
  async deleteAllJobs (name) {
    Attorney.assertQueueName(name)
    const { table, partition } = await this.getQueueCache(name)

    if (partition) {
      const sql = plans.deleteAllJobs(this.config.schema, table)
      await this.db.executeSql(sql, [name])
    } else {
      const sql = plans.truncateTable(this.config.schema, table)
      await this.db.executeSql(sql)
    }
  }
```

It seems the code would delete all rows in `job_common` if partition is false.

leMedi:
I faced the same issue
and put a PR https://github.com/timgit/pg-boss/pull/595

PR #595, opened 2025-10-01
fix: delete jobs by queue name

leMedi:
bug: when calling `deleteAllJobs` on a non partitioned queue it will delete all jobs in job_common table 

Will help solve https://github.com/timgit/pg-boss/issues/593


timgit:
Thanks

PR #599, opened 2025-10-02
Fix #598 : empty queue stats

### LINKED ISSUE #598: getQueueStats() returns no value if no job in queue
`await pgBoss.getQueueStats(name)` returns null if no job has ever been queued

IMHO it should return the same datastructure than a case with jobs (with values set to zero)

pomarec:
It might be cleaner to set defaults values in the sql statement but i'm not sure how to make it properly

### DIFF
--- a/src/manager.js
+++ b/src/manager.js
@@ -654,7 +654,13 @@ class Manager extends EventEmitter {
 
     const { rows } = await this.db.executeSql(sql)
 
-    return rows.at(0) || null
+    return rows.at(0) || {
+      name,
+      deferredCount: 0,
+      queuedCount: 0,
+      activeCount: 0,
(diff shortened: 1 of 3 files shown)


src/manager.js
```diff
@@ -654,7 +654,13 @@ class Manager extends EventEmitter {
 
     const { rows } = await this.db.executeSql(sql)
 
-    return rows.at(0) || null
+    return rows.at(0) || {
```
timgit, line 657: I think this result should be merged into the result of `getQueueCache(name)` so it can overwrite cached values found.  This would also remove the need of introducing a new type.

timgit (changes requested):

PR #600, opened 2025-10-02
fix(types): add overloads for createQueue and updateQueue

bcomnes:
Added overloaded type signatures to support both calling patterns:
- createQueue(name, options) with Omit<Queue, 'name'>
- createQueue(options) with required name in Queue

This fixes type errors when name is passed as the first parameter, matching the actual implementation which supports both patterns.


bcomnes:
I added types for the still emitted 'stopped' event.

Issue #602, opened 2025-10-02
Database connection is not opened error when used with executeSql


bcomnes:
I'm getting:

```
"AssertionError [ERR_ASSERTION]: Database connection is not opened\n    at Manager.assertDb (./node_modules/pg-boss/src/manager.js:683:5)\n    at Manager.fetch (./node_modules/pg-boss/src/manager.js:429:21)\n    at Worker.fetch (./node_modules/pg-boss/src/manager.js:194:30)\n    at Worker.start (./node_modules/pg-boss/src/worker.js:49:33)\n    at Manager.watch (./node_modules/pg-boss/src/manager.js:228:12)\n    at Manager.work (./node_modules/pg-boss/src/manager.js:120:23)\n    at Timekeeper.start (./node_modules/pg-boss/src/timekeeper.js:52:24)\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n    at async PgBoss.start (./node_modules/pg-boss/src/index.js:131:7)\n    at async fp.dependencies (file://./packages/web/plugins/pgboss.js:39:3)
```

and

```
"AssertionError [ERR_ASSERTION]: Database connection is not opened\n    at Manager.assertDb (./node_modules/pg-boss/src/manager.js:683:5)\n    at Manager.fetch (./node_modules/pg-boss/src/manager.js:429:21)\n    at Worker.fetch (./node_modules/pg-boss/src/manager.js:194:30)\n    at Worker.start (./node_modules/pg-boss/src/worker.js:49:33)"
```

with the following options:

```
const boss = new PgBoss({
    ...defaultBossOptions,
    db: {
      executeSql: async (text, values) => {
        return await fastify.pg.query(text, values)
      }
    },
  })
```

where fastify.pg is an active node-pg pool. This was working in v10.

Any ideas or suggestions?

PR #603, opened 2025-10-02
Fix support for executeSql

bcomnes:
Potential fix for https://github.com/timgit/pg-boss/issues/602

The main issue being, _pgbdb or db.opened wont exist when utilizing executeSql, and thus this assertion fails.

### DIFF
--- a/src/manager.js
+++ b/src/manager.js
@@ -682,7 +682,7 @@ class Manager extends EventEmitter {
 
     // Custom db objects (without _pgbdb flag) are assumed to be ready
     // Only check opened flag for pg-boss managed db instances
-    assert(!this.db._pgbdb || this.db.opened, 'Database connection is not opened')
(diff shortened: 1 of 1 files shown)


src/manager.js
```diff
@@ -680,7 +680,9 @@ class Manager extends EventEmitter {
       return options.db
     }
 
-    assert(this.db._pgbdb && this.db.opened, 'Database connection is not opened')
+    // Custom db objects (without _pgbdb flag) are assumed to be ready
+    // Only check opened flag for pg-boss managed db instances
+    assert(!this.db._pgbdb || this.db.opened, 'Database connection is not opened')
```
timgit, line 685:
```suggestion
    assert(!this.db._pgbdb || (this.db._pgbdb && this.db.opened), 'Database connection is not opened')
```

timgit: I think this logic was intended

timgit (changes requested):

PR #617, opened 2025-10-23
feat : Added delete all jobs from all queues

pomarec:


timgit:
@pomarec I've included this in a new PR #652

PR #620, opened 2025-10-23
fix(types): have getMigrationPlans and getRollbackPlans accept version as a number

### LINKED ISSUE #619: getMigrationPlans and getRollbackPlans version args are typed as string but used as numbers
The typings for `getMigrationPlans` and `getRollbackPlans` specify their `version` args as `version?: string`:

https://github.com/timgit/pg-boss/blob/eaec51e4036992c5ca3879972dfe68418d2aaebf/types.d.ts#L224-L225

However, the implementation passes `version` as-is to `migrationStore`'s `migrate` and `rollback`, where they are compared with `===` and `>=` to the `number` version fields encoded in the migration definitions' `version` and `previous` fields:

https://github.com/timgit/pg-boss/blob/eaec51e4036992c5ca3879972dfe68418d2aaebf/src/migrationStore.js#L21

https://github.com/timgit/pg-boss/blob/eaec51e4036992c5ca3879972dfe68418d2aaebf/src/migrationStore.js#L52

We're working around this with `PgBoss.getRollbackPlans(schema, version /* a number */ as unknown as string)`, but it'd be nice to fix the typings.

dbjorge:
closes #619

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -221,8 +221,8 @@ declare class PgBoss extends EventEmitter {
   constructor (options: PgBoss.ConstructorOptions)
 
   static getConstructionPlans (schema?: string): string
-  static getMigrationPlans (schema?: string, version?: string): string
-  static getRollbackPlans (schema?: string, version?: string): string
+  static getMigrationPlans (schema?: string, version?: number): string
+  static getRollbackPlans (schema?: string, version?: number): string
 
   static states: PgBoss.JobStates
   static policies: PgBoss.QueuePolicies


dbjorge:
(I'd consider this a `fix` rather than a breaking `fix!` since anyone that would be impacted by this is already broken, but feel free to adjust the PR title and save it for v12 if you'd prefer to treat it as breaking)

bcomnes (approved):

timgit:
This will be included in the v12 release that migrates the codebase to typescript

PR #623, opened 2025-10-26
 Feat: Migrate codebase to ESM and TypeScript

### LINKED ISSUE #621: Wrong typing for boss.complete overloads
These `.complete()` typings overloads are wrong, since the code always uses the third arg as data and the fourh as options.

https://github.com/timgit/pg-boss/blob/eaec51e4036992c5ca3879972dfe68418d2aaebf/types.d.ts#L293-L295

This is the implementation:
https://github.com/timgit/pg-boss/blob/eaec51e4036992c5ca3879972dfe68418d2aaebf/src/manager.js#L488-L496

davbrito:
### Key Changes:

* **ESM & TypeScript Conversion:** All core .js files have been converted to .ts.
* **Build Configuration:** I configured the build to use [tsdown](https://tsdown.dev/) for a fast and straightforward compilation process.
* **API Refactoring for Type Safety:** The previous "mixin" pattern (using `.functions` property and `.#promoteFunctions`) has been replaced. Functionality is now exposed via explicit wrapper methods on the `PgBoss` class. This refactoring was essential for providing accurate type definitions.
* **API Cleanup**: As part of the migration, some method overloads were simplified, and others were corrected to better align with the actual runtime API. All this without making changes to the API.

**Notes:**

**Preserved Functionality:** All existing functionality remains unchanged. This PR focuses on the migration and internal refactoring for type safety.

**Test Suite:** All tests continue to pass. The test files themselves remain in JavaScript for now and can be progressively migrated in future PRs.

**Known Type Issues:** You may notice a few remaining type errors when running `tsc`. These where left "as is" to not mess with implementation details and can be addressed in follow-up work.

**Breaking Changes:** This change might be published as a new major version. The way types are exported has changed (they were previously inside a namespace), which is a breaking change for TypeScript users.

fixes #621, #546

### DIFF
--- a/test/configTest.js
+++ b/test/configTest.js
@@ -1,6 +1,8 @@
 import assert from 'node:assert'
+import Db from '../src/db.ts'
 import PgBoss from '../src/index.ts'
 import * as helper from './testHelper.js'
+import versionJson from '../version.json' with { type: 'json' }
 
 describe('config', function () {
   it('should allow a 50 character custom schema name', async function () {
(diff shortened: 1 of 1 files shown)


bcomnes:
+1 to esm and types. My 2c would be to skip the complexity nightmare of dual package publishing and consumption and either go esm only or cjs only (if there is a desire to continue supporting eol node for whatever reason). Both are fully compatible in both environments at this point. I would also consider just using tsc to keep things vanilla as possible and set up the on-ramp to tsgo when that comes out.

davbrito:
> I would also consider just using tsc to keep things vanilla as possible and set up the on-ramp to tsgo when that comes out.

I believe tsdown is a good choice. It's simple enough, fast, and it enables bundling, which reduces package size and generates clean type declarations (`.d.ts`) that only expose the public API, removing internal types and dependencies.

timgit:
Thanks for submitting.  Reviewing.  Looks like `.npmignore` is obsolete now that files is a thing.

boredland:
Would be amazing to get this landed! 🚀

timgit:
I published this to 12.0.0 under the beta tag.  Please review and confirm.

PR #624, opened 2025-10-28
Add generic return type to WorkHandler and work method

bcomnes:
This allows one to set the expected return type on a WorkHandler function so the return value can be typed checked too.


timgit:
Please reopen after back merging from latest

bcomnes:
Sorry I was traveling all last week and then got sick on the return home. Will rebase and refresh all open PRs that need it this week.

bcomnes:
Ok, I rebased this PR, should be ready to land.

PR #625, opened 2025-10-28
Fix Omit type usage and startAfter types

bcomnes:
A second argument with a type union is instead of variadic arguments is the correct way to omit type fields with Omit.
I also fixed startAfter types which accept a number in both callsites.


timgit:
This will be included in the v12 release that migrates the codebase to typescript and esm

Issue #618, opened 2025-10-23
Error emitted at runtime with a custom db


sylvaingi:
After integrating v11 with my DB client, Kysely, using the `db` option, I started seeing the following errors emitted by pg-boss:

```
TypeError: results.flatMap is not a function
    at #monitor (/Users/sylvain/work/anct/mec/node_modules/.pnpm/pg-boss@11.1.1/node_modules/pg-boss/src/boss.js:137:29)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async Boss.supervise (/Users/sylvain/work/anct/mec/node_modules/.pnpm/pg-boss@11.1.1/node_modules/pg-boss/src/boss.js:119:9)
    at async #onSupervise (/Users/sylvain/work/anct/mec/node_modules/.pnpm/pg-boss@11.1.1/node_modules/pg-boss/src/boss.js:85:27)
```

After some debugging, I found that `Boss.#monitor()` expects the cacheQueueStats query to return a result shaped like `{ rows: any[] }[]`, which matches what `pg.Pool.execute()` returns.

However, when using a custom executeQuery, the result shape is `{ rows: any[] }`, so we can’t replicate the same behavior.

Maybe `Boss.#monitor()` could check whether it receives an array of results or a single result, so we wouldn’t need to change the executeQuery signature?

timgit:
Good catch.  Did you confirm that Kysley returns the right array in `rows`?

sylvaingi:
At first I was seeing empty arrays, I had to split the query using `;` and execute each nested query individually. Then i was seeing batches of results like

```ts
[
  { pgAdvisoryXactLock: '' },
  { name: 'test', queuedCount: 4, warningQueueSize: 0 }
]
```

Still had to silence TS due to the typing issue but it works fine for now.

swymmwys:
encountered the same problem using Kysely, would you accept a PR to fix this?

timgit:
Yep!

swymmwys:
https://github.com/timgit/pg-boss/pull/626

callumsteele4:
Came here to report this and love that it's already fixed! How often do you cut releases @timgit?

timgit:
This was just released in 11.1.2.  Thanks for the nudge

PR #626, opened 2025-10-30
fix: unwrap sql result in case it returns array during supervise

swymmwys:


swymmwys:
fixes https://github.com/timgit/pg-boss/issues/618

Issue #605, opened 2025-10-03
Option to disable automatic schema creation when migrate: true


instilled:
First of all, thank you for this great library. I love it and have been using it for quite a while.

Currently, when migrate: true is set, pg-boss will automatically create the schema if it does not exist. While this is convenient in simple setups, it causes challenges in more complex deployment environments where database schema provisioning is managed externally (e.g., by migrations tooling, DBA-managed processes, or infrastructure-as-code pipelines).

### Problem
In many enterprise or production environments:

- Schema creation is tightly controlled and should not be performed automatically by application code.
- Database users may not even have CREATE SCHEMA permissions, causing failures at runtime.
- Schema lifecycle (creation, grants, ownership, migrations) is typically managed by separate tooling and audited processes.

In such cases, the automatic schema creation behavior can be undesirable or even break deployments.

### Proposed solution
Introduce a configuration option (e.g., allowSchemaCreate: false) that:
- Still allows pg-boss to run migrations (when migrate: true) if the schema already exists.
- Skips any attempt to create a schema if it does not exist.
- Fails fast with a clear error if the schema is missing and allowSchemaCreate: false is set.

### Benefits
- Better fit for organizations with strict DB provisioning practices.
- Clear separation of responsibilities between infrastructure setup and application runtime.
- Maintains backward compatibility for those who want the current behavior by default.

### Current workaround

```ts
export class PgBossBackend {

  private readonly pgBoss!: PgBoss;
  private _started = false;

  ...

  async start(): Promise<void> {
    const coordinatorPool = ... // some connection creation fn
    const dbOptions: ConstructorOptions = {
      db: {
        executeSql: async (text: string, values: any[]) => {
          if (!this._started) {
            const createSchemaRegex = /CREATE SCHEMA IF NOT EXISTS.*/;
            const match = text.match(createSchemaRegex);
            if (match) {
              text = text.replace(createSchemaRegex, "");
            }
          }
          return await coordinatorPool.query(text, values);
        },
      },
      schema: this.config.database.coordinator.schema,
      migrate: true,
    };
...
```

Happy to provide a PR if this is something of interest.

timgit:
There is already an API created specifically for this reason: https://timgit.github.io/pg-boss/#/./api/utils

instilled:
> There is already an API created specifically for this reason: https://timgit.github.io/pg-boss/#/./api/utils

Thanks for pointing it out. I hadn’t realized that.

Looking at the code:

[...]

PR #628, opened 2025-11-05
feat: optionally disabling schema creation

instilled:
This solves #605. It also includes a resolution for a minor `Omit` syntax bug in `types.d.ts` (can also undo).


timgit:
I wanted to merge this before the ESM PR, but it was not meant to be, unfortunately.  Would you re-submit this after back merging?  The types issue was resolved, so it would just be your changes applied to the new typescript files.  Thanks!

instilled:
@timgit done!

Issue #627, opened 2025-10-31
`deleteAllJobs`: Cannot read properties of null


boredland:
I am trying to delete queues that contain jobs, and therefor figured I first needed to get rid of the jobs. Sadly, these error for me:

```typescript
await pgBoss.deleteAllJobs('1.balances.r1');
```

```
TypeError: Cannot read properties of null (reading '1.balances.r1')
     at Manager.getQueueCache (/app/node_modules/pg-boss/src/manager.js:80:28)
     at Manager.deleteAllJobs (/app/node_modules/pg-boss/src/manager.js:643:45)
     at <computed> [as deleteAllJobs] (/app/node_modules/pg-boss/src/index.js:103:43)
```

Ideas what might be going on?

timgit:
Queues can't have periods in the name.

boredland:
But... I created them and used them like this?

timgit:
createQueue will throw if you give it that name. What version are you running?

boredland:
currently running 11.1.1

boredland:
... and I definitely created them also on the same version.

davbrito:
> Ideas what might be going on?

You should call `await pgBoss.start()` before doing anithing with queues.

davbrito:
In #623 I added an assertion that shows a meaningful message if queues are not initialized.
https://github.com/timgit/pg-boss/blob/584b89b4a77947c5057794b4d3e47a023a93704e/src/manager.ts#L58-L61

PR #629, opened 2025-11-05
fix rejection tests and queue name assertion

davbrito:
This PR addresses an issue where queue name assertion tests were not correctly failing due to a faulty try/catch pattern. (see #627 and https://github.com/timgit/pg-boss/issues/627#issuecomment-3476458782).

This pattern actually makes the test to never fail:

https://github.com/timgit/pg-boss/blob/6c1450a1b7fa08b92ab658635e2126484bbe8bb5/test/queueTest.js#L25-L30

The tests now use the dedicated Node.js `assert.rejects` and `assert.throws` methods, ensuring that validation failures are properly caught and reported.

This also fixes the regex used to validate queue names. I added a new `randomQueueName` since some tests were using uuids as queue names, but UUIDs may start with a number.

### DIFF
--- a/src/attorney.ts
+++ b/src/attorney.ts
@@ -10,7 +10,7 @@ const POLICY = {
 
 function validateQueueArgs (config: any = {}) {
   assert(!('deadLetter' in config) || config.deadLetter === null || (typeof config.deadLetter === 'string'), 'deadLetter must be a string')
-  assert(!('deadLetter' in config) || config.deadLetter === null || /[\w-]/.test(config.deadLetter), 'deadLetter can only contain alphanumeric characters, underscores, or hyphens')
+  assert(!('deadLetter' in config) || config.deadLetter === null || /^[\w-]+$/.test(config.deadLetter), 'deadLetter can only contain alphanumeric characters, underscores, or hyphens')
 
   validateRetryConfig(config)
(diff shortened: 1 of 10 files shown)


src/attorney.ts
```diff
@@ -10,7 +10,7 @@ const POLICY = {
 
 function validateQueueArgs (config: any = {}) {
   assert(!('deadLetter' in config) || config.deadLetter === null || (typeof config.deadLetter === 'string'), 'deadLetter must be a string')
-  assert(!('deadLetter' in config) || config.deadLetter === null || /[\w-]/.test(config.deadLetter), 'deadLetter can only contain alphanumeric characters, underscores, or hyphens')
+  assert(!('deadLetter' in config) || config.deadLetter === null || /^[\w-]+$/.test(config.deadLetter), 'deadLetter can only contain alphanumeric characters, underscores, or hyphens')
```
boredland: As I understand this, non-usage of simple `.` periods is enforced. Sadly, I actually have many queues with them being present. Is this really necessary? If so: how do I migrate away?

boredland: also: wouldn't it make sense to define this regex in one place instead of 3?

davbrito:
Closing in favor of #630

PR #630, opened 2025-11-07
ESM + Typescript

timgit:


timgit:
@davbrito and @bcomnes, this PR contains the ESM rewrite PR + Typescript tests.  Asking for your opinions on the index.cjs pattern I used here (and anything else).  This pattern seems to be unpopular, but I was wanting a more clean way to require(esm) than what typescript is currently producing during transpile where `export default boss` becomes `export.default = boss` instead of module.exports = boss like it used to be for `require()`.  Obviously this pattern has a shelf life once cjs become extinct, but it at least avoid dual package publishing.  I bumped up the minimum node version to 22.12 for require(esm) as well.

bcomnes:
I can take a peek tomorrow.

davbrito:
> This pattern seems to be unpopular, but I was wanting a more clean way to require(esm) than what typescript is currently producing during transpile where `export default boss` becomes `export.default = boss` instead of module.exports = boss like it used to be for `require()`

It seems it could be achieved with a single default export: https://tsdown.dev/options/cjs-default

I tried removing these lines:
image

image

And had this cjs output thanks to `cjsDefault` option:
image

---

A **different approach** would be to get rid of the default export and just use named exports. This way it can be published as esm only, cjs users will just need a node version compatible with `require(esm)`.

image

davbrito:
Here is an example how would it be with esm only & named exports only:

https://github.com/davbrito/pg-boss/commit/fef79d4c4c6371aa2cabe190c78112471965a51d

timgit:
Thanks for the feedback.  I'm leaning towards consolidating to named exports and abandoning default

bcomnes:
I’m plus one on not supporting default exports and avoiding a cjs wrapper. In theory it works well in modern node but things get murky in tortured environments like next.js or deno/bun node comparability layers.

src/types.ts
```diff
@@ -215,6 +221,7 @@ export interface JobWithMetadata<T = object> extends Job<T> {
 
 export interface JobInsert<T = object> {
   id?: string;
+  name: string;
```
davbrito: This `name` property is not used, and causes type errors. Queue name is passed on a separate parameter:
https://github.com/timgit/pg-boss/blob/2c9009fb00a4c28e11cdf724c3d772c143ef59e5/src/manager.ts#L357

test/queueTest.ts
```diff
@@ -1,63 +1,51 @@
...
-    })
-  })
-
-  it('should reject a queue that starts with a number', async function () {
-    this.boss = await helper.start({ ...this.bossConfig, noDefault: true }) as PgBoss
-    const queue = `4${this.bossConfig.schema}`
-
-    assert.rejects(async () => {
+    await assert.rejects(async () => {
+      this.boss = await helper.start({ ...this.bossConfig, noDefault: true })
```
davbrito: You should keep inside the assertion only the thing being tested (in this case the createQueue method).

If for some reason, .start fails, the test would keep passing, even if it's not working the way it's desired.

PR #636, opened 2025-11-16
Update ops.md

lexuzieel:
Removed the 'wait' option from the arguments section and added a note about the 'graceful' option.


timgit:
Thanks

PR #638, opened 2025-11-17
Bundle serliaze-error so that CJS projects can work in edge cases

baudehlo:
This PR makes pg-boss available as both an ESM module and a CJS module depending on which you need.

Rationale: I've had a ton of problems with pg-boss in jest and nest.js and tried every which way to fix it, but the ultimate issue is that CJS projects don't play well with ESM modules.

This is a simple fix to that problem.

### DIFF
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -5,6 +5,5 @@ export default defineConfig({
   format: 'es',
   dts: true,
   exports: true,
-  clean: false,
   noExternal: ['serialize-error'],
 })


baudehlo:
I've converted this to draft as I tried using it and am still getting the same errors due to the use of ESM-only serialize-error.

timgit:
What version of Node are you using?

baudehlo:
24.2.0.

The problems stem from the fact that serialize-error is also ESM-only, and
so jest can cope with one level deep of ESM, but not two, it seems. I've
tried just about everything. So far the only thing that truly worked was
installing my branch, then removing pg-boss/node_modules/serialize-error
and installing ***@***.***

I'll keep poking at it, as converting a nestjs project to ESM is apparently
an almost impossible task. My next step is to see if I can just compile
serialize-error inline like a browser distro.

baudehlo:
OK this version works. It bundles serialize-error.

timgit:
Out of curiosity, would including serialize-error in the esm build allow your code to use require(esm)?

baudehlo:
Is that your preference? I can certainly try that.

Is the aim to stick with being ESM only?

timgit:
Yes

baudehlo:
Since the project isn't in JS any more, but simply transpiled to it, mind sharing why you care, for my curiosity?

baudehlo:
I've updated it to your preference and confirmed it also works.

timgit:
It's not a high priority but just a preference to keep the package size small and simple

timgit:
Why does the clean option need to be set to false?

baudehlo:
copy-pasta. Let me try without tomorrow.

baudehlo:
Confirmed it works fine with clean left alone.

bcomnes:
> jest can cope with one level deep of ESM

This seems like more of a jest issue (among many, many other jest issues) than a pg boss issue.  Was there ever a reproduction or detailing of the error? There is likely better solutions here than for pg-boss to take on cruft like this on behalf of jest.

bcomnes:
The other thing here is that `serialize-error` should be moved to devDependencies since it's no longer a runtime dependency when it is bundled... But I'm getting strong 'wrong solution' vibes on this one.

baudehlo:
The issue came back with 12.3.1 anyway so I ended up switching to vitest.

I would say that Jest is everywhere, and it's a bit shitty to force ESM
only on people who want to just use pg-boss in their project. A lot of
times you don't get to just switch out testing environments, especially if
you work in a company. And especially when the fix was trivial to add in
CJS output too - it's not like it's hurting anyone to do that. But that's
my 2c and I don't own this project. If pg-boss were written in ESM I would
accept it as normal, but it's written in Typescript and transpiled.

bcomnes:
I see what the issue was:

`import { serializeError } from "../node_modules/serialize-error/index.mjs";`

https://app.unpkg.com/pg-boss@12.3.0/files/dist/src/manager.mjs#L3

This is just incorrect output that will interfere with the module resolution algorithm in node or other node-like environments. It's generated by `tsdown` and would also explain why a less processed previously published version of the module was having difficulties in wacky environments like the bun compatibility layer or whatever whackery jest is doing. 

(By including `../node_modules/` in the import path to an external dependency, the build bakes in external resolution structure inside the module, which will break in many scenarios. It it is never safe to directly resolve to files external to the module boundary. In this case, this appears to be a defect generated by tsdown for some reason, as the source code appears correct).  

It should have generated something like:

`import { serializeError } from "serialize-error"` as written in: https://github.com/timgit/pg-boss/blob/a379f7ec433a5cbfa1a7aa88a8f71035cf60d223/src/manager.ts#L4

Jest definitely CAN cope with more than one level deep of ESM, assuming published artifacts are defect free.

I have no idea why tsdown is doing this, could be a bug or config error, would require more research. 

@timgit would you be open to backing away from the bundled artifacts in favor of

a) a fixed `tsdown` build that isn't generating incorrect output, but also not bundling.
b) a `tsc` publish time build where correctness is far less of a question and eliminates a devDependency

If you just prefer to bundle everything thats totally fair as well, but it could just be papering over deeper configuration issues or bugs in `tsdown` and isn't without its own set of downstream drawbacks.

bcomnes:
>  it's a bit shitty to force ESM

This isn't about forcing esm, I'm trying to get to the bottom of the issue here.

bcomnes:
Also sorry to bag on jest, it was not the issue here.

timgit:
My first reaction to the Typescript + ESM PR was "yeah, it's 2025, we've got require(esm), let's do this!".  It was unfortunate to run into bundling and module resolution issues after getting cjs + esm tests working within the repo.  Alas, this repo isn't agressive enough to try and verify all runtimes, so it seems in retrospect that cjs is still king of the compatibility game.

Having acknowledged all that, my preferences were to ship the simplest output possible, which was why I attempted tsdown's Unbundle mode.  I'm not yet convinced to embrace minification or sourcemaps, especially when troubleshooting issues via live debugging, for example.  

@bcomnes it seems I'm in favor of what you're proposing unless I'm misunderstanding you.

bcomnes:
I’ll get you and an and a b PR asap. Upon more research you were running into tsdown bugs. Sent from my mobileOn Nov 25, 2025, at 12:12 PM, Tim Jones ***@***.***> wrote:﻿timgit left a comment (timgit/pg-boss#638)
My first reaction to the Typescript + ESM PR was "yeah, it's 2025, we've got require(esm), let's do this!".  It was unfortunate to run into bundling and module resolution issues after getting cjs + esm tests working within the repo.  Alas, this repo isn't agressive enough to try and verify all runtimes, so it seems in retrospect that cjs is still king of the compatibility game.
Having acknowledged all that, my preferences were to ship the simplest output possible, which was why I attempted tsdown's Unbundle mode.  I'm not yet convinced to embrace minification or sourcemaps, especially when troubleshooting issues via live debugging, for example.
@bcomnes it seems I'm in favor of what you're proposing unless I'm misunderstanding you.

baudehlo:
Thanks for keeping this going - by force ESM I just meant that when there's
an easy way to build and ship both it seems like the easy choice.

I can highly recommend people reading this thread looking for a solution
make the switch to vitest though - so far it's much better (faster) than
jest.

timgit:
I agree as well.  Dual shipping cjs and esm is a perfectly fine option if that provides the widest support atm.

bcomnes:
I apologize, I was slightly off target here:

The combo of unbundle: true and noExternal causes serliaze-error to bundle into a published [nested node_modules](https://app.unpkg.com/pg-boss@12.3.0/files/dist/node_modules) NOT next to an adjacent package.json (part of package.json ends up in a package.mjs as well, not sure what this is about). Typically modules publishing customizations to node_modules would use something like bundleDeps, and it seems like it should work because of the relative import into it. However, running this code through bundlers, or bun resolution algos, the chances of something not working gets higher and higher. 

My reply got garbled earlier but I'm working on some ideas I will PR asap.

bcomnes:
Looking at the failing example in https://github.com/timgit/pg-boss/issues/645

I see that they use some kind of bundler thing to build out their runtime code out of node_modules (unrelated to bun actually). Because of the non-standard arrangement inside of pg-boss@12.3.0 generated by tsdown, the bundled deps get stripped for their build output:

Screenshot 2025-11-25 at 7 03 25 PM

bcomnes:
wrt to the jest case: https://github.com/jestjs/jest/issues/15275

Jest doesn't support `require(esm)` yet. That sucks, but also fairly representative of the state of that project (you should switch off if you can). Investigating work arounds. 

Dual package is probably the worst options out of all the options. Fully bundling or shipping cjs-only far outweigh dual package in terms of simplicity, reliability, and compatibility in nearly every case. (This is on top of nearly every package publishing dual package screwing something up especially around types).

Pure esm with working `require(esm)` is really the way out of this mess and I think I can get it working for most cases. Please bear with me.

bcomnes:
https://gist.github.com/bcomnes/df5dc907a9766f4c8dc581b86c34e317

Here is a demonstration of how to get pg-boss@12 working in jest.

PR #640, opened 2025-11-19
Backport: prevent singleton/stately batch double-activation on 10.3.x

nrempel:
This is a small 10.3.x maintenance backport for the singleton/stately batch-activation bug described in #535. The fix matches the 11.0.8 SQL change while keeping the 10.x parameterized fetch shape, and we backported the upstream regression test to prevent double promotion of the same singletonKey in a batch and added a couple more.

- Fixes the singleton/stately batch activation bug where a batch fetch could promote two jobs with the same singletonKey (e.g., a retried job plus a newly queued one) when batchSize > 1.
- Mirrors the upstream 11.0.8 change (commit 5a12df77627f22b4d73520e425dfe0d02db9ee5a) by adding a grouped CTE with row_number() partitioned on singleton_key for singleton/stately queues; only row_number = 1 is promoted in each fetch. Other policies are unchanged and SQL remains parameterized for 10.x.
- Adds regression coverage:
    - Stately: ensure only one of two same-key jobs is activated per batch, the other activates on the next fetch.
    - Singleton: ensure a batch returns at most one per singletonKey while still filling with other keys.
    - Stately mixed keys: ensure retry + different key can co-activate in the same batch (no over-filtering).

Issue: https://github.com/timgit/pg-boss/issues/535 (“singleton key active bypass with batch > 1”)
Upstream fix/release: https://github.com/timgit/pg-boss/releases/tag/11.0.8 (commit 5a12df77627f22b4d73520e425dfe0d02db9ee5a)

### DIFF
--- a/src/plans.js
+++ b/src/plans.js
@@ -506,7 +506,7 @@ function insertVersion (schema, version) {
 function fetchNextJob (schema) {
   return ({ includeMetadata, priority = true, ignoreStartAfter = false } = {}) => `
     WITH next as (
-      SELECT id
+      SELECT id, singleton_key, policy
       FROM ${schema}.job
       WHERE name = $1
         AND state < '${JOB_STATES.active}'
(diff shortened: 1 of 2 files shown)


nrempel:
This should cherry pick cleanly onto 10.3.3. I think github only allows merging to branches, not tags.

timgit:
Thanks. What are your upgrade plans for the latest version?

nrempel:
We don't have immediate upgrade plans as the migration path seems non-trivial. But this fixes our problem in the immediate term without the need for migration work. (We've been bitten by https://github.com/timgit/pg-boss/issues/535)

nrempel:
Thanks for the quick release!

Issue #645, opened 2025-11-24
Import error when starting my app with pg-boss 12.x


scesbron:
I have an app based on solidstart that works well with pg-boss 11.1.2.

I tried to update it to version 12.x but when I start my app I have an import error:

```
ERROR  Cannot find module '/home/sebastien/dev/solid-pgboss/.output/server/node_modules/pg-boss/dist/node_modules/serialize-error/index.mjs' imported from /home/sebastien/dev/solid-pgboss/.output/server/node_modules/pg-boss/dist/src/manager.mjs                                                       12:20:19

    at finalizeResolution (node:internal/modules/esm/resolve:265:11)
    at moduleResolve (node:internal/modules/esm/resolve:933:10)
    at defaultResolve (node:internal/modules/esm/resolve:1169:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:540:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:509:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:239:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:96:40)
    at link (node:internal/modules/esm/module_job:95:36) 

```

Here is a simple project that reproduces the error: https://github.com/scesbron/solid-pgboss

timgit:
I've not tested with this tech stack before, which I'm assuming is the root cause.  I cloned your code but I was not able to get as far as the error you included.  It does match [this issue](https://github.com/nksaraf/vinxi/issues/177) in the vinxi repo, however.  Let me know what you find out.

scesbron:
On my side I did not find out the root cause of this problem. I reverted my upgrade to stay with pg-boss 11.x

Solidstart is currently working on getting rid of vinxi so I may wait for a version without vinxi and try again to see if the problem is gone.

timgit:
Out of curiosity, can you try with 12.2.0 instead of 12.3.0?  All deps were included in a bundle file in that release.  I changed this in 12.3.0 to evaluate a more simple dist format.

scesbron:
On my test app it works well with version 12.2.0 instead of 12.3.0 so I think it is the new bundle that does not work correctly with vinxi.

timgit:
I just pushed 12.3.1 to switch back to full bundling mode in tsdown.  There's a [solidjs compat page on tsdown](https://tsdown.dev/recipes/solid-support) but I didn't go down that rabbit hole. :)

PR #647, opened 2025-11-26
Make unbundle: true work

bcomnes:
This PR is one option to address the following errors

- https://github.com/timgit/pg-boss/pull/638 - jest not supporting require(esm)
- https://github.com/timgit/pg-boss/issues/645 - pg-boss missing `tsdown` unbundled deps in bundled vinxi output.

It seemed like there was some interest in trying to get the `unbundle` in `tsdown` working, where `tsdown` strips types and retains the overall source structure of the output. 

By removing `noExternal: ['serialize-error'],` but keeping `unbundle: true` tsdown type strips ts syntax without making one giant bundle which provides better stack traces to callers and makes applying downstream patches easier and more straight forward. It also removes some undefined behavior around output artifacts allowing code bundled by `vinxi` and others to work.

It seems as though the point of `tsdown` IS to bundle the published artifact and we are still outputting esm only either way, so it makes me wonder if just sticking with the bundled artifact is the preferable arrangement when using `tsdown`. Jest still requires downstream workaround either way, and un-bundling `serialize-error` and letting the jest callers figure that one out on their own is the better solution if we are sticking with esm only.  I offered some solutions in the jest thread related to that issue. 

I have another solution I will propose in another PR.

### DIFF
--- a/tsdown.config.ts
+++ b/tsdown.config.ts
@@ -5,6 +5,5 @@ export default defineConfig({
   format: 'es',
   dts: true,
   exports: true,
-  noExternal: ['serialize-error'],
   unbundle: true
 })


bcomnes:
Alternative PR here: https://github.com/timgit/pg-boss/pull/648

timgit:
Pretty sure I will go with the tsc option after all this tsdown drama.  Thanks!

PR #648, opened 2025-11-26
Build typescript with vanilla tsc

bcomnes:
This PR is the OTHER option to address the following errors (the first was here https://github.com/timgit/pg-boss/pull/647)

https://github.com/timgit/pg-boss/pull/638 - jest not supporting require(esm)
https://github.com/timgit/pg-boss/issues/645 - pg-boss missing tsdown unbundled deps in bundled vinxi output.

I proposed this solution earlier in the conversion process to ts/esm but let it rest once things settled on `tsdown`, but given the confusion and defects around the output of that tool, I figured a more concrete proposal would at least illustrate what I was suggesting better.

### This PR:

 - Drops `tsdown` and instead just builds using vanilla typescript's tsc
 - Fixes a few under-defined tsconfig settings around node type stripping which are in use (needed either way)
 - Adds a new `clean` and `build` script that is run only during the publish process (or if you want to debug output before publishing)
   - The build script builds in place and the clean script cleans out the artifacts. The difference between the source artifacts and the build artifacts is dramatically minimized compared to `tsdown` artifacts which moves things around in unexpected ways.
   - Example build can be seen here: https://app.unpkg.com/@bret/pg-boss@12.3.3/files/src
 - Adds source map support (`tsdown` I'm sure has this, but it wasn't on for some reason).
 - Simplifies `package.json` entry points to only use `main` (none of the fancy export map features were in use, and this dramatically simplifies module resolution, lmk if you want exports back if you prefer closed exports)
 - A dramatic reduction in dev time transitive dependencies including large native binaries that are impossible to inspect (see the lock file)
- `tsconfig.json` is used by your lsp and `tsc` for type checking and doesn't emit.
- `tsconfig.declaration.json` is used during the build step and inherits from `tsconfig.json`. It emits!

### Why this is better (please excuse the rhetoric hat, I'm trying to sell you!)

- Pretty much the way the module was before, but with ts and esm.
- A large complex build tool is completely eliminated, so now we only have ONE (tsc) large complex build tool instead of 2.
- In-place builds are closer to their source artifacts, easier to debug inside of node_modules (less indirection) and give you stack traces that are more helpful to upstreaming fixes. 
- One less complex config schema
- Will be easier to update to tsgo once that lands.
- The entire concept of bundling, and all the complexity and confusion around how or why or when to do that correctly is eliminated by eliminating the bundler (a factor in the above two issues). It's really not necessary here and actually makes it harder/messier for dependents to bundle deps in cases where you actually need to bundle. 
- Output is more well behaved. It will never generate a `node_modules` in the published tarball, lacking an adjacent `package.json` or `package.json` inside the nested deps (malformed node_modules). It will never write out a `package.mjs`, whatever that is. It doesn't over rely on `.mjs` or `.cjs` (over used escape hatches) and has the correct use of the `type` field in package.json and adjacent file extensions. 
- I can go on!


bcomnes:
You should also feel free to close this! Take the option that you feel best matches your goals/vision for the project.

timgit:
I really appreciate your assistance with this.  I will open a branch to merge this PR so I can kick it around with different options. I hate to bring up Windows, but some poor souls out there are forced to use it so I'll probably go with a clean package like rimraf over bash scripts for that part of the build.

bcomnes:
Should work fine with wsl but if you want to cover someone trying to fire this up in cmd or posh you could swap to rimraf.

PR #650, opened 2025-11-26
package npm package with tsc instead of tsdown

timgit:


timgit:
@bcomnes I made some changes to your suggestions. I think I've simplified the configs and settings based on the inheritance of the 2 base configs.

bcomnes:
Lgtm

bcomnes:
I thought about adding the strictest base config as well but decided to suggest that separately depending on how things shake out.

PR #655, opened 2025-11-29
Use spies for faster test execution

kibertoad:
1 min 31 seconds -> 1 min 23 seconds.

Most of other tests seem to be fairly timing-sensitive, it is possible to convert them to spy usage, but that would harm the clarity of what and how they are testing. I think they are better left as-is.


kibertoad:
I think these are the only tests that are clearly safe to convert, but please let me know if I missed something significant.

PR #673, opened 2026-01-08
Abort jobs after graceful shutdown period expires

bcomnes:
When jobs exceed their maxExpiration, the abort signal passed into the job will abort. Similarly, when pg boss shuts down, it will wait for a grace period before marking the wip jobs as failed. The intention here as well is: times up! Time to stop working if you are still working. 

Currently, the abort controller in the job does not notify of shutdown (only maxExpiration), but it should, because we are going to fail the job and it might start up somewhere else, and the worker should stop working (if it still is)!

This adds a slot on workers to keep track of the currently active job abortController. In the manager, a workerRef is used to populate and clear the currently active abortController onto the worker, as it fetches jobs. When failWip is called, any workers with abortControllers are aborted, right before the job is marked as failed.

### DIFF
--- a/src/manager.ts
+++ b/src/manager.ts
@@ -201,12 +201,18 @@ class Manager extends EventEmitter implements types.EventsMixin {
   async #processJobs<T> (
     name: string,
     jobs: types.Job<T>[],
-    callback: types.WorkHandler<T>
+    callback: types.WorkHandler<T>,
+    worker?: Worker<T>
   ): Promise<void> {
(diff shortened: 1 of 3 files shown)


bcomnes:
Rebased after fixing the conflict on https://github.com/timgit/pg-boss/pull/678 and added a test with concurrency. Also fixed a race condition on outcome: fail the job (as we were before), then send the abort signal (stop working, if you are listening).

bcomnes:
Rebased cleanly. Should be an easy land!

PR #676, opened 2026-01-09
feat: allow returning ids with spies disabled

JuanoD:
`boss.insert` doesn't return job IDs unless spies are enabled. Not sure why is not true by default, but to keep that behavior, it can now be enabled with an option


bhamiltoncx:
I just ran into this bug myself. It's hard to use `insert()` if you can't get the resulting job IDs, because you can't then deal with cancelling them in the future if needed.

PR #679, opened 2026-01-11
Improve distributed database support

### LINKED ISSUE #669: Better distributed database support
There are certain caveats involved in using SKIP LOCKED in databases like CockroachDB (which are technically PostgreSQL-compatible).

SKIP LOCKED, while technically supported there, does not perform well in clustered distributed databases.

There are open issues showing SKIP LOCKED performs poorly under high concurrency. In one test with 500 workers, the mean latency for SELECT FOR UPDATE SKIP LOCKED was 311ms - far higher than expected. https://github.com/cockroachdb/cockroach/issues/97135

Even more concerning: there's an issue where SELECT FOR UPDATE SKIP LOCKED sometimes skips unlocked rows unexpectedly,  causing workers to miss available work items. https://github.com/cockroachdb/cockroach/issues/121917

A Cockroach Labs engineer Andrew Werner wrote about this exact problem. His recommended pattern uses the atomic claim approach:
```sql
UPDATE items SET claim = $1 WHERE claim IS NULL LIMIT 1 RETURNING id
```

He explicitly notes this pattern is "especially well suited to tasks where the processing time is much larger than the time it takes to acquire the lease"
https://dev.to/ajwerner/quick-and-easy-exactly-once-distributed-work-queues-using-serializable-transactions-jdp

It would be great if we could automatically switch the query to using atomic updates when we detect the connection to CockroachDB instance, or it can be an opt-in parameter.

What are your thoughts on this?

kibertoad:
fixes #669

This ended up being way more complicated than I expected, as there were several incompatible features used in pg-boss, but the end solution should significantly increase the pg-boss compatibility with postgres-compatible solutions across the board.


kibertoad:
Some pg-boss functionality is failing on CockroachDB, will investigate

kibertoad:
will try to address coverage tomorrow, to sleepy now

kibertoad:
@timgit It's ready!

kibertoad:
@timgit before I resolve the conflicts, would be good to get a high level review of the approach

timgit:
High level thoughts so far: 

Advantage: increases usefulness and reach of pg-boss
Disadvantage: opens a can of worms to fix every different pg wire protocol variant db

timgit:
@kibertoad I think it would be good to revive this and make it available.  I'm planning on recruiting Claude to help with the effort of maintenance going forward, so it seems to be less of a long-term overhead concern as it was pre-agent.  Your thoughts?

kibertoad:
@timgit Sounds good! I'll resolve the conflicts then.

kibertoad:
@timgit this is ready to be reviewed again!

timgit:
Thanks! What are your thoughts on running the full suite through a cockroach connected backend to build a compatibility matrix of sorts?  This would be something I could later add to the docs, and it would also serve as a way of knowing if future PRs break something in cockroach.  I am hesitant to lean too much on distributedDatabaseTest.ts.

kibertoad:
@timgit lemme cook something up quickly

kibertoad:
@timgit Done!

kibertoad:
@timgit is there anything left?

timgit:
Thanks!  This looks good so far.  I'll open a tracking branch and merge into that so I can run more tests

PR #687, opened 2026-01-16
feat: Add CLI support

### LINKED ISSUE #686: Support running pg-boss with non-admin user
The fact that pg-boss is trying to run its own migrations is problematic if service is only ever executed with a user account that doesn't have privileges to modify DB schema.

Would you consider introducing some solution for cases like this? For example, adding a CLI command that can be executed with a different user prior to running the application itself.

Documentation currently says this:
```
If the CREATE privilege is not available or desired, you can use the included [static functions](https://timgit.github.io/pg-boss/#/install?id=static-functions) to export the SQL commands to manually create or upgrade the required database schema. This means you will also need to monitor future releases for schema changes.

NOTE: Using an existing schema is supported for advanced use cases but discouraged, as this opens up the possibility that creation will fail on an object name collision, and it will add more steps to the uninstallation process.
```

Static functions part doesn't lead anywhere, though.

kibertoad:
fixes #686

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -6,7 +6,13 @@ pg-boss will automatically create a dedicated schema (`pgboss` is the default na
 GRANT CREATE ON DATABASE db1 TO leastprivuser;
 ```
 
-If the CREATE privilege is not available or desired, you can use the included [static functions](#static-functions) to export the SQL commands to manually create or upgrade the required database schema.  **This means you will also need to monitor future releases for schema changes**.
+If the CREATE privilege is not available or desired, you have two options:
+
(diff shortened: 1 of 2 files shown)


kibertoad:
@timgit Ping?..

timgit:
Thanks for putting this together.  Any opinions on "pg-boss" vs. "pgboss" in the naming?

kibertoad:
@timgit I'd keep it as "pg-boss", it is pretty common to use hyphens in Unix commands, e. g. "docker-compose". But I don't have a strong opinion either way

PR #690, opened 2026-01-22
strict fifo queue policy - wip

sparrovv:


sparrovv:
this is not ready yet

PR #691, opened 2026-01-22
feat: new queue policy key_strict_fifo

sparrovv:
Hi, I wanted to propose a new Queue policy: singleton strict

Requirements:

- Only one job active per `singletonKey` at a time
- While a job is retrying, block the queue for that `singletonKey`
- If retries exhaust and the job fails permanently, block until manual intervention
- Unblock by either deleting the failed job (skip) or retrying it (retry)

## why

We do have a use case where we need to ensure strict sequential processing and even when a job is retrying then we can't pick the next job for the same singletonKey. 
If job fails then it should just block until manual intervention.

Please let me know whether there's any problem of having that new policy in?

thanks 🙇

### DIFF
--- a/src/migrationStore.ts
+++ b/src/migrationStore.ts
@@ -445,8 +445,8 @@ function getAll (schema: string): types.Migration[] {
       previous: 27,
       install: [
         // Create key_strict_fifo index and CHECK constraint on job_common (the default partition)
-        `CREATE UNIQUE INDEX IF NOT EXISTS job_i8 ON ${schema}.job_common (name, singleton_key) WHERE state IN ('active', 'retry', 'failed') AND policy = '${plans.QUEUE_POLICIES.key_strict_fifo}'`,
-        `ALTER TABLE ${schema}.job_common ADD CONSTRAINT job_key_strict_fifo_singleton_key_check CHECK (NOT (policy = '${plans.QUEUE_POLICIES.key_strict_fifo}' AND singleton_key IS NULL))`,
+        `SELECT ${schema}.job_table_run($cmd$CREATE UNIQUE INDEX IF NOT EXISTS job_i8 ON ${schema}.job (name, singleton_key) WHERE state IN ('active', 'retry', 'failed') AND policy = 'key_strict_fifo'$cmd$, 'job_common')`,
+        `SELECT ${schema}.job_table_run($cmd$ALTER TABLE ${schema}.job ADD CONSTRAINT job_key_strict_fifo_singleton_key_check CHECK (NOT (policy = 'key_strict_fifo' AND singleton_key IS NULL))$cmd$, 'job_common')`,
         // Update create_queue function to include the FIFO index for partitioned tables
         `
         CREATE OR REPLACE FUNCTION ${schema}.create_queue(queue_name text, options jsonb)


sparrovv:
Hi @timgit 👋 . I'd appreciate any feedback on this. Thanks 🙇

timgit:
Thank you for submitting a PR.  It took me a while to think about this one.  Overall, I'm in favor of adding policies like this, which was the original intent of creating them in the first place.  I'd also eventually like a way to allow these to be created without a PR, but that's for another day.

I have a few minor suggestions before sending it.

1. Since one of the design goals of pg-boss is to allow direct insertion into the job table when needed, I think you should consider how to handle NULLs in singleton_key when using this policy.  Either support them gracefully or we will need some sort of CHECK constraint added that is scoped to the policy as well.
2. I think the policy name should be changed to something related to keys, such as key_strict_fifo.  This should prevent confusion between this policy and the current singleton policy, and perhaps align better with the new API function `getBlockedKeys()`
3. Please include docs for this policy so it can be understood by the next consumer.  It took me while reading through the code to understand the design goals.
4. Unfortunately, there has been a schema migration since you opened this, so please look through v27 to see recommendations on how to create new indexes, as an example.  Since this index would only apply to a new queue and not affect existing ones, instead of using the "all tables" overload of `job_table_run_async()`, you would only want to run this against the default partition, `job_common`.  This invalidated your function change as well, but hopefully the new changes are easier to follow than the old mixed js/sql technique.

sparrovv:
Hi, thanks for coming back to me.

>Since one of the design goals of pg-boss is to allow direct insertion into the job table when needed, I think you should consider how to handle NULLs in singleton_key when using this policy. Either support them gracefully or we will need some sort of CHECK constraint added that is scoped to the policy as well.

ok, will think about it.

>I think the policy name should be changed to something related to keys, such as key_strict_fifo. This should prevent confusion between this policy and the current singleton policy, and perhaps align better with the new API function getBlockedKeys()

makes sense, will rename to 'key_strict_fifo'

>Please include docs for this policy so it can be understood by the next consumer. It took me while reading through the code to understand the design goals.

sure, sorry, thought the PR desc would be enough. 

> Unfortunately, there has been a schema migration since you opened this, so please look through v27 to see recommendations on how to create new indexes, as an example. Since this index would only apply to a new queue and not affect existing ones, instead of using the "all tables" overload of job_table_run_async(), you would only want to run this against the default partition, job_common. This invalidated your function change as well, but hopefully the new changes are easier to follow than the old mixed js/sql technique.

will have a look.

sparrovv:
@timgit I hadn't realised tests weren't passing, updated now.

timgit:
Looking good!  I think this is one last thing that I need to address in the test suite, but the migrationStore doesn't have the correct function code for create_queue() in the install and it also doesn't restore the previous version in the uninstall section.

src/migrationStore.ts
```diff
@@ -438,6 +438,102 @@ function getAll (schema: string): types.Migration[] {
...
+
+          EXECUTE format('ALTER TABLE ${schema}.%I ADD CONSTRAINT cjc CHECK (name=%L)', tablename, queue_name);
+          EXECUTE format('ALTER TABLE ${schema}.job ATTACH PARTITION ${schema}.%I FOR VALUES IN (%L)', tablename, queue_name);
+        END;
+        $$
+        LANGUAGE plpgsql;
+        `
+      ],
+      uninstall: [
+        `DROP INDEX IF EXISTS ${schema}.job_i8`,
```
timgit:
```suggestion
        `SELECT ${schema}.job_table_run('DROP INDEX IF EXISTS ${schema}.job_i8')`,
```

src/migrationStore.ts
```diff
@@ -438,6 +438,102 @@ function getAll (schema: string): types.Migration[] {
...
+          END IF;
+
+          EXECUTE format('ALTER TABLE ${schema}.%I ADD CONSTRAINT cjc CHECK (name=%L)', tablename, queue_name);
+          EXECUTE format('ALTER TABLE ${schema}.job ATTACH PARTITION ${schema}.%I FOR VALUES IN (%L)', tablename, queue_name);
+        END;
+        $$
+        LANGUAGE plpgsql;
+        `
+      ],
+      uninstall: [
```
timgit: Missing prev version of create_queue()

sparrovv:
I just realised that commits are in a bit mess. I could squash and create a new PR?

timgit:
No worries. I will squash and merge

src/migrationStore.ts
```diff
@@ -438,6 +438,181 @@ function getAll (schema: string): types.Migration[] {
...
         `ALTER TABLE ${schema}.job DROP COLUMN group_id`
       ]
+    },
+    {
+      release: '12.10.0',
+      version: 28,
+      previous: 27,
+      install: [
+        // Create key_strict_fifo index and CHECK constraint on job_common (the default partition)
+        `CREATE UNIQUE INDEX IF NOT EXISTS job_i8 ON ${schema}.job_common (name, singleton_key) WHERE state IN ('active', 'retry', 'failed') AND policy = '${plans.QUEUE_POLICIES.key_strict_fifo}'`,
```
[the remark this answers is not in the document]
timgit: the index name won't match the uninstall's drop command unless this is used to create it.  Confirm the syntax I typed works, lol, since I'm just using GH's suggestion editor in hopes that it helps.

EvertEt:
Hi all, sorry to dig up this PR!

  I'm wondering if there's a possibility to configure the behaviour for failed jobs — or if you'd be open to a change related to this.

  We'd like to use key_strict_fifo, but we don't want to require manual intervention when a job fails permanently. Our jobs are idempotent, so we'd ideally want the queue for a given key to continue processing subsequent jobs even after a failure. Something like a failurePolicy: 'continue' option would cover our use case.

  The workaround of always completing jobs (even on error) doesn't sit well with us since it makes failure visibility worse when filtering by job state.

timgit:
Have you considered using a dead letter queue?

EvertEt:
I have not, thank you for the suggestion! 

Would it be a DLQ  that does the deleteJob on the original queue then? And then the DLQ keeps the failures for visibility?

timgit:
Yes

PR #705, opened 2026-02-06
Dashboard

timgit:


timgit:
@kibertoad fyi.  Still wip, but I think it's definitely crossed the MVP threshold

kibertoad:
@timgit I think there is a major gap now - dashboard is not mentioned anywhere in README.md of pg-boss itself, even though it's likely to be very broadly used. Maybe at least a brief mention with a link to a dedicated readme would be appropriate?

PR #719, opened 2026-02-16
Proxy

timgit:
An HTTP proxy to be published via `@pg-boss/proxy`.

### DIFF
--- a/packages/proxy/package-lock.json
+++ b/packages/proxy/package-lock.json
@@ -12,14 +12,14 @@
         "@hono/node-server": "^1.19.9",
         "@hono/swagger-ui": "^0.5.3",
         "@hono/zod-openapi": "^1.2.2",
-        "@logtape/hono": "^2.0.2",
-        "@logtape/logtape": "^2.0.2",
-        "hono": "^4.11.10",
+        "@logtape/hono": "^2.0.4",
+        "@logtape/logtape": "^2.0.4",
+        "hono": "^4.12.3",
         "pg-boss": "^12.13.0",
         "zod": "^4.3.6"
       },
       "devDependencies": {
-        "@types/node": "^22.19.11",
+        "@types/node": "^22.19.13",
         "cross-env": "^10.1.0",
         "tsx": "^4.21.0",
         "typescript": "^5.9.3",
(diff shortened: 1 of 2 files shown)


timgit:
@kibertoad Do you have any opinions on if this should be a thing or not?  I threw it together mostly via agents, and it hasn't undergone significant testing yet.

kibertoad:
@timgit Definitely a good idea and fills a useful niche!

That said, some questions:
1) Is it intentional that there is no auth by default, including destructive operations?
2) Is subscribe/unsubscribe mechanism intended to be fully functioning and complete? I might be missing something, but it looks like you can only subscribe, but there is not already implemented mechanism for delivering events back to the caller;
3) Job claiming over HTTP doesn't seem to handle crashes gracefully at all, you can only wait for the timeout to happen at that point, it seems. Is the implementation intentional, or is it a known gap? I wonder if some documentation for this case is needed.
4) Relying on copy-paste version of pg-boss contracts seems very brittle. Wouldn't it be better to have some shared package that both pg-boss and proxy rely upon? Alternatively, maybe you could generate these interfaces/contracts automatically, since it's a monorepo anyway.

kibertoad:
Btw, you may want to enable CodeRabbit integration for automatic code reviews, they have free plan for open source, and it works surprisingly well

timgit:
> 1. Is it intentional that there is no auth by default, including destructive operations?

Yes, so far I haven't added hono's basic auth yet.

> 2. Is subscribe/unsubscribe mechanism intended to be fully functioning and complete? I might be missing something, but it looks like you can only subscribe, but there is not already implemented mechanism for delivering events back to the caller;

It's fully functioning, but I question its popularity.  Even after this PR landed, I never used it in my apps.  Proposals for full workflow functionality would make this obsolete.

> 3. Job claiming over HTTP doesn't seem to handle crashes gracefully at all, you can only wait for the timeout to happen at that point, it seems. Is the implementation intentional, or is it a known gap? I wonder if some documentation for this case is needed.

Would you clarify what you mean by handling crashes?   There shouldn't be any background activity in the proxy.

> 4. Relying on copy-paste version of pg-boss contracts seems very brittle. Wouldn't it be better to have some shared package that both pg-boss and proxy rely upon? Alternatively, maybe you could generate these interfaces/contracts automatically, since it's a monorepo anyway.

There are currently build scripts that generate the contracts from the type defs in the primary types.ts.  Let me know if I'm misunderstanding you.

kibertoad:
> It's fully functioning, but I question its popularity. Even after this PR landed, I never used it in my apps. Proposals for full workflow functionality would make this obsolete.

Ah, so it's intended to be used through polling? I thought that maybe it may potentially work via something like SSE, or you'd prefer proxy not to have any hanging connections?

One use-case that I can easily imagine - in one of our current systems when an import or export process are finished, user gets a toast notification in UI, delivered via WebSockets. Subscription sounds like a good fit for this use-case, although it would work better if polling wasn't necessary;

> Would you clarify what you mean by handling crashes? There shouldn't be any background activity in the proxy.

Ah, I see, so proxy is supposed to be a stateless passthrough? Might be helpful to explain it on a conceptual level in readme more.

> There are currently build scripts that generate the contracts from the type defs in the primary types.ts. Let me know if I'm misunderstanding you.

I must be looking in the wrong place, can you point out where it is happening? Are you saying that when pg-boss adds a field (like `heartbeatSeconds`), someone will not have to manually remember to update the proxy's contracts too? If there isn't a safeguard for contracts running out of sync already, maybe adding something for CI would be helpful.

timgit:
> Ah, so it's intended to be used through polling? I thought that maybe it may potentially work via something like SSE, or you'd prefer proxy not to have any hanging connections?

When you call publish(), it looks up the queues that are subscribed and it sends the job any that it finds in the lookup table.  It's an abstraction over calling send() to multiple queues if that's the use case.

> Ah, I see, so proxy is supposed to be a stateless passthrough? Might be helpful to explain it on a conceptual level in readme more.

Yes.

> I must be looking in the wrong place, can you point out where it is happening? Are you saying that when pg-boss adds a field (like heartbeatSeconds), someone will not have to manually remember to update the proxy's contracts too? If there isn't a safeguard for contracts running out of sync already, maybe adding something for CI would be helpful.

Good point.  I'm back merging heartbeat atm, so I'll come up with a way to sync these better.

kibertoad:
@timgit Can I help with pushing it over the finish line? We would like to consider using it in our prod system

timgit:
Surely. What are the biggest gaps left?

kibertoad:
With auth added I'd say the basics are already covered, worth a test!

PR #720, opened 2026-02-16
Dashboard search by job id

### LINKED ISSUE #718: dashboard - Need search job by id functionality
dashboard - Need search job by id functionality,
There should be a universal search that searches for the job by id.

kibertoad:
fixes #718

Note that search is currently implemented in an inefficient way as there is no suitable index for its query.


timgit:
Resolved in #802

PR #751, opened 2026-04-07
Implement transaction adapters for popular ORMs/SQL builders

### LINKED ISSUE #667: Library-agnostic transaction support
https://github.com/timgit/pg-boss/issues/584 seems to only support certain persistence libraries but doesn't provide a universal transaction support.

Would you be open to introducing a system that would expose a shared interface for providing a transaction for the job scheduling command, which would support introducing adapters for different libraries?

kibertoad:
fixes #667


kibertoad:
@timgit ready for the review now!

src/adapters/knex.ts
```diff
@@ -0,0 +1,16 @@
...
+export interface KnexTransactionLike {
+  raw<T = any>(sql: string, bindings?: readonly unknown[]): Promise<{ rows: T[] }>
+}
+
+export function fromKnex (trx: KnexTransactionLike): IDatabase {
+  return {
+    async executeSql (text: string, values?: unknown[]) {
+      // pg-boss emits $1, $2, … placeholders; knex.raw() expects ?
+      const knexSql = text.replace(/\$(\d+)/g, '?')
+      const result = await trx.raw(knexSql, values as readonly unknown[])
```
KazimirPodolski, line 12: @timgit **in general case** this is wrong implementation as nothing guarantees that params will be in the query in the same order as their values. I don't know about particular queries pg-boss does, but here you create an implicit contract to follow.

Consider e.g.:
```sql
INSERT INTO ololo (id, created_at, priority, started_at)
VALUES ($1, $2, $3, $2); -- notice $2 twice
```
Run with parameters `[1, '2026-04-25T13:04:29-12:00', 2]` it will fail with this `fromKnex` as the translated query will have 4 `?`s. Many more examples can be made.

I suggest converting it into knex's [named params syntax](https://knexjs.org/guide/raw.html#raw-parameter-binding) instead.

kibertoad: thank you, I'll create a PR

PR #757, opened 2026-04-08
Ahead of line blocking Fix

bcomnes:
Follow up fix to: https://github.com/timgit/pg-boss/pull/756

I wanted to have the failing test run separate from the fix, but the fix is to filter out saturated concurrency group jobs before looking for the next job.


bcomnes:
Closing in favor of the first PR since its running tests.

Issue #533, opened 2024-12-17
Suggesting new documentation


AdrielBento:
Hi Tim, great job creating and maintaining this library. When I tried to use the library, I had some difficulty reading and interpreting the documentation. I think we could make some improvements to it.

Do you mind if I make a contribution to improving the documentation, switching to [VitePress](https://vitepress.dev/) among other improvements?

timgit:
I don't mind docs improvements.  They are a work in progress, but I'd prefer this is deferred until after v11 is merged, since this would invalidate some of the changes

michalczukm:
+1. I am happy to help there as well!

VitePress sounds like a good idea. 
I'd put one requirement there - **it should support search!**
- For VitePress https://vitepress.dev/reference/default-theme-search
- For Docusaurus https://docusaurus.io/docs/search#using-local-search

I used to docs being in the plain markdown - and just search via cmd+f on the page 😅  
Now I am confused where the pieces of docs live 😅

AdrielBento:
No problem @timgit, I'll wait for the version 11 merge. Any prediction or estimate of when these merge will happen?

timgit:
It just took 8 months, that's all ;)

jmisur:
It would be fantastic if you improve the documentation. A bunch of use cases or examples explaining (or visualizing!) different send functions and their options would be invaluable. Right now it's impossible to be sure about the behavior without verifying your interpretation with manual experiments.

timgit:
I agree!  I will see if I can recruit an agent to help me out with this.

blue2cat:
@timgit, has any work been done toward documentation improvements? I'm interested in contributing.

timgit:
Not yet.  Sounds good!

PR #770, opened 2026-04-24
Move docs to Vitepress

blue2cat:
This PR converts the existing documentation to a [Vitepress](https://vitepress.dev/) site as suggusted in #533. 

Changes include: 
- Add new `package.json` at `docs`
- Install Vitepress and perform inital setup
- Make minor linking adjustments
- Create a new docs homepage, largely based off the existing README in `docs`

Viewing the docs: 
- Run `npm run docs:dev` in the `docs` folder

Deploying to Github Pages: https://vitepress.dev/guide/deploy#github-pages


blue2cat:
@timgit I've ported the existing docs to [VitePress](https://vitepress.dev/). I'd appricaite it if you'd take a look at these inital changes when you've got a moment.

 A few other ideas for improvement, possibly for a future PR: 
- Add an examples section
- Break out the dashboard and proxy pages into separate sections

timgit:
Thanks!  I will need to switch the legacy github pages config over to GH actions to deploy this.  It looks like the majority of this should be handled by the sample workflow file @ https://vitepress.dev/guide/deploy#github-pages.  Do you see anything else that I may be overlooking?

blue2cat:
I think the Vitepress GitHub pages guide should mostly cover what's needed for deploy. ~I'm not sure exactly how the base url will play out.~

EDIT: Looks like `/pg-boss/` is the correct `base` URL for Vitepress: https://vitepress.dev/guide/deploy#setting-a-public-base-path

blue2cat:
A few other things that might be helpful: 
- Adding `paths: ['docs/**']` to the `on > push` section so the workflow only triggers when the docs are updated. 
- Since Vitepress is installed in `docs/`, both `npm ci` and `npm run docs:build` will need `--prefix docs` appended so npm finds the correct install directory.

PR #777, opened 2026-04-29
Add support for job dependencies in pg boss

klesgidis:
## Summary

- Added `createFlow(jobs, options)` for atomic dependency graph creation.
- Added transaction support to the internal `Db` class via `withTransaction()`.
- Added validation for:
  - duplicate refs
  - missing dependency refs
  - self-dependencies
  - dependency cycles, including the detected cycle in the error message
- Updated dependency tests to use `createFlow()`.
- Updated docs to describe `createFlow()` as the dependency API.

## Example

```js
const flow = await boss.createFlow([
  { ref: 'extract-a', name: 'extract', data: { file: '1.csv' } },
  { ref: 'extract-b', name: 'extract', data: { file: '2.csv' } },
  {
    ref: 'load',
    name: 'load',
    data: { output: 'report.csv' },
    dependsOn: ['extract-a', 'extract-b']
  }
])

const loadJobId = flow.load
```
Resolves https://github.com/timgit/pg-boss/issues/745

### DIFF
--- a/docs/api/jobs.md
+++ b/docs/api/jobs.md
@@ -201,6 +201,52 @@ interface JobInsert<T = object> {
 }
 ```
 
+```js
+const [idA, idB] = await boss.insert('etl', [
+  { data: { step: 'extract' } },
+  { data: { step: 'transform' } }
+], { returnId: true })
+```
+
+### `createFlow(jobs, options)`
+
+Create a set of jobs and their dependencies atomically in one transaction.
+
+Use `createFlow()` when jobs depend on other jobs. Dependencies are not configured on `send()` or `insert()` because creating jobs and dependencies in separate calls can race with job completion.
+
+The method accepts a flat array of jobs in any order. Each job has a local `ref`, and dependent jobs reference parent refs with `dependsOn`.
+
+```ts
+interface FlowJob {
+  ref: string;
+  name: string;
+  data?: object;
+  options?: Omit<JobInsert, 'data'>;
+  dependsOn?: string[];
+}
+```
+
+Returns a map of `ref` to created job id.
+
+```js
+const flow = await boss.createFlow([
+  { ref: 'extract-a', name: 'extract', data: { file: '1.csv' } },
+  { ref: 'extract-b', name: 'extract', data: { file: '2.csv' } },
+  {
+    ref: 'load',
+    name: 'load',
+    data: { output: 'report.csv' },
+    dependsOn: ['extract-a', 'extract-b']
+  }
+])
+
+const loadJobId = flow.load
+```
+
+Dependent jobs are created in a `blocked` state and won't be eligible for fetching until all parent jobs have completed. If a parent job fails or is cancelled, the child remains blocked. You can explicitly `cancel` or `fail` the blocked child if needed.
+
+When a dependent job uses `startAfter`, both conditions must be met: all dependencies completed and `startAfter` has passed.
+
 ### `fetch(name, options)`
 
 Returns an array of jobs from a queue
@@ -261,6 +307,8 @@ Returns an array of jobs from a queue
       createdOn: Date;
       completedOn: Date | null;
       keepUntil: Date;
+      blocked: boolean,
+      blocking: boolean,
       deadLetter: string,
       policy: string,
       output: object
@@ -472,4 +520,22 @@ const jobs = await boss.findJobs('my-queue', {
   data: { type: 'email' },
   queued: true
 })
+```
+
+### `getDependencies(name, id, options)`
+
+Returns an array of parent job references that the specified job depends on.
+
+```js
+const parents = await boss.getDependencies('aggregate-results', jobId)
+// [{ name: 'process-data', id: '...' }, { name: 'process-data', id: '...' }]
+```
+
+### `getDependents(name, id, options)`
+
+Returns an array of child job references that depend on the specified job.
+
+```js
+const children = await boss.getDependents('process-data', parentJobId)
+// [{ name: 'aggregate-results', id: '...' }]
 ```
--- a/docs/api/jobs.md
+++ b/docs/api/jobs.md
@@ -214,6 +214,8 @@ Create a set of jobs and their dependencies atomically in one transaction.
 
 Use `createFlow()` when jobs depend on other jobs. Dependencies are not configured on `send()` or `insert()` because creating jobs and dependencies in separate calls can race with job completion.
 
+Atomicity is handled by pg-boss when it owns the database connection. If you pass a custom `db` in `options`, wrap the call in your own transaction if you need the job and dependency inserts to commit or roll back together.
+
 The method accepts a flat array of jobs in any order. Each job has a local `ref`, and dependent jobs reference parent refs with `dependsOn`.
 
 ```ts
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "db:construct": "node --import=tsx -e 'console.log(require(\"./src\").getConstructionPlans())'"
   },
   "pgboss": {
-    "schema": 30
+    "schema": 31
   },
   "repository": {
     "type": "git",
--- a/packages/proxy/src/contracts.ts
+++ b/packages/proxy/src/contracts.ts
@@ -148,6 +148,8 @@ export const jobWithMetadataSchema = jobSchemaBase.extend({
   createdOn: z.iso.datetime().transform((val) => new Date(val)),
   completedOn: z.iso.datetime().nullable().transform((val) => val ? new Date(val) : null),
   keepUntil: z.iso.datetime().transform((val) => new Date(val)),
+  blocked: z.boolean(),
+  blocking: z.boolean(),
   policy: z.string(),
   deadLetter: z.string(),
   output: jsonRecordSchema,
--- a/packages/proxy/src/contracts.ts
+++ b/packages/proxy/src/contracts.ts
@@ -150,6 +150,7 @@ export const jobWithMetadataSchema = jobSchemaBase.extend({
   keepUntil: z.iso.datetime().transform((val) => new Date(val)),
   blocked: z.boolean(),
   blocking: z.boolean(),
+  pendingDependencies: z.number(),
   policy: z.string(),
   deadLetter: z.string(),
   output: jsonRecordSchema,
--- a/src/attorney.ts
+++ b/src/attorney.ts
@@ -85,6 +85,104 @@ function validateGroupConfig (config: any) {
   assert(!('tier' in config.group) || (typeof config.group.tier === 'string' && config.group.tier.length > 0), 'group.tier must be a non-empty string if provided')
 }
 
+function validateFlowJobs (jobs: types.FlowJob[]) {
+  assert(Array.isArray(jobs), 'createFlow requires an array of jobs')
+  assert(jobs.length >= 2, 'createFlow requires at least 2 jobs')
+
+  const refs = new Set<string>()
+  for (const job of jobs) {
+    assert(typeof job.ref === 'string' && job.ref.length > 0, 'each flow job must have a non-empty ref')
+    assert(!refs.has(job.ref), `duplicate ref: "${job.ref}"`)
+    refs.add(job.ref)
+
+    assert(typeof job.name === 'string' && job.name.length > 0, 'each flow job must have a non-empty name')
+    assertObjectName(job.name)
+  }
+
+  const hasDeps = jobs.some(j => j.dependsOn && j.dependsOn.length > 0)
+  assert(hasDeps, 'createFlow requires at least one job with dependsOn')
+
+  for (const job of jobs) {
+    if (!job.dependsOn) continue
+    assert(Array.isArray(job.dependsOn), `dependsOn for ref "${job.ref}" must be an array`)
+    for (const dep of job.dependsOn) {
+      assert(typeof dep === 'string' && dep.length > 0, `dependsOn entries must be non-empty strings`)
+      assert(dep !== job.ref, `job "${job.ref}" cannot depend on itself`)
+      assert(refs.has(dep), `dependsOn ref "${dep}" not found in flow`)
+    }
+  }
+
+  // Cycle detection via topological sort
+  const inDegree = new Map<string, number>()
+  const edges = new Map<string, string[]>()
+  for (const job of jobs) {
+    inDegree.set(job.ref, 0)
+    edges.set(job.ref, [])
+  }
+  for (const job of jobs) {
+    if (!job.dependsOn) continue
+    for (const dep of job.dependsOn) {
+      edges.get(dep)!.push(job.ref)
+      inDegree.set(job.ref, inDegree.get(job.ref)! + 1)
+    }
+  }
+  const queue: string[] = []
+  for (const [ref, deg] of inDegree) {
+    if (deg === 0) queue.push(ref)
+  }
+  let visited = 0
+  while (queue.length > 0) {
+    const current = queue.shift()!
+    visited++
+    for (const child of edges.get(current)!) {
+      const newDeg = inDegree.get(child)! - 1
+      inDegree.set(child, newDeg)
+      if (newDeg === 0) queue.push(child)
+    }
+  }
+  if (visited !== jobs.length) {
+    const cycle = findDependencyCycle(edges)
+    assert(false, `createFlow contains a dependency cycle: ${cycle.join(' -> ')}`)
+  }
+}
+
+function findDependencyCycle (edges: Map<string, string[]>): string[] {
+  const visiting = new Set<string>()
+  const visited = new Set<string>()
+  const path: string[] = []
+
+  function visit (ref: string): string[] | null {
+    if (visiting.has(ref)) {
+      const start = path.indexOf(ref)
+      return [...path.slice(start), ref]
+    }
+
+    if (visited.has(ref)) return null
+
+    visiting.add(ref)
+    path.push(ref)
+
+    for (const child of edges.get(ref) || []) {
+      const cycle = visit(child)
+      if (cycle) return cycle
+    }
+
+    path.pop()
+    visiting.delete(ref)
+    visited.add(ref)
+
+    return null
+  }
+
+  let cycle: string[] | null = null
+  for (const ref of edges.keys()) {
+    cycle = visit(ref)
+    if (cycle) break
+  }
+
+  return cycle!
+}
+
 function validateGroupConcurrencyValue (value: any, optionName: string) {
   if (typeof value === 'number') {
     assert(Number.isInteger(value) && value >= 1, `${optionName} must be an integer >= 1`)
@@ -374,5 +472,6 @@ export {
   checkWorkArgs,
   getConfig,
   POLICY,
+  validateFlowJobs,
   validateQueueArgs
 }
--- a/src/attorney.ts
+++ b/src/attorney.ts
@@ -106,7 +106,7 @@ function validateFlowJobs (jobs: types.FlowJob[]) {
     if (!job.dependsOn) continue
     assert(Array.isArray(job.dependsOn), `dependsOn for ref "${job.ref}" must be an array`)
     for (const dep of job.dependsOn) {
-      assert(typeof dep === 'string' && dep.length > 0, `dependsOn entries must be non-empty strings`)
+      assert(typeof dep === 'string' && dep.length > 0, 'dependsOn entries must be non-empty strings')
       assert(dep !== job.ref, `job "${job.ref}" cannot depend on itself`)
       assert(refs.has(dep), `dependsOn ref "${dep}" not found in flow`)
     }
--- a/src/boss.ts
+++ b/src/boss.ts
@@ -246,6 +246,9 @@ class Boss extends EventEmitter implements types.EventsMixin {
       const queues = rows.map((q) => q.name)
       const sql = plans.deletion(this.#config.schema, table, queues)
       await this.#executeQuery(sql)
+
+      const depSql = plans.cleanupDependencies(this.#config.schema, table, queues)
+      await this.#executeQuery(depSql)
     }
   }
 }
--- a/src/db.ts
+++ b/src/db.ts
@@ -54,6 +54,26 @@ class Db extends EventEmitter implements types.IDatabase, types.EventsMixin {
 
     return await this.pool.query(text, values)
   }
+
+  async withTransaction<T> (fn: (db: types.IDatabase) => Promise<T>): Promise<T> {
+    assert(this.opened, 'Database not opened. Call open() before executing SQL.')
+
+    const client = await this.pool.connect()
+    try {
+      await client.query('BEGIN')
+      const txDb: types.IDatabase = {
+        executeSql: (text: string, values?: unknown[]) => client.query(text, values)
+      }
+      const result = await fn(txDb)
+      await client.query('COMMIT')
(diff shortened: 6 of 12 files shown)


klesgidis:
@timgit let me know what you think

klesgidis:
I've also tested this locally

bcomnes:
Like the idea of the feature, need to really sit down and read the migration file. Have you had a chance to run any load through this feature to see how it scales/performs?

klesgidis:
> Like the idea of the feature, need to really sit down and read the migration file. Have you had a chance to run any load through this feature to see how it scales/performs?

With the help of my friend cursor, I performed a stress test locally and bellow we have the results

# Dependency Stress Test Results

I added a local stress harness for the dependency implementation and ran several workload shapes to exercise both normal queue behavior and dependency-heavy paths.

## Setup

> `scripts/stress-dependencies.ts`
>
> ```ts
> import { performance } from 'node:perf_hooks'
> import Db from '../src/db.ts'
> import { PgBoss } from '../src/index.ts'
> import type { ConstructorOptions, FlowJob, Job } from '../src/types.ts'
>
> type Pattern = 'baseline' | 'chain' | 'fanout' | 'fanin' | 'layered'
>
> type Config = {
>   pattern: Pattern
>   jobs: number
>   batchSize: number
>   concurrency: number
>   layers: number
>   depsPerJob: number
>   idlePolls: number
>   schema: string
>   queue: string
>   cleanup: boolean
> }
>
> type TimedStats = {
>   samples: number[]
>   total: number
> }
>
> const PATTERNS = new Set<Pattern>(['baseline', 'chain', 'fanout', 'fanin', 'layered'])
>
> function envString (name: string, defaultValue: string): string {
>   return process.env[name] || defaultValue
> }
>
> function envNumber (name: string, defaultValue: number): number {
>   const value = process.env[name]
>   if (!value) return defaultValue
>
>   const numberValue = Number(value)
>   if (!Number.isFinite(numberValue) || numberValue <= 0) {
>     throw new Error(`${name} must be a positive number`)
>   }
>
>   return Math.floor(numberValue)
> }
>
> function getConfig (): Config {
>   const pattern = envString('PATTERN', 'layered') as Pattern
>   if (!PATTERNS.has(pattern)) {
>     throw new Error(`PATTERN must be one of: ${[...PATTERNS].join(', ')}`)
>   }
>
>   const schema = envString('SCHEMA', `pgboss_stress_${process.pid}`)
>
>   return {
>     pattern,
>     jobs: envNumber('JOBS', 1000),
>     batchSize: envNumber('BATCH_SIZE', 100),
>     concurrency: envNumber('CONCURRENCY', 1),
>     layers: envNumber('LAYERS', 10),
>     depsPerJob: envNumber('DEPS_PER_JOB', 3),
>     idlePolls: envNumber('IDLE_POLLS', 600),
>     schema,
>     queue: envString('QUEUE', `${schema}_q`),
>     cleanup: envString('CLEANUP', '1') !== '0'
>   }
> }
>
> function getBossConfig (config: Config): ConstructorOptions {
>   const connectionString = process.env.DATABASE_URL
>
>   return {
>     ...(connectionString
>       ? { connectionString }
>       : {
>           host: envString('POSTGRES_HOST', '127.0.0.1'),
>           port: envNumber('POSTGRES_PORT', 5432),
>           database: envString('POSTGRES_DATABASE', 'pgboss'),
>           user: envString('POSTGRES_USER', 'postgres'),
>           password: envString('POSTGRES_PASSWORD', 'postgres')
>         }),
>     max: envNumber('POSTGRES_MAX', Math.max(3, config.concurrency + 2)),
>     schema: config.schema,
>     createSchema: true,
>     migrate: true,
>     supervise: false,
>     schedule: false
>   }
> }
>
> function createTimer (): TimedStats {
>   return {
>     samples: [],
>     total: 0
>   }
> }
>
> async function time<T> (stats: TimedStats, fn: () => Promise<T>): Promise<T> {
>   const start = performance.now()
>   try {
>     return await fn()
>   } finally {
>     const elapsed = performance.now() - start
>     stats.samples.push(elapsed)
>     stats.total += elapsed
>   }
> }
>
> function percentile (values: number[], p: number): number {
>   if (values.length === 0) return 0
>   const sorted = [...values].sort((a, b) => a - b)
>   const index = Math.min(sorted.length - 1, Math.ceil((p / 100) * sorted.length) - 1)
>   return sorted[index]
> }
>
> function formatMs (value: number): string {
>   return `${value.toFixed(1)}ms`
> }
>
> function formatRate (count: number, ms: number): string {
>   if (ms <= 0) return 'n/a'
>   return `${(count / (ms / 1000)).toFixed(1)}/s`
> }
>
> function summarizeTimer (name: string, stats: TimedStats): void {
>   console.log(`${name}: total=${formatMs(stats.total)} p50=${formatMs(percentile(stats.samples, 50))} p95=${formatMs(percentile(stats.samples, 95))} p99=${formatMs(percentile(stats.samples, 99))} samples=${stats.samples.length}`)
> }
>
> function buildFlowJobs (config: Config): FlowJob[] {
>   if (config.jobs < 2) {
>     throw new Error('JOBS must be at least 2 for dependency patterns')
>   }
>
>   switch (config.pattern) {
>     case 'chain':
>       return Array.from({ length: config.jobs }, (_, index) => ({
>         ref: `j${index}`,
>         name: config.queue,
>         data: { index },
>         ...(index === 0 ? {} : { dependsOn: [`j${index - 1}`] })
>       }))
>
>     case 'fanout':
>       return [
>         { ref: 'parent', name: config.queue, data: { role: 'parent' } },
>         ...Array.from({ length: config.jobs - 1 }, (_, index) => ({
>           ref: `child${index}`,
>           name: config.queue,
>           data: { index, role: 'child' },
>           dependsOn: ['parent']
>         }))
>       ]
>
>     case 'fanin':
>       return [
>         ...Array.from({ length: config.jobs - 1 }, (_, index) => ({
>           ref: `parent${index}`,
>           name: config.queue,
>           data: { index, role: 'parent' }
>         })),
>         {
>           ref: 'child',
>           name: config.queue,
>           data: { role: 'child' },
>           dependsOn: Array.from({ length: config.jobs - 1 }, (_, index) => `parent${index}`)
>         }
>       ]
>
>     case 'layered':
>       return buildLayeredFlowJobs(config)
>
>     default:
>       throw new Error(`Unsupported flow pattern: ${config.pattern}`)
>   }
> }
>
> function buildLayeredFlowJobs (config: Config): FlowJob[] {
>   const layers = Math.min(config.layers, config.jobs)
>   const jobsByLayer: string[][] = []
>   const jobs: FlowJob[] = []
>
>   for (let layer = 0; layer < layers; layer++) {
>     const remainingJobs = config.jobs - jobs.length
>     const remainingLayers = layers - layer
>     const layerSize = Math.ceil(remainingJobs / remainingLayers)
>     const currentLayerRefs: string[] = []
>     const previousLayerRefs = jobsByLayer[layer - 1] || []
>
>     for (let index = 0; index < layerSize; index++) {
>       const ref = `l${layer}_j${index}`
>       currentLayerRefs.push(ref)
>
>       if (previousLayerRefs.length === 0) {
>         jobs.push({ ref, name: config.queue, data: { layer, index } })
>         continue
>       }
>
>       const depCount = Math.min(config.depsPerJob, previousLayerRefs.length)
>       const dependsOn = Array.from({ length: depCount }, (_, depIndex) => {
>         return previousLayerRefs[(index + depIndex) % previousLayerRefs.length]
>       })
>
>       jobs.push({ ref, name: config.queue, data: { layer, index }, dependsOn })
>     }
>
>     jobsByLayer.push(currentLayerRefs)
>   }
>
>   return jobs
> }
>
> async function createWorkload (boss: PgBoss, config: Config): Promise<number> {
>   if (config.pattern !== 'baseline') {
>     const jobs = buildFlowJobs(config)
>     await boss.createFlow(jobs)
>     return jobs.length
>   }
>
>   const jobs = Array.from({ length: config.jobs }, (_, index) => ({ data: { index } }))
>   for (let offset = 0; offset < jobs.length; offset += config.batchSize) {
>     await boss.insert(config.queue, jobs.slice(offset, offset + config.batchSize))
>   }
>
>   return jobs.length
> }
>
> async function workerLoop (
>   boss: PgBoss,
>   config: Config,
>   totals: { completed: number },
>   stats: { fetch: TimedStats, complete: TimedStats },
>   expectedJobs: number
> ): Promise<void> {
>   let idlePolls = 0
>
>   while (totals.completed < expectedJobs) {
>     const jobs = await time(stats.fetch, async () => {
>       return boss.fetch(config.queue, { batchSize: config.batchSize })
>     })
>
>     if (jobs.length === 0) {
>       idlePolls++
>       if (idlePolls > config.idlePolls) {
>         throw new Error(`No fetchable jobs remain after completing ${totals.completed}/${expectedJobs}`)
>       }
>       await new Promise(resolve => setTimeout(resolve, 100))
>       continue
>     }
>
>     idlePolls = 0
>
>     await completeJobs(boss, config.queue, jobs, stats.complete)
>     totals.completed += jobs.length
>   }
> }
>
> async function completeJobs (boss: PgBoss, queue: string, jobs: Job<unknown>[], stats: TimedStats): Promise<void> {
>   await time(stats, async () => {
>     await boss.complete(queue, jobs.map(job => job.id))
>   })
> }
>
> async function getDb (bossConfig: ConstructorOptions): Promise<Db> {
>   const db = new Db(bossConfig)
>   await db.open()
>   return db
> }
>
> async function getCounts (db: Db, schema: string): Promise<{ blocked: number, blocking: number, completed: number, dependencies: number }> {
>   const [jobs, dependencies] = await Promise.all([
>     db.executeSql(`
>       SELECT
>         count(*) FILTER (WHERE blocked) as blocked,
>         count(*) FILTER (WHERE blocking) as blocking,
>         count(*) FILTER (WHERE state = 'completed') as completed
>       FROM ${schema}.job
>     `),
>     db.executeSql(`SELECT count(*) as dependencies FROM ${schema}.job_dependency`)
>   ])
>
>   return {
>     blocked: Number(jobs.rows[0].blocked),
>     blocking: Number(jobs.rows[0].blocking),
>     completed: Number(jobs.rows[0].completed),
>     dependencies: Number(dependencies.rows[0].dependencies)
>   }
> }
>
> async function dropSchema (bossConfig: ConstructorOptions, schema: string): Promise<void> {
>   const db = await getDb(bossConfig)
>   try {
>     await db.executeSql(`DROP SCHEMA IF EXISTS ${schema} CASCADE`)
>   } finally {
>     await db.close()
>   }
> }
>
> async function main (): Promise<void> {
>   const config = getConfig()
>   const bossConfig = getBossConfig(config)
>   const createStats = createTimer()
>   const fetchStats = createTimer()
>   const completeStats = createTimer()
>   const totals = { completed: 0 }
>
>   console.log(`dependency stress: pattern=${config.pattern} jobs=${config.jobs} batchSize=${config.batchSize} concurrency=${config.concurrency} schema=${config.schema}`)
>
>   await dropSchema(bossConfig, config.schema)
>
>   const boss = new PgBoss(bossConfig)
>   let db: Db | undefined
>
>   try {
>     await boss.start()
>     await boss.createQueue(config.queue)
>     db = await getDb(bossConfig)
>
>     const created = await time(createStats, async () => createWorkload(boss, config))
>     const processStart = performance.now()
>
>     await Promise.all(Array.from({ length: config.concurrency }, async () => {
>       await workerLoop(boss, config, totals, { fetch: fetchStats, complete: completeStats }, created)
>     }))
>
>     const processMs = performance.now() - processStart
>     const counts = await getCounts(db, config.schema)
>
>     console.log('')
>     console.log('Results')
>     console.log(`created=${created} completed=${totals.completed} blocked=${counts.blocked} blocking=${counts.blocking} dependencies=${counts.dependencies}`)
>     console.log(`processing=${formatMs(processMs)} throughput=${formatRate(totals.completed, processMs)}`)
>     summarizeTimer('create', createStats)
>     summarizeTimer('fetch', fetchStats)
>     summarizeTimer('complete', completeStats)
>
>     if (counts.completed !== created || counts.blocked !== 0) {
>       process.exitCode = 1
>     }
>   } finally {
>     await db?.close()
>     await boss.stop({ graceful: false })
>
>     if (config.cleanup) {
>       await dropSchema(bossConfig, config.schema)
>     }
>   }
> }
>
> main().catch(err => {
>   console.error(err)
>   process.exitCode = 1
> })
> ```

It supports these environment variables:

```bash
PATTERN=baseline|fanout|fanin|chain|layered
JOBS=10000
BATCH_SIZE=500
CONCURRENCY=4
LAYERS=20
DEPS_PER_JOB=3
```

The script creates a disposable pg-boss schema, creates the workload, repeatedly fetches and completes jobs until all expected jobs are completed, then verifies that no jobs remain blocked.

## Polling Strategy

The harness uses a manual fetch/complete loop rather than `boss.work()`.

Each worker:

1. Calls `boss.fetch(queue, { batchSize })`.
2. Completes any returned jobs with `boss.complete(queue, ids)`.
3. If no jobs are returned, sleeps for `100ms` and retries.
4. Fails the run if no fetchable jobs appear after `IDLE_POLLS` empty polls.

This keeps the measurement focused on dependency creation, fetch eligibility, and completion-time unblocking. The dependency implementation itself does not add a dependency-specific polling loop. Jobs are unblocked synchronously inside the `complete_jobs` SQL path when parent jobs complete.

## Workloads

I tested these patterns:

- `baseline`: 10k normal jobs with no dependencies, to compare against non-dependent queue behavior.
- `fanout`: 1 parent and 9,999 children depending on it.
- `fanin`: 9,999 parents and 1 child depending on all parents.
- `layered`: 20 DAG layers, 10k jobs total, each dependent layer job depending on 3 jobs from the previous layer.
- `chain`: 10k jobs in a single sequential chain, where each job depends on the previous one.

## Results

All final 10k runs completed with:

```text
created=10000
completed=10000
blocked=0
```

| Pattern | Dependency rows | Processing time | Throughput | Create total | Fetch p95 | Complete p95 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| baseline | 0 | 583.3ms | 17,143.0/s | 307.5ms | 36.4ms | 138.6ms |
| fanout | 9,999 | 895.4ms | 11,168.8/s | 419.0ms | 32.4ms | 147.4ms |
| fanin | 9,999 | 1,258.1ms | 7,948.6/s | 889.7ms | 59.5ms | 330.0ms |
| layered | 28,500 | 1,325.3ms | 7,545.3/s | 1,364.9ms | 38.3ms | 108.5ms |
| chain | 9,999 | 76,641.9ms | 130.5/s | 761.6ms | 5.2ms | 7.7ms |

The chain case is expectedly much slower because only one job can become fetchable at a time. That means the harness performs roughly 10k sequential fetch/complete/unblock cycles rather than processing jobs in large batches.

## Notes

Validation performed:

- Smoke tested all workload patterns.
- Ran 10k workloads for all patterns listed above.

klesgidis:
@timgit Could you please take a look, since this is a feature we need to add to our production system

bcomnes:
Upstreaming OSS patches should never be a production blocker imo. A pattern that has worked very well for my own use of pb-boss is running pnpm patches (or similar) or git ref forks of work I'm working on upstreaming at the schedule I need it by and also accommodating the time Tim has available for review. It also helps validate the work under real production load, and in turn improves the quality of what I end up landing for everyone.

klesgidis:
> Upstreaming OSS patches should never be a production blocker imo. A pattern that has worked very well for my own use of pb-boss is running pnpm patches (or similar) or git ref forks of work I'm working on upstreaming at the schedule I need it by and also accommodating the time Tim has available for review. It also helps validate the work under real production load, and in turn improves the quality of what I end up landing for everyone.

Thanks, that makes sense. I did not mean to imply that upstreaming this should be a production blocker. What I meant is that we would like to use this feature as soon as we reasonably can.

We can of course carry a patch or fork internally, but since this feature has gone through a few implementation changes already and includes database migrations, I would be hesitant to put it into production before getting some signal from the maintainers that the overall direction is acceptable.

If the approach looks broadly OK, even if there are follow-up changes needed, I’d be much more comfortable validating it under real production load and feeding that experience back into the PR.

the-mediocre-dev:
Super excited about this, its been painful defining dependencies manually in code.

If this is basically just a declarative abstraction over `onComplete` from v9, could it make sense for the abstraction to capture an `onFail` analogue which allows scheduling child jobs upon failure rather than completion?

For example:

```
const flow = await boss.createFlow([
  { ref: 'extract', name: 'extract', data: { file: '1.csv' } },
  {
    ref: 'load',
    name: 'load',
    data: { output: 'report.csv' },
    dependsOnSuccess: ['extract']
  },
  {
    ref: 'notify',
    name: 'notify',
    data: { },
    dependsOnFailure: ['extract']
  }
])
```

Also wonder if it makes sense to allow failed jobs to still schedule children, modeling a hard vs soft dependency, hard dependency being fatal to the workflow execution if its not met, soft being not fatal. Some workflows contain units of work that have to be executed in specific order but dont all have to complete, for example a heavy telemetry
type event that doesnt matter if it gets dropped, or syncing / caching data for future workflow runs.

klesgidis:
> If this is basically just a declarative abstraction over onComplete from v9, could it make sense for the abstraction to capture an onFail analogue which allows scheduling child jobs upon failure rather than completion?

Yes I think we could do that and it is a nice idea!

> Also wonder if it makes sense to allow failed jobs to still schedule children, modeling a hard vs soft dependency, hard dependency being fatal to the workflow execution if its not met, soft being not fatal. Some workflows contain units of work that have to be executed in specific order but dont all have to complete, for example a heavy telemetry type event that doesnt matter if it gets dropped, or syncing / caching data for future workflow runs.

Yes, this was something described also in the initial issue [here](https://github.com/timgit/pg-boss/issues/745) (See Scope section), but I decided to not include it in the initial implementation but as future work

the-mediocre-dev:
> > If this is basically just a declarative abstraction over onComplete from v9, could it make sense for the abstraction to capture an onFail analogue which allows scheduling child jobs upon failure rather than completion?
> 
> Yes I think we could do that and it is a nice idea!
> 
> > Also wonder if it makes sense to allow failed jobs to still schedule children, modeling a hard vs soft dependency, hard dependency being fatal to the workflow execution if its not met, soft being not fatal. Some workflows contain units of work that have to be executed in specific order but dont all have to complete, for example a heavy telemetry type event that doesnt matter if it gets dropped, or syncing / caching data for future workflow runs.
> 
> Yes, this was something described also in the initial issue [here](https://github.com/timgit/pg-boss/issues/745) (See Scope section), but I decided to not include it in the initial implementation but as future work

Actually now I think on it the `onFail` mechanism is expressive enough to describe a soft relationship

``` javascript
const flow = await boss.createFlow([
  { ref: 'job-1', name: 'job-1', data: { },
  {
    ref: 'job-2-success',
    name: 'job-2',
    data: { },
    dependsOnSuccess: ['job-1']
  },
  {
    ref: 'job-2-failed',
    name: 'job-2',
    data: { },
    dependsOnFailure: ['job-1']
  }
])
```

Its verbose, but it would do the trick!

EDIT: or even


``` javascript
const flow = await boss.createFlow([
  { ref: 'job-1', name: 'job-1', data: { },
  {
    ref: 'job-2',
    name: 'job-2',
    data: { },
    dependsOnSuccess: ['job-1'],
    dependsOnFailure: ['job-1']
  }
])
```

timgit:
Keep in mind the overlap between "on fail" and a DLQ defined on the queue as well.

klesgidis:
So how do you want me to proceed here? 

Are we ok with the above implementation? Could we merge the basic functionality before we proceed to enhancements?

klesgidis:
> So how do you want me to proceed here?
> 
> Are we ok with the above implementation? Could we merge the basic functionality before we proceed to enhancements?

any update here?

timgit:
Apologies for the delay in messaging. I will have an answer for you by next week.

timgit:
Thanks for your patience on the review, and of course the submission.  This looks like a cool new feature!  In order to help me reason through a few things, I asked Claude to help me out with the review.  Following is a summary our chat and has the next steps needed to merge.

# Claude Fable PR Review

## Summary

Adds a `blocked` column, dependency table, and unblocking at completion time. A couple of items are correctness issues rather than enhancements, and should be resolved before merging.

## Requested changes

**1. Fan-in race in the unblock check.** If 2 workers complete the last 2 parents of the same child in overlapping transactions, each snapshot still sees the other parent as not completed, so neither clears `blocked` and the child is stranded permanently. A dependency counter on the child resolves this: `createFlow` sets it to the number of parents, completion decrements it, and `blocked` clears when it reaches 0. Row locking on the child serializes concurrent decrements, so the visibility race goes away. For example:

```sql
WITH completed AS (
  UPDATE <queue_table>
  SET state = 'completed', completed_on = now(), output = $1
  WHERE name = $2 AND id = ANY($3) AND state = 'active'
  RETURNING id, blocking
),
decremented AS (
  SELECT d.child_name, d.child_id, count(*) AS n
  FROM job_dependency d
  JOIN completed c ON c.blocking AND d.parent_name = $2 AND d.parent_id = c.id
  GROUP BY 1, 2
)
UPDATE job j
SET pending_dependencies = j.pending_dependencies - d.n,
    blocked = (j.pending_dependencies - d.n) > 0
FROM decremented d
WHERE j.name = d.child_name AND j.id = d.child_id AND j.blocked
```

This also resolves the concern with the `complete_jobs` function. Since the decrement doesn't read parent states, `completeJobs` folds back into a single statement, so the function and the temp table can be removed. The temp table is a problem on its own: it adds catalog churn to every `complete()` call, including all internal completions, and `CREATE TEMP TABLE` will fail if `complete()` runs twice inside one user-supplied transaction. The CTE costs nothing for jobs without dependencies, since the `c.blocking` join is empty and `job_dependency` is never probed.

The retry path will need to carry `pending_dependencies` through the fail/reinsert cycle, the same as it already does for `blocked` and `blocking`.

**2. `createFlow` should verify insert counts.** `insertJobs` ends with `ON CONFLICT DO NOTHING`, so a duplicate id or a singleton conflict can silently skip a parent while the dependency rows are still created, leaving the child blocked forever (and with the counter, permanently above 0). Return the ids from the insert and throw inside the transaction if the count doesn't match.

**3. Set `blocking` in the insert payload.** Since the flow graph is known up front, parents can be inserted with `blocking: true` directly, which removes `markJobsBlocking` and closes a race for custom `db` usage where the statements don't run in a transaction. Related: the docs should mention that `createFlow` atomicity requires a pg-boss-managed connection.

**4. Revert the formatting-only changes** (function spacing, ternary re-indents, removed comments in plans.ts and elsewhere). They conflict with the lint config and make a migration-bearing diff much harder to review.

## Smaller items

- The `group` → `groupId`/`groupTier` change in `insert()` looks like a legitimate bug fix, but it's unrelated — it should move to its own PR with its own test.
- `blocked`/`blocking` are now accepted through the public `insert()` payload, so an untyped caller could create a permanently blocked job with no dependencies. These should be stripped in `insert()`.
- Tests to add: concurrent fan-in completion (2 connections completing sibling parents), the conflict-skip case from item 2, two `complete()` calls in one transaction, and `includeQueued: true` completing a blocked child and its parent in the same batch.

## Guiding invariant

Blocked jobs should never become permanently unreachable without an explicit user action. Items 1–3 all reduce to this.

klesgidis:
Thanks for your feedback! I think that I addressed most of the changes. Please have a look and let me know

timgit:
This looks good. I'm going to merge and make a minor API adjustment to flow() instead of createFlow()

Issue #730, opened 2026-03-02
dashboard - run as middleware


pradel:
There are currently multiple ways to install the dashboard, but I would like to install it as a middleware and plug it into my Fastify server (it would work with any other framework as we just forward the request) so that it can have the same auth rules and run in my app.
I am currently using pg-bossman, which is quite easy to use as a middleware. This is how I am using it:

```ts
const bossmanClient = createClient({
    connectionString: env.DATABASE_URL,
  });

  const dashboard = createDashboard({
    basePath: "/api/admin/dashboard",
    client: bossmanClient,
  });

  const handleDashboardRequest = async (request: FastifyRequest) => {
    // Require admin authentication
    await requireAuth(request, "admin");

    const headers = convertFastifyHeaders(request.headers);
    const nativeRequest = new Request(
      `${request.protocol}://${request.hostname}${request.url}`,
      {
        method: request.method,
        headers,
        body:
          request.method !== "GET" && request.method !== "HEAD"
            ? JSON.stringify(request.body)
            : undefined,
      },
    );
    return dashboard(nativeRequest);
  };

  fastify.route({
    method: "GET",
    url: "/dashboard",
    handler: handleDashboardRequest,
  });

  fastify.route({
    method: "GET",
    url: "/dashboard/*",
    handler: handleDashboardRequest,
  });
```

PR #798, opened 2026-06-04
Dashboard: support dynamic credentials via a config module

harrisoncramer:
This is still WIP, will verify before I mark it as non-draft.

We run the dashboard against AWS RDS using IAM authentication, where the database password is a short-lived token (15 minutes) minted per connection. The dashboard currently only accepts a static `DATABASE_URL`, so any connection opened after the token expires fails. pg-boss itself handles this fine since node-postgres accepts `password` as a function, but there's no way to get a function through an environment variable.

This adds an optional `PGBOSS_DASHBOARD_CONFIG` env var pointing to a JS module that default-exports the database list, where each entry can carry extra connection options (password function, ssl, etc.) that are merged into the `pg.Pool` and pg-boss configs. When the variable isn't set, nothing changes — `DATABASE_URL`/`PGBOSS_SCHEMA` parsing is untouched.

Related: #730 (middleware mode) would also solve this class of problem, but connection options felt like a much smaller change.

### DIFF
--- a/packages/dashboard/README.md
+++ b/packages/dashboard/README.md
@@ -44,6 +44,7 @@ The dashboard is configured via environment variables:
 |----------|-------------|---------|
 | `DATABASE_URL` | PostgreSQL connection string(s) | `postgres://localhost/pgboss` |
 | `PGBOSS_SCHEMA` | pg-boss schema name(s) | `pgboss` |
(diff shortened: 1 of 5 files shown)


timgit:
Closing as stale.  There have been several releases to dashboard which may resolve this as well, if you'd like re-open.

Issue #714, opened 2026-02-13
TypeError in #monitor: Cannot read properties of undefined (reading 'filter')


hmwill:
## Description

In pg-boss v12.9.0, the `#monitor` method in `boss.ts` throws a recurring TypeError during the `supervise` cycle:

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at #monitor (file:///app/node_modules/.pnpm/pg-boss@12.9.0/node_modules/pg-boss/dist/boss.js:138:45)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async Boss.supervise (file:///app/node_modules/.pnpm/pg-boss@12.9.0/node_modules/pg-boss/dist/boss.js:120:17)
    at async #onSupervise (file:///app/node_modules/.pnpm/pg-boss@12.9.0/node_modules/pg-boss/dist/boss.js:88:32)
```

The error fires approximately every minute via the `error` event.

## Root Cause

In [`src/boss.ts` line ~183](https://github.com/timgit/pg-boss/blob/12.9.0/src/boss.ts#L183), `#monitor` calls `cacheQueueStats` and destructures the result:

```typescript
const { rows: rowsCacheStats } = await this.#executeSql(cacheStatsSql)
```

Then immediately calls `.filter()` on it:

```typescript
const warnings = rowsCacheStats.filter(i => i.queuedCount > (i.warningQueueSize || WARNINGS.LARGE_QUEUE.size))
```

When `#executeSql` returns a result where `rows` is `undefined`, this throws the TypeError.

A similar issue was fixed in commit 3805bb6c ("fix: unwrap sql result in case it returns array during supervise"), but that fix addressed a different code path and didn't cover the `cacheQueueStats` call in `#monitor`.

## Impact

- **Non-critical** — all scheduled jobs continue to execute and complete successfully
- **Noisy** — the error fires every ~1 minute, polluting logs
- Observed in a production deployment on Azure Container Apps with PostgreSQL (Azure Flexible Server)

## Environment

- pg-boss: 12.9.0 (also verified the code path is unchanged through 12.11.2)
- Node.js: 24
- PostgreSQL: 16 (Azure Flexible Server)

## Suggested Fix

Add a guard or default for `rowsCacheStats` similar to the pattern used in the supervise fix:

```typescript
const warnings = (rowsCacheStats || []).filter(i => i.queuedCount > (i.warningQueueSize || WARNINGS.LARGE_QUEUE.size))
```

[...]

PR #799, opened 2026-06-09
minor: Fixes the issue where results come back as an array of objects, instead of a single object

DavidTanner:
When multiple SELECT statements are executed in a single `raw()` query string, many database drivers return an array of result sets rather than a single result object. Each element in the array represents the results from one of the `SELECT` statements. For example, a query such as `SELECT * FROM users; SELECT * FROM roles;` may return an array where the first entry contains the user records and the second entry contains the role records. This behavior allows the driver to preserve the results from each statement separately, since there is no single result set that can represent multiple independent queries. The exact structure of the returned array depends on the database driver and configuration being used.
kysely and prisma have other issues, causing this same fix not to work completely.


DavidTanner:
Resolves #714

PR #801, opened 2026-06-10
Redirect legacy docsify URLs to VitePress

blue2cat:
### Changes
- Set `cleanUrls: true` in the VitePress config 
- Rewrite links from the old docs (#770) via a redirect script in the page `<head>`

### Notes: 
-  `cleanUrls: true` will work with GitHub pages to generate clean internal links that don't have `.html` in the URL
- An IIFE is added in the `<head>` tag of the new docs. If `window.location.hash` starts with docsify's routing convention of `#/`, it strips the hash prefix and redirects to the equivalent VitePress path via `window.location.replace()`
  - A slightly custom redirect is needed for the `SQL` section since one page was broken up into three
  - The `/pg-boss/` prefix prevents open redirect risk in the final URLs
  - There shouldn't be a way to generate a `#/` link via normal navigaiton within VitePress

### DIFF
--- a/docs/.vitepress/config.ts
+++ b/docs/.vitepress/config.ts
@@ -6,6 +6,17 @@ export default defineConfig({
   description: 'Queueing jobs in Postgres from Node.js like a boss',
   base: '/pg-boss/',
   lastUpdated: true,
+  head: [
+    ['script', {}, `(function(){
+      var hash = window.location.hash;
+      if(hash.startsWith('#/')){
+        var path = hash.slice(2);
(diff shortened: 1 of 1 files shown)


blue2cat:
Fairly certain the [failure](https://github.com/timgit/pg-boss/actions/runs/27304598975/job/80658931097) on https://github.com/timgit/pg-boss/pull/801/commits/8852625133ba535719f6ba3be6391b20f9325ad8 is test flake. The error is unrelated to the contents of this PR.

PR #811, opened 2026-06-14
listen notify support

timgit:


bcomnes:
Alternative/complementary idea was thinking of PRing: workers all wait for the `pollingIntervalSeconds` between jobs. This means for large numbers of jobs (even if they are short), a queue can be slow to drain unless the worker utilizes the batch feature (which complicates error state and cancellation).

I was thinking of adding an option that would make workers immediately poll once for work after finishing work, before sleeping for `pollingIntervalSeconds`. This way workers de-queue quickly when ready and work is available, and otherwise sleep if no work is available.

timgit:
I agree. A strategy I've used in the past is increasing concurrency and/or spinning up more worker microservices to drain a backlog, which would result in the same more queries/more jobs outcome. In my experience, if this flooded the db with too many queries and caused a new availability issue, I could always nuke a microservice or reduce the concurrency and redeploy. 

If this became a setting, we'd want a manual toggle to turn it off if too many workers flood the db with queries.

bcomnes:
> if this flooded the db with too many queries

I'll take your word for it!

timgit:
What if we added a config on `work()` like the following.  This could evaluate the queue backlog per worker fetch loop, and once a threshold is exceeded it would change the polling interval until after the queue recovers.

```js
{ backlog: { threshold: 100, pollingIntervalSeconds: 0.5, } }
```

bcomnes:
So when a queue gets deeper than that, workers default to fast check before sleeping? Seems reasonable if the depth check isn’t too expensive.

timgit:
@bcomnes I took a stab at a new polling behavior in this PR since the hood was open. Let me know if you get a chance to review.

> Seems reasonable if the depth check isn’t too expensive.

The depth check from the existing queue cache, so it's basically free to use.

docs/api/workers.md
```diff
@@ -43,7 +43,23 @@ The default options for `work()` is 1 job every 2 seconds.
...
+
+* **notifyPollingIntervalSeconds**, int, *(default=30)*
+
+  Polling interval used only while [LISTEN/NOTIFY](#low-latency-dispatch-with-listennotify) is active for the queue (the queue has `notify: true` and the instance listener is established). Since NOTIFY wakes workers immediately, polling only needs to run as a slow safety net, so this can be much larger than `pollingIntervalSeconds`. When notify is off or unavailable, `pollingIntervalSeconds` is used instead. Must be >=0.5 (500ms).
+
+* **burstWhenReadyExceeds**, int
+
+  When the queue's ready count — created + retry jobs runnable now, i.e. `queuedCount - deferredCount` from the cached queue stats — exceeds this value, the worker fetches continuously with no delay until it catches up; the first fetch that comes back short ends burst mode. Takes precedence over `notifyPollingIntervalSeconds` and `pollingIntervalSeconds`. Must be an integer >=1.
```
bcomnes: LGTM. What would be a reasonable setting for this? 100?

bcomnes (approved):
I didnt check the implementation that deeply, but high level design seems sane to me. Thanks for sneaking that in.

docs/api/workers.md
timgit: It's probably queue-specific

timgit: But this does make me wonder if this setting should be consolidated into `burstWhenBatchFull`, since I can't think of a good reason to keep single job pollers from bursting after it receives a job.  Currently, I have this only for workers with `batchSize > 1`

bcomnes: Hrmm maybe I didn't understand the feature correctly. Let me take a deeper look

bcomnes: I did not find time. Merge whenever you think is ready.

PR #820, opened 2026-06-20
Correctly handle manual failures in spies

kibertoad:
In manager.ts, the worker loop (#processJobs) inferred the Spy's job state from whether the handler threw, not from what actually happened in the database:

```ts  
const result = await callback(jobs)   // returned normally → assumed "completed"
await this.complete(name, jobIds, …)  // SQL only updates jobs where state = 'active'
this.#trackJobsCompleted(name, jobs, result)  // ← recorded 'completed' unconditionally
```

When a handler manually fails a job (e.g. boss.fail() on a validation error) and then returns normally, the job is no longer active, so complete() is a no-op - the DB correctly stays failed. But #trackJobsCompleted ran anyway, so the Spy reported completed.

### DIFF
--- a/src/manager.ts
+++ b/src/manager.ts
@@ -208,12 +208,24 @@ class Manager extends EventEmitter implements types.EventsMixin {
     }
   }
 
-  #trackJobsFailed<T> (name: string, jobs: types.Job<T>[], err: Error): void {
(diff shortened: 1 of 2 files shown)


CarlosGamero (approved):

timgit:
I merged your other PR for batch per-job handling, this has a conflict. I can bundle them together in a release, or this one can wait if you want.

kibertoad:
will fix in an hour or so

PR #821, opened 2026-06-20
feat(work): per-job settlement for batch handlers via perJobResults

kibertoad:
# Description

## Summary

Adds an opt-in way for a batch `work()` handler to settle each job in the batch **individually** — completing some and failing others, each with its own `output` — instead of the current all-or-nothing behavior.

Enable it with the new `perJobResults: true` work option and resolve the handler with an array of `JobResult` objects:

```js
await boss.work('resize-image', { batchSize: 10, perJobResults: true }, async (jobs) => {
  return jobs.map(job => {
    try {
      return { id: job.id, status: 'completed', output: resize(job.data) }
    } catch (err) {
      return { id: job.id, status: 'failed', output: err }
    }
  })
})
```

## Motivation

When `batchSize > 1`, a `work()` handler is all-or-nothing (`manager.ts`):

- resolve → every job in the batch is completed, and the return value is only stored as `output` when `batchSize === 1` (for larger batches per-job output is discarded);
- throw → every job in the batch fails, so a single poison job forces the whole batch to retry.

Consumers that process heterogeneous batches currently have to work around this by calling `boss.complete()` / `boss.fail()` manually inside the handler and relying on the fact that the post-handler auto-complete only matches `state = 'active'`. That is undocumented, fights the framework's own settlement, and still can't store a per-job success output. This PR makes per-job settlement a first-class, typed feature.

### Public API

- New `WorkOptions.perJobResults?: boolean` (default `false`).
- New exported types: `JobResult`, `JobResultStatus`, `PerJobWorkHandler`, `PerJobWorkWithMetadataHandler`.
- New typed `work()` overloads: when `perJobResults: true` is set, the handler must resolve with `JobResult[]` (`{ id, status: 'completed' | 'failed', output? }[]`). This mirrors how `includeMetadata: true` already switches the handler type, so the contract is enforced by the compiler and unambiguous at runtime.

### Behavior

- Each job is settled individually, **preserving its own output** (fixes the per-job-output loss for batches `> 1`).
- A `deadletter` disposition fails the job **terminally and routes it straight to the queue's configured dead letter queue**, bypassing any remaining retries (the per-job `output` travels to the dead letter job). If the queue has no dead letter queue configured, it degrades to a plain terminal failure — equivalent to a `failed` job that has exhausted its retries. This gives handlers a first-class way to send a poison job to the DLQ immediately instead of waiting out its retry budget.
- A job the handler **omits** from the result array is failed with a descriptive error (`"no disposition returned by handler"`) so it retries / dead-letters per the queue config — a result is never assumed.
- **Throwing** from the handler still fails the entire batch, exactly as before. The returned array expresses per-job failures; throwing is reserved for batch-wide errors.
- Resolving with a non-array is treated as a contract violation and fails the whole batch with a clear error.
- Default (non-`perJobResults`) behavior is **unchanged**.

### Implementation

- Per-job settlement runs in a **constant number of statements** regardless of batch size, so it scales to large batches:
  - **New, separate SQL** `completeJobsWithOutputs` and `failJobsByIdWithOutputs` apply per-id outputs via a `json_to_recordset($2::json)` join (the same binding pattern as bulk `insert`). Existing `completeJobs` / `failJobs` are left intact.
  - The shared `failJobs` CTE chain was extracted into a `failJobsBody` helper so the new fail SQL reuses it without duplication; the existing `failJobs` output is **byte-for-byte identical** after the refactor (verified).
- **Distributed backends (CockroachDB / `noMultiMutationCte`) are fully supported**: `completeJobsWithOutputsDistributed` plus the existing `decrementDependents` step, and the existing select → delete → reinsert path with per-id outputs threaded through `reinsertFailedJobs` (new optional, backward-compatible argument).

### Docs

- `docs/api/workers.md` documents the `perJobResults` option, the `JobResult[]` contract, per-job output, the omitted-job rule, and that throwing still fails the whole batch.

## Backward compatibility

Fully backward compatible. The feature is opt-in via `perJobResults`; existing handlers, the existing `complete` / `fail` SQL, and default batch semantics are unchanged.

### DIFF
--- a/docs/api/workers.md
+++ b/docs/api/workers.md
@@ -36,14 +36,17 @@ The default options for `work()` is 1 job every 2 seconds.
         const output = resize(job.data)
         return { id: job.id, status: 'completed', output }
       } catch (err) {
-        return { id: job.id, status: 'failed', output: err }
+        return err.fatal
+          ? { id: job.id, status: 'deadletter', output: err }
+          : { id: job.id, status: 'failed', output: err }
       }
     })
   })
   ```
 
-  Each `JobResult` is `{ id, status, output? }` where `id` matches a job from the batch, `status` is `'completed'` or `'failed'`, and `output` is stored on that job (the completion result, or the failure detail). Notes:
+  Each `JobResult` is `{ id, status, output? }` where `id` matches a job from the batch, `status` is `'completed'`, `'failed'`, or `'deadletter'`, and `output` is stored on that job (the completion result, or the failure detail). Notes:
 
+  - **`deadletter`** fails the job terminally and routes it straight to the queue's configured dead letter queue, bypassing any remaining retries (the `output` travels to the dead letter job). If the queue has no dead letter queue configured, the job simply fails terminally — equivalent to a `failed` job that has exhausted its retries.
   - Any job in the batch the handler omits from the array is **failed** with a descriptive error so it retries (or dead-letters) per the queue config — a returned result is never assumed.
   - **Throwing** from the handler still fails the entire batch, exactly as without `perJobResults`. Use the returned array to express per-job failures; reserve throwing for batch-wide errors.
   - Resolving with anything other than an array is treated as a contract violation and fails the whole batch.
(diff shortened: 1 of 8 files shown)


kibertoad:
@timgit Could you take a look at this one?

PR #825, opened 2026-06-24
Dashboard: support dynamic views on jobs page

kalyvasio:
## Summary

Adds configurable, shareable job table views to the dashboard Jobs page and each queue detail page.

Users can manage which job fields appear in the table, choose friendly column names, include built-in job metadata, full JSON payloads, nested `data.*` / `output.*` JSON paths, and supported DB-backed fields, then copy a URL that preserves the selected view alongside the active filters.

## What changed

- Added a reusable job column model with route-specific defaults:
  - Jobs page: `id`, `name`, `state`, `retries`, `createdOn`
  - Queue detail page: `id`, `state`, `priority`, `retries`, `createdOn`
- Added a `Manage view` editor for adding, renaming, removing, resetting, and selecting columns, plus a `Copy link` action for shareable filtered views.
- Replaced fixed job table columns on both Jobs surfaces with dynamic column rendering while preserving specialized display for job links, state badges, retry counts, and dates.
- Supports built-in columns such as `id`, `name`, `state`, `priority`, `retryCount`, `retryLimit`, `startAfter`, `startedOn`, `completedOn`, `createdOn`, and `singletonKey`.
- Supports full JSON columns (`data`, `output`), nested JSON paths such as `data.tenantId` / `output.status`, and extra DB-backed fields such as `groupId`, `groupTier`, `deadLetter`, and `policy`.
- Serializes non-default views as repeated `col=path|name` params, keeps default views out of the URL, and preserves selected columns across filter and pagination changes.
- Adds server-side SQL projections for requested custom columns while keeping default job-list queries lightweight when no extra columns are selected.

## Why

The Jobs page and queue detail page previously had fixed table layouts, which made it hard to inspect job-specific metadata without opening individual job details.
In our setup, we manage multiple queues, each with its own policies and responsibility for executing a specific set of jobs. 
The actual job name is stored inside the job payload, so monitoring these jobs effectively requires exposing fields such as `data.name` and tenant-specific metadata like `data.tenantId` directly in the job table.

## Example
Screen Recording 2026-06-24 at 12 22 44 PM

@timgit tell me what you think


timgit:
Including PR review from Opus.  Let me know your thoughts.  

# Review

This PR replaces the fixed job tables on the Jobs page and queue-detail page with
user-configurable columns. Users pick built-in fields, full `data`/`output` JSON,
nested `data.*`/`output.*` paths, or extra DB-backed columns; selections serialize
into shareable `col=path|name` URL params and drive server-side SQL projections.

The SQL projection path is well-guarded — paths are validated server-side via
`parseJobColumns` before reaching SQL, nested segments are escaped with
`quoteSqlString`, aliases with `quoteSqlIdentifier`, and all the
`EXTRA_DB_JOB_COLUMNS` (`group_id`, `group_tier`, `dead_letter`, `policy`, etc.)
genuinely exist on the `job` table, so no injection or missing-column crash was found.

## Findings

### 1. Zero-valued numeric built-in columns render as `—` (regression for Priority)

**File:** `packages/dashboard/app/components/job-column-cell.tsx:86`
**Severity:** High

`JobColumnCell` has no `case 'priority'`, so `priority` falls through to the
`default` branch, which calls `getRowCellValue`. That helper does
`if (!value) return null` (`job-columns.ts:548`), and `!0 === true`, so a job with
the default `priority` of `0` renders `—` instead of `0`.

Priority is a **default column on the queue-detail page**
(`DEFAULT_QUEUE_JOB_COLUMN_PATHS`), so nearly every job will show `—` for priority —
the old `JobRow` rendered `{job.priority}` correctly. The same falsy bug hits
`retryCount`/`retryLimit` if added as standalone columns. (Nested `data.*` paths are
unaffected because `#>>` returns text, so `"0"` is truthy.)

### 2. Distinct paths that collapse to the same prop name silently show duplicated data

**File:** `packages/dashboard/app/lib/queries.server.ts:625`
**Severity:** Low/Medium

`buildJobColumnProjections` dedups by `jobColumnProp(col.path)`. Two different paths
can map to the same prop — e.g. `data.tenant.id` and `data.tenantId` both become
`dataTenantId`. Only the first is projected, but both header cells render; the second
column reads the first column's value via `getRowCellValue(row, 'dataTenantId')`, so
it displays the wrong field's data under its own header with no error. Edge case
(requires a camelCase collision) but produces silently incorrect output.

## Recommendation

The first finding is worth fixing before merge (it visibly breaks a default column);
the second is lower-severity. Everything else — URL round-tripping, SQL escaping,
colSpans, share-URL `db` preservation, and test coverage — looked solid.

kalyvasio:
@timgit I think i have addressed both of the comments...PTAL when you have the time

PR #835, opened 2026-07-03
fix: scope warning thresholds to each instance

blue2cat:
### Description

Warning thresholds were stored on a module-level `WARNINGS` constant and mutated in the `Boss` constructor. Because the object is shared across every `PgBoss` instance in a process, constructing one instance with a custom `warningSlowQuerySeconds` or `warningQueueSize` changed the thresholds for all other instances.

This scopes the resolved thresholds to per-instance private fields so each `Boss` uses its own config and the module-level defaults stay immutable.

### Code Changes

- Add `#slowQuerySeconds` and `#largeQueueSize` private fields to `Boss`, resolved from config in the constructor with a fallback to the `WARNINGS` defaults.
- Stop mutating `WARNINGS.SLOW_QUERY.seconds` and `WARNINGS.LARGE_QUEUE.size` in the constructor.
- Read the instance fields in the slow-query and large-queue checks instead of the shared constant.
- Add a test asserting that constructing one instance with `warningQueueSize: 1` does not cause another instance to emit a queue warning.

### Developer Notes

The `WARNINGS` constant is now used only for default values and the warning message strings, so it is left as a shared module-level object.

### DIFF
--- a/docs/api/constructor.md
+++ b/docs/api/constructor.md
@@ -107,6 +107,14 @@ The following configuration options should not normally need to be changed, but 
 
   How often the background flow resolver runs to unblock dependent jobs (created via [`flow()`](./jobs.md#flowjobs-options)) whose parents have completed. Completing a job no longer unblocks its dependents inline; this resolver handles it shortly after, off the completion hot path. Only runs when `supervise` is enabled.
 
+* **warningSlowQuerySeconds**, int, default 30
+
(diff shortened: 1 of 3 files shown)


blue2cat:
@timgit would you prefer I file an issue before fixing things like this?

timgit:
> @timgit would you prefer I file an issue before fixing things like this?

No, free-standing PRs like this are great

PR #836, opened 2026-07-03
Increase archive to 25k

valigula:


timgit:
Would you guys be interested in pulling upstream v12 back into your app? If there's a blocker, let me know.

timgit:
See https://github.com/supabase/storage/pull/1198

PR #844, opened 2026-07-08
docs: clarify retryBackoff requires a non-zero retryDelay

### LINKED ISSUE #839: retryBackoff: true with unset retryDelay produces zero backoff (retries fire instantly), contradicting the documented default
### Description
The docs for `retryBackoff` (`docs/api/queues.md`) state: *"Enables exponential backoff... Sets initial retryDelay to 1 if not set."* No code implements that default. `retryDelay` defaults to `0` (`QUEUE_DEFAULTS.retry_delay = 0`, column default `0`, queue-create/insert both `COALESCE(..., 0)`), and the backoff formula multiplies by `retry_delay`. So with backoff enabled and `retryDelay` unset the delay is `0 * (...) = 0` — the job retries immediately with no backoff, which defeats the purpose.

Regression from the fix for #672 (PR #677), which changed the additive `+` to a multiplicative `*`. Before, `retry_delay=0` still yielded a nonzero first-retry delay via the additive term; after, the zero-default path became "no delay whatsoever."

### Steps to reproduce
```js
const boss = new PgBoss({ connectionString });
await boss.start();
await boss.createQueue('q', { retryLimit: 5, retryBackoff: true }); // no retryDelay
const id = await boss.send('q', {});
const [job] = await boss.fetch('q');
const t0 = Date.now();
await boss.fail('q', job.id);
const row = await boss.getJobById('q', id);
console.log((new Date(row.startAfter) - t0) / 1000); // 0.00 (immediate) -- expected ~1-2s
```
(Verified against real Postgres 17.)

### Root cause
`src/plans.ts` -- `QUEUE_DEFAULTS.retry_delay = 0`, the `create_queue`/job-insert `COALESCE(..., 0)`, and the backoff formula `retryDelay * 2^retryCount` in `failJobsBody` (mirrored in `manager.ts`). Nothing sets `retryDelay = 1` when `retryBackoff` is true and `retryDelay` is unset — `attorney.ts` only asserts on the options.

### Suggested fix
Either (a) implement the documented behavior: set `retryDelay = 1` when `retryBackoff === true` and `retryDelay` is not provided (queue-create + per-job); or (b) if the contract changed, fix the docs to state backoff requires a nonzero `retryDelay`. Option (a) matches the long-standing documented contract.

blue2cat:
## Description

Fixes #839.

With `retryBackoff` enabled and `retryDelay` left at its default,
every computed delay is `0 * 2^retryCount = 0`, so retries fire immediately
instead of backing off.

This updates the documentation to match actual behavior: `retryDelay` must be
set to a non-zero value for backoff to have any effect. It also corrects typos
in the backoff formula shown in the docs and JSDoc.

## Code Changes

- `docs/api/jobs.md`: removed the incorrect "Sets initial retryDelay to 1 if not
  set" claim, noted that `retryDelay` must be non-zero for backoff to take
  effect, and fixed the backoff formula (`Math.Min` to `Math.min`, added the
  missing `**` and the `retryCount + 1` used by the code).
- `docs/api/queues.md`: same changes as `jobs.md`.
- `src/types.ts`: updated the `retryDelay` and `retryBackoff` JSDoc to match, and
  corrected the same formula typos shown in editor tooltips.

### DIFF
--- a/docs/api/jobs.md
+++ b/docs/api/jobs.md
@@ -36,11 +36,11 @@ Creates a new job and returns the job id.
 
 * **retryDelay**, int
 
-  Default: 0. Delay between retries of failed jobs, in seconds.
+  Default: 0. Delay between retries of failed jobs, in seconds. When `retryBackoff` is enabled, this is the base delay that the exponential backoff multiplies, so it must be non-zero for backoff to have any effect.
 
 * **retryBackoff**, bool
 
-  Default: false. Enables exponential backoff retries based on retryDelay instead of a fixed delay. Sets initial retryDelay to 1 if not set. A simplified function to get the delay between runs is: `retryDelay * 2 ^ retryCount` with some jitter. The full function to determine the backoff delay is `Math.min(retryDelayMax, retryDelay * (2 ** Math.Min(16, retryCount) / 2 + 2 ** Math.Min(16, retryCount) / 2 * Math.random()))`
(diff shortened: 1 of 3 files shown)


timgit:
This is a regression bug I think.  I used to default retryDelay to 1 if it wasn't set

blue2cat:
Ah that's good to know, I'll close this docs PR in favor of a fix.

PR #845, opened 2026-07-09
Fix groupConcurrency fetch plan regression

### LINKED ISSUE #842: Query planner de-optimization with large number of jobs when using group concurrency
Just ran into a situation where a a group id with a large number of waiting jobs (~10k) in a queue that was using the group concurrency feature caused the job selection query to to de-optimize from a 4ms to 400ms query, causing DB CPU load to spike to about 100% causing overall system degradation. Still gathering details and ideas on how to improve, but wanted to start an issue thread to capture and discuss.

bcomnes:
Closes https://github.com/timgit/pg-boss/issues/842

## Problem

Queues using `groupConcurrency` can hit a severe PostgreSQL plan regression when:

- A saturated group has many pending jobs near the front of the queue.
- Eligible jobs from other groups are behind that backlog.
- Table statistics underestimate the number of active groups.

The original fetch query materialized `active_group_counts`, then joined that relation both while finding candidates and after `LIMIT`. `MATERIALIZED` ensures the CTE is computed once, but PostgreSQL can still repeatedly scan its result. With stale statistics, it can choose nested-loop plans that rescan active counts and group ranking once per candidate or job-table row.

The regression fixture produces roughly 12,000 repeated scans in the pathological case.

## Changes

For `groupConcurrency: 1`, the fetch query uses `NOT EXISTS` to determine whether a candidate's group already has an active job.

For `groupConcurrency > 1` and tiered limits, the query now:

- Groups active jobs once and aggregates their counts into one materialized JSONB object mapping `group_id` to `active_cnt`.
- Reads candidate counts through keyed, uncorrelated scalar InitPlans instead of joining a multi-row count relation in the candidate path.
- Carries `active_cnt` through the post-`LIMIT` ranking rather than joining active counts again.
- Supplies the final filtered IDs to the update with `id = ANY (ARRAY(SELECT ...))`, preventing PostgreSQL from repeatedly executing the ranking query.
- Keeps tier limits in the JSONB parameter, making generated SQL independent of configured tier count.

This does not add tables, migrations, or persistent count state.

## Regression coverage

The first commit adds plan regression tests around pg-boss's generated fetch update. The fixture includes a saturated backlog, eligible jobs behind it, many active groups, and both current and deliberately stale statistics.

Coverage includes:

- `groupConcurrency: 1` and `groupConcurrency: 2`
- Current and stale statistics
- Batch size 1 and a large batch
- Correct claimed-job counts
- No repeated active-count CTE scan
- No repeated group-ranking execution
- No repeated job-table scan caused by the group-concurrency subplan
- Tiered query shape remaining independent of configured tier count
- Existing normal and tiered group-concurrency behavior

Before the fix, the stale-statistics cases reproduce the repeated-scan plan. The second commit changes the query shape and makes the full suite pass.

## Validation

- Standard CI and coverage
- Distributed test suite
- PGlite test suite
- CockroachDB test suite
- Existing group-concurrency behavior tests
- Local PostgreSQL 17 plan and performance comparison across the original, array, and current map implementations

The detailed three-way benchmark, plan-loop counts, representative SQL for all three implementations, and remaining tradeoffs are documented in [this PR discussion](https://github.com/timgit/pg-boss/pull/845#issuecomment-5030088177).

### DIFF
--- a/src/plans.ts
+++ b/src/plans.ts
@@ -1343,30 +1343,42 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
   const hasGroupConcurrency = groupConcurrency != null
   const hasMinPriority = minPriority != null
   const hasMaxPriority = maxPriority != null
+  const groupConcurrencyConfig = hasGroupConcurrency
+    ? (typeof groupConcurrency === 'number' ? { default: groupConcurrency } : groupConcurrency)
+    : null
   const hasTiers = hasGroupConcurrency &&
-    typeof groupConcurrency === 'object' &&
-    groupConcurrency.tiers &&
-    Object.keys(groupConcurrency.tiers).length > 0
+    groupConcurrencyConfig?.tiers &&
+    Object.keys(groupConcurrencyConfig.tiers).length > 0
+  const hasSingleGroupConcurrency = hasGroupConcurrency && !hasTiers && groupConcurrencyConfig?.default === 1
+  const hasActiveGroupCounts = hasGroupConcurrency && !hasSingleGroupConcurrency
 
   const params = buildFetchParams(options)
+  const groupLimit = hasTiers
+    ? `COALESCE((${params.tiersParam} ->> group_tier)::int, ${params.defaultGroupLimitParam})`
+    : params.defaultGroupLimitParam
+  const activeGroupCountExpression = hasActiveGroupCounts
+    ? 'COALESCE(((SELECT counts FROM active_group_count_map) ->> j.group_id)::int, 0)'
+    : ''
 
   const selectCols = [
     'j.id',
     singletonFetch ? 'j.singleton_key' : '',
-    hasGroupConcurrency ? 'j.group_id, j.group_tier' : ''
+    hasGroupConcurrency ? 'j.group_id, j.group_tier' : '',
+    hasActiveGroupCounts ? `${activeGroupCountExpression} as active_cnt` : ''
   ].filter(Boolean).join(', ')
 
-  // MATERIALIZED forces Postgres to compute this aggregation once and cache the
-  // result. Without it, Postgres 12+ may inline the CTE and re-evaluate the
-  // COUNT query at each reference site. active_group_counts is referenced twice:
-  // once in the next CTE join (to pre-filter saturated groups before LIMIT) and
-  // once in group_ranking (to enforce the per-batch concurrency limit).
-  const activeGroupCountsCte = hasGroupConcurrency
-    ? `active_group_counts AS MATERIALIZED (
-        SELECT group_id, COUNT(*)::int as active_cnt
-        FROM ${schema}.${table}
-        WHERE name = '${name}' AND state = '${JOB_STATES.active}' AND group_id IS NOT NULL
-        GROUP BY group_id
+  // For limits above 1, aggregate active counts into a single JSONB value. Each
+  // candidate uses a keyed lookup through an uncorrelated InitPlan, so the planner
+  // cannot turn stale active-group estimates into a per-candidate relation scan.
+  const activeGroupCountMapCte = hasActiveGroupCounts
+    ? `active_group_count_map AS MATERIALIZED (
+        SELECT COALESCE(jsonb_object_agg(group_id, active_cnt), '{}'::jsonb) as counts
+        FROM (
+          SELECT group_id, COUNT(*)::int as active_cnt
+          FROM ${schema}.${table}
+          WHERE name = '${name}' AND state = '${JOB_STATES.active}' AND group_id IS NOT NULL
+          GROUP BY group_id
+        ) active_groups
       ), `
     : ''
 
@@ -1375,8 +1387,22 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
   const lockClause = noSkipLocked ? '' : 'FOR UPDATE OF j SKIP LOCKED'
 
   // Column references are qualified with j. throughout so both the base case and
-  // the groupConcurrency branch (which joins active_group_counts) share one set of
-  // expressions. The join introduces agc.group_id which would otherwise be ambiguous.
+  // the groupConcurrency branches share one set of expressions.
+  const groupConcurrencyFilter = hasGroupConcurrency
+    ? hasSingleGroupConcurrency
+      ? `(j.group_id IS NULL
+            OR NOT EXISTS (
+              SELECT 1
+              FROM ${schema}.${table} active_group_probe
+              WHERE active_group_probe.name = '${name}'
+                AND active_group_probe.state = '${JOB_STATES.active}'
+                AND active_group_probe.group_id IS NOT NULL
+                AND active_group_probe.group_id = j.group_id
+            ))`
+      : `(j.group_id IS NULL
+            OR ${activeGroupCountExpression} < ${groupLimit})`
+    : ''
+
   const whereConditions = [
     `j.name = '${name}'`,
     `j.state < '${JOB_STATES.active}'`,
@@ -1391,20 +1417,13 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
     hasIgnoreGroups ? `(j.group_id IS NULL OR j.group_id <> ALL(${params.ignoreGroupsParam}))` : '',
     hasMinPriority ? `j.priority >= ${params.minPriorityParam}` : '',
     hasMaxPriority ? `j.priority <= ${params.maxPriorityParam}` : '',
-    hasGroupConcurrency
-      ? `(j.group_id IS NULL
-            OR agc.active_cnt IS NULL
-            OR agc.active_cnt < ${hasTiers
-              ? `COALESCE((${params.tiersParam} ->> j.group_tier)::int, ${params.defaultGroupLimitParam})`
-              : params.defaultGroupLimitParam})`
-      : ''
+    groupConcurrencyFilter
   ].filter(Boolean).join('\n          AND ')
 
   const nextCte = `
       next AS (
         SELECT ${selectCols}
         FROM ${schema}.${table} j
-        ${hasGroupConcurrency ? 'LEFT JOIN active_group_counts agc ON j.group_id = agc.group_id' : ''}
         WHERE ${whereConditions}
         ORDER BY ${priority ? 'j.priority desc, ' : ''}${orderByCreatedOn ? 'j.created_on, ' : ''}j.id
         LIMIT ${limit}
@@ -1413,7 +1432,7 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
 
   const singletonCte = singletonFetch
     ? `, singleton_ranking AS (
-        SELECT id, ${hasGroupConcurrency ? 'group_id, group_tier, ' : ''}
+        SELECT id, ${hasGroupConcurrency ? 'group_id, group_tier, ' : ''}${hasActiveGroupCounts ? 'active_cnt, ' : ''}
           row_number() OVER (PARTITION BY singleton_key) as singleton_rn
         FROM next
       )`
@@ -1427,25 +1446,30 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
           , t.group_tier
           ${singletonFetch ? ', singleton_rn' : ''}
           , ROW_NUMBER() OVER (PARTITION BY t.group_id ORDER BY t.id) as group_rn
-          , COALESCE(agc.active_cnt, 0) as active_cnt
+          , ${hasActiveGroupCounts ? 't.active_cnt' : '0'} as active_cnt
         FROM ${singletonFetch ? 'singleton_ranking' : 'next'} t
-        LEFT JOIN active_group_counts agc ON t.group_id = agc.group_id
         ${singletonFetch ? 'WHERE singleton_rn = 1' : ''}
       ),
       group_filtered AS (
         SELECT id FROM group_ranking
         WHERE group_id IS NULL
-          OR (active_cnt + group_rn) <= ${hasTiers
-          ? `COALESCE((${params.tiersParam} ->> group_tier)::int, ${params.defaultGroupLimitParam})`
-          : params.defaultGroupLimitParam}
+          OR (active_cnt + group_rn) <= ${groupLimit}
       )`
     : ''
 
-  const finalCte = (hasGroupConcurrency)
+  const finalCte = hasGroupConcurrency
     ? 'group_filtered'
     : (singletonFetch)
         ? 'singleton_ranking'
         : 'next'
+
+  // An uncorrelated array InitPlan makes the selected ids a one-time input to the
+  // UPDATE. Without it, stale estimates can make Postgres put the inlined ranking
+  // query on the inner side of a nested loop and execute it once per job table row.
+  const updateSource = hasGroupConcurrency ? '' : `FROM ${finalCte}`
+  const updateMatch = hasGroupConcurrency
+    ? `j.id = ANY (ARRAY(SELECT id FROM ${finalCte}))`
+    : `j.id = ${finalCte}.id`
 
   // Without SKIP LOCKED, add a state check to prevent duplicate processing
   // when multiple workers try to claim the same jobs concurrently
@@ -1454,7 +1478,7 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
   return {
     text: `
       WITH
-      ${activeGroupCountsCte}
+      ${activeGroupCountMapCte}
       ${nextCte}
       ${singletonCte}
       ${groupConcurrencyCtes}
@@ -1463,8 +1487,8 @@ export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): S
         started_on = now(),
         heartbeat_on = now(),
         retry_count = CASE WHEN started_on IS NOT NULL THEN retry_count + 1 ELSE retry_count END
-      FROM ${finalCte}
-      WHERE name = '${name}' AND j.id = ${finalCte}.id
+      ${updateSource}
+      WHERE name = '${name}' AND ${updateMatch}
       ${singletonFetch && !hasGroupConcurrency ? 'AND singleton_rn = 1' : ''}
       ${distributedStateCheck}
       RETURNING j.${includeMetadata ? JOB_COLUMNS_ALL : JOB_COLUMNS_MIN}
--- a/test/concurrencyGroupTest.ts
+++ b/test/concurrencyGroupTest.ts
@@ -190,6 +190,71 @@ describe('groupConcurrency', function () {
     expect(maxByGroup[freeGroup]).toBeGreaterThanOrEqual(1)
   })
 
+  it('should preserve tiered batch capacity when groups are already active', async function () {
+    ctx.boss = await helper.start(ctx.bossConfig)
+
+    const schema = ctx.schema
+    const queueName = ctx.schema
+    const db = await helper.getDb()
+
+    try {
+      // vip-full is already at its tier limit, so its pending jobs must be skipped.
+      await db.executeSql(`
+        INSERT INTO ${schema}.job_common (name, data, state, group_id, group_tier, start_after, created_on, started_on)
+        SELECT $1, jsonb_build_object('group', 'vip-full'), 'active'::${schema}.job_state, 'vip-full', 'vip', now() - interval '5 minutes', now() - interval '5 minutes', now() - interval '5 minutes'
+        FROM generate_series(1, 3)
+      `, [queueName])
+
+      // vip has one active job and a tier limit of three, leaving two slots in this fetch batch.
+      await db.executeSql(`
+        INSERT INTO ${schema}.job_common (name, data, state, group_id, group_tier, start_after, created_on, started_on)
+        VALUES ($1, jsonb_build_object('group', 'vip'), 'active'::${schema}.job_state, 'vip', 'vip', now() - interval '5 minutes', now() - interval '5 minutes', now() - interval '5 minutes')
+      `, [queueName])
+
+      await db.executeSql(`
+        INSERT INTO ${schema}.job_common (name, data, state, group_id, group_tier, start_after, created_on)
+        SELECT $1, jsonb_build_object('group', 'vip-full'), 'created'::${schema}.job_state, 'vip-full', 'vip', now() - interval '5 minutes', now() - interval '4 minutes' + (g * interval '1 millisecond')
+        FROM generate_series(1, 3) g
+      `, [queueName])
+
+      await db.executeSql(`
+        INSERT INTO ${schema}.job_common (name, data, state, group_id, group_tier, start_after, created_on)
+        SELECT $1, jsonb_build_object('group', 'vip'), 'created'::${schema}.job_state, 'vip', 'vip', now() - interval '5 minutes', now() - interval '3 minutes' + (g * interval '1 millisecond')
+        FROM generate_series(1, 3) g
+      `, [queueName])
+
+      await db.executeSql(`
+        INSERT INTO ${schema}.job_common (name, data, state, group_id, start_after, created_on)
+        SELECT $1, jsonb_build_object('group', 'default'), 'created'::${schema}.job_state, 'default', now() - interval '5 minutes', now() - interval '2 minutes' + (g * interval '1 millisecond')
+        FROM generate_series(1, 2) g
+      `, [queueName])
+
+      const jobs = await ctx.boss.fetch(queueName, {
+        batchSize: 10,
+        includeMetadata: true,
+        priority: false,
+        orderByCreatedOn: true,
+        groupConcurrency: {
+          default: 1,
+          tiers: { vip: 3 }
+        }
+      })
+
+      const groupCounts = jobs.reduce<Record<string, number>>((counts, job) => {
+        const group = (job.data as { group: string }).group
+        counts[group] = (counts[group] ?? 0) + 1
+        return counts
+      }, {})
+
+      expect(jobs).toHaveLength(3)
+      expect(groupCounts['vip-full'] ?? 0).toBe(0)
+      expect(groupCounts.vip).toBe(2)
+      expect(groupCounts.default).toBe(1)
+    } finally {
+      await db.close()
+    }
+  })
+
   it('should allow jobs without group to bypass group concurrency limits', async function () {
     ctx.boss = await helper.start({ ...ctx.bossConfig, __test__enableSpies: true })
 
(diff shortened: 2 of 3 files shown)


bcomnes:
This looks reasonable, but I haven't tested in the wild yet. I plan on doing that if you want to sit on it for a few days.

timgit:
Thanks for opening and your research.  I've been looking at this with claude and attached some findings below.  It suggested an optimization with ARRAY(), but there's a caveat that it needs to remain below 500 **distinct active group ids**.  That seems quite an enormous amount of concurrency, and it's likely no one would ever reach that level in order for the perf benefit to start degrading. Seems worth some real world testing if you're up for it.  DB query planning and optimization is always a dark art....

---
Empirical findings on a local repro of #842 (PG18, ~12k saturated-pending fixture, stale stats):

| Approach | Time | Hot-path shape | Robust to stale stats? |
|---|---|---|---|
| **current master** — `MATERIALIZED` count CTE, joined twice | ~7ms good / **~286ms** bad | hash join → nested-loop CTE rescan `loops=12001` | **no** — this is #842 |
| `groupConcurrency: 1` `NOT EXISTS` (this PR) | ~3.7ms | de-correlated, single `job_i7` scan | yes — already optimal |
| `> 1` / tiers lateral (this PR) | ~33–37ms | probe `loops=12001` | yes, but always slow |
| correlated `saturated_groups` `NOT EXISTS` | ~33–55ms | pushed down, aggregate `loops=12k` | **no** — relapses into same trap |
| **`ARRAY()` InitPlan + post-`LIMIT` count join** | **~3.8–4.9ms** | `HashAggregate loops=1`, uncorrelated | **yes** — InitPlan can't be pushed down |

---

The query in master today materializes `active_group_counts` (all active groups) and `LEFT JOIN`s it twice. `MATERIALIZED` only forces the CTE to be *computed* once — the join to it is still the planner's choice. On good stats it hashes (~7ms here); when stats are stale it goes nested-loop and rescans the tuplestore once per candidate row — `CTE Scan ... loops=12001`, ~286ms on this fixture (and much worse in prod with a bigger CTE / deeper backlog). So the failure mode is a *plan cliff*: same query, 40× swing depending on stats. (Note: on PG18 my fixture naturally picked the good hash plan; I had to force the nested loop to reproduce the cliff. The pathology is stats/version-dependent — the PR's regression test pins the exact condition that triggers it.)

**On the current fix.** The `groupConcurrency: 1` path is already optimal — the `NOT EXISTS` de-correlates to a single `job_i7` scan (~3.7ms here). The `> 1` / tiers lateral is correct and robust, but because the count is a `SELECT`-list scalar the filter depends on, the planner must evaluate it per candidate row — ~12k bounded index probes, ~33–37ms.

**Heads up on the "obvious" alternative.** A precomputed `saturated_groups` CTE with a correlated `NOT EXISTS` does *not* reliably help — the planner pushes `group_id = j.group_id` down through the `GROUP BY` and re-runs the aggregate per row (`loops=12005`, ~33–55ms). It relapses into exactly the stale-stats trap this PR fixes. Dead end.

**What does work — an uncorrelated `ARRAY()` InitPlan**, split into two phases:
- Hot path (pre-`LIMIT`): filter saturated groups with `j.group_id <> ALL(ARRAY(SELECT ... HAVING COUNT(*) >= limit))`. `ARRAY(subquery)` is an InitPlan — computed once, can't be pushed down, immune to the misestimate.
- Cold path (`group_ranking`, post-`LIMIT`): keep the exact-count join against a small `active_group_counts`, which now runs on ≤ `limit` rows.

Measured ~3.8–4.9ms (tiers + batch included), ~7–10× faster than the lateral and stable, because the original bug lived entirely in doing the saturation check pre-`LIMIT` over the full backlog. Correct on batch (`limit>1`) and tiers (`{vip:3}`, default 1): claims `vip-full=0, vip=2, def=1`. Tradeoff: `<> ALL(array)` is O(saturated-groups) per candidate row — fine while the saturated set is small (bounded by active capacity).

**But the tradeoff is regime-dependent** — and it favors your lateral as the general choice. `<> ALL(array)` is O(saturated-groups) per candidate row, vs the lateral's bounded O(limit) index probe. Measured on ~12k backlog as I vary S = number of saturated groups:

| S (saturated groups) | ARRAY InitPlan | lateral (this PR) |
|---|---|---|
| 1 (classic #842: one hot tenant) | **~3.8ms** | ~33ms |
| 500 | ~54ms | ~50ms |
| 3000 | **~285ms** | **~59ms** |

The lateral stays roughly flat because per-row cost is bounded by the concurrency limit, independent of S. The ARRAY has a lower floor (S small) but a higher ceiling (S large). Crossover ≈ 500 saturated groups.

**What S=500 means physically.** S = groups saturated *at that instant* — each holds ≥ `limit` active jobs, and every active job is an occupied worker slot. So the crossover is ~500 (at `groupConcurrency:1`) to ~1,000 (at `:2`) *concurrently in-flight jobs on one queue*. S is bounded by live worker throughput, not total group/tenant count — 10k tenants doesn't mean S=10k, only the ones actively saturated with backlog count. Most single-queue deployments run S in the tens, maybe low hundreds. S=3000 (where the lateral is 5× ahead) is ~6,000 concurrent active jobs — essentially nobody.

So the practical read favors the ARRAY: it's ~10× faster for realistic concurrency, and the O(S) term is noise until you're running ~1k+ concurrent jobs on a single queue. The lateral's flat ceiling is real insurance, but it costs ~10× on the common case (33ms vs 3.8ms) to guard a tail few reach.

None of this blocks the PR — the lateral is correct and robust, and "bounded worst case" is a legitimate design choice. The call is: optimize the common case (ARRAY, ~10× faster below ~1k concurrent jobs) or guard the extreme-concurrency tail (lateral). Worth a conscious decision rather than defaulting to whichever we wrote first.

### Standalone repro (PG14+, run with `psql`)

```sql
DROP SCHEMA IF EXISTS repro CASCADE;
CREATE SCHEMA repro; SET search_path = repro;
CREATE TYPE repro.job_state AS ENUM ('created','retry','active','completed','cancelled','failed');
CREATE TABLE repro.job (
  id bigserial primary key, name text not null, state repro.job_state not null,
  group_id text, group_tier text, priority int not null default 0,
  start_after timestamptz not null default now(), created_on timestamptz not null default now(),
  started_on timestamptz, heartbeat_on timestamptz, blocked boolean not null default false,
  retry_count int not null default 0);
CREATE INDEX job_i7 ON repro.job (name, group_id) WHERE state='active' AND group_id IS NOT NULL;
CREATE INDEX job_fetch ON repro.job (name, created_on, id) WHERE state<'active';

-- concurrency=2 fixture. Seed 2 saturated actives, ANALYZE, THEN add the rest so
-- planner stats stay stale (the #842 condition).
INSERT INTO repro.job (name,state,group_id,start_after,created_on,started_on)
SELECT 'Q','active','group-saturated',now()-interval '5 min',now()-interval '5 min',now()-interval '5 min' FROM generate_series(1,2);
ANALYZE repro.job;
INSERT INTO repro.job (name,state,group_id,start_after,created_on,started_on)
SELECT 'Q','active','active-group-'||g,now()-interval '5 min',now()-interval '5 min',now()-interval '5 min' FROM generate_series(1,114) g;
INSERT INTO repro.job (name,state,group_id,start_after,created_on)
SELECT 'Q','created','group-saturated',now()-interval '5 min',now()-interval '4 min'+(g*interval '1 ms') FROM generate_series(1,12000) g;
INSERT INTO repro.job (name,state,group_id,start_after,created_on)
SELECT 'Q','created','available-'||g,now()-interval '5 min',now()-interval '3 min'+(g*interval '1 ms') FROM generate_series(1,36) g;
-- no final ANALYZE: stats stay stale on purpose

\echo '--- 0) current MASTER: MATERIALIZED CTE. On PG18 this fixture hashes (~7ms).'
\echo '       Force the #842 plan (nested-loop CTE rescan) to see the cliff: ~286ms ---'
SET enable_hashjoin=off; SET enable_mergejoin=off;
EXPLAIN (ANALYZE, BUFFERS)
WITH active_group_counts AS MATERIALIZED (
  SELECT group_id, COUNT(*)::int active_cnt FROM repro.job
  WHERE name='Q' AND state='active' AND group_id IS NOT NULL GROUP BY group_id),
next AS (
  SELECT j.id, j.group_id, j.group_tier FROM repro.job j
  LEFT JOIN active_group_counts agc ON j.group_id = agc.group_id
  WHERE j.name='Q' AND j.state<'active' AND NOT j.blocked AND j.start_after<now()
    AND (j.group_id IS NULL OR agc.active_cnt IS NULL OR agc.active_cnt < 2)
  ORDER BY j.created_on, j.id LIMIT 1 FOR UPDATE OF j SKIP LOCKED),
group_ranking AS (
  SELECT t.id, t.group_id, t.group_tier, ROW_NUMBER() OVER (PARTITION BY t.group_id ORDER BY t.id) group_rn,
    COALESCE(agc.active_cnt,0) active_cnt
  FROM next t LEFT JOIN active_group_counts agc ON t.group_id = agc.group_id),
group_filtered AS (SELECT id FROM group_ranking WHERE group_id IS NULL OR (active_cnt+group_rn) <= 2)
SELECT * FROM group_filtered;
RESET enable_hashjoin; RESET enable_mergejoin;

\echo '--- A) current PR lateral (conc=2): ~33-37ms, probe loops=12001 ---'
EXPLAIN (ANALYZE, BUFFERS)
WITH next AS (
  SELECT j.id, j.group_id, COALESCE(agc.active_cnt,0) active_cnt FROM repro.job j
  LEFT JOIN LATERAL (SELECT COUNT(*)::int active_cnt FROM (
    SELECT 1 FROM repro.job p WHERE p.name='Q' AND p.state='active'
      AND p.group_id IS NOT NULL AND p.group_id=j.group_id LIMIT 2) x) agc
    ON j.group_id IS NOT NULL
  WHERE j.name='Q' AND j.state<'active' AND NOT j.blocked AND j.start_after<now()
    AND (j.group_id IS NULL OR agc.active_cnt IS NULL OR agc.active_cnt < 2)
  ORDER BY j.created_on, j.id LIMIT 1 FOR UPDATE OF j SKIP LOCKED)
SELECT * FROM next;

\echo '--- B) correlated saturated_groups NOT EXISTS: relapses, aggregate loops=12k ---'
EXPLAIN (ANALYZE, BUFFERS)
WITH saturated_groups AS (
  SELECT group_id FROM repro.job WHERE name='Q' AND state='active' AND group_id IS NOT NULL
  GROUP BY group_id HAVING COUNT(*) >= 2),
next AS (
  SELECT j.id, j.group_id FROM repro.job j
  WHERE j.name='Q' AND j.state<'active' AND NOT j.blocked AND j.start_after<now()
    AND (j.group_id IS NULL OR NOT EXISTS (SELECT 1 FROM saturated_groups s WHERE s.group_id=j.group_id))
  ORDER BY j.created_on, j.id LIMIT 1 FOR UPDATE OF j SKIP LOCKED)
SELECT * FROM next;

\echo '--- C) ARRAY() InitPlan (conc=2): ~3.8ms, HashAggregate loops=1, uncorrelated ---'
EXPLAIN (ANALYZE, BUFFERS)
WITH next AS (
  SELECT j.id, j.group_id FROM repro.job j
  WHERE j.name='Q' AND j.state<'active' AND NOT j.blocked AND j.start_after<now()
    AND (j.group_id IS NULL OR j.group_id <> ALL (ARRAY(
      SELECT group_id FROM repro.job
      WHERE name='Q' AND state='active' AND group_id IS NOT NULL
      GROUP BY group_id HAVING COUNT(*) >= 2)))
  ORDER BY j.created_on, j.id LIMIT 1 FOR UPDATE OF j SKIP LOCKED)
SELECT * FROM next;

DROP SCHEMA repro CASCADE;
```

### Full two-phase shape (batch `limit>1` + tiers correct)

```sql
WITH
active_group_counts AS (            -- small: bounded by active jobs, not backlog
  SELECT group_id, COUNT(*)::int active_cnt FROM job
  WHERE name = $q AND state = 'active' AND group_id IS NOT NULL GROUP BY group_id),
next AS (
  SELECT j.id, j.group_id, j.group_tier FROM job j
  WHERE ...                          -- hot path: uncorrelated InitPlan, computed once
    AND (j.group_id IS NULL OR j.group_id <> ALL (ARRAY(
      SELECT group_id FROM job
      WHERE name = $q AND state = 'active' AND group_id IS NOT NULL
      GROUP BY group_id, group_tier
      HAVING COUNT(*) >= COALESCE(($tiers ->> group_tier)::int, $default))))
  ORDER BY j.created_on, j.id LIMIT $limit FOR UPDATE OF j SKIP LOCKED),
group_ranking AS (                   -- cold path: join on <= $limit rows
  SELECT t.id, t.group_id, t.group_tier,
    ROW_NUMBER() OVER (PARTITION BY t.group_id ORDER BY t.id) group_rn,
    COALESCE(agc.active_cnt, 0) active_cnt
  FROM next t LEFT JOIN active_group_counts agc ON t.group_id = agc.group_id),
group_filtered AS (
  SELECT id FROM group_ranking
  WHERE group_id IS NULL
     OR (active_cnt + group_rn) <= COALESCE(($tiers ->> group_tier)::int, $default))
-- ... UPDATE ... FROM group_filtered
```

---

## Notes / caveats

- **Master baseline:** the current `MATERIALIZED` CTE query is ~7ms when the planner
  hashes the join and ~286ms (`CTE Scan loops=12001`) when it nested-loops. `MATERIALIZED` fixes
  *re-computation*, not *re-scan* — the join plan is what flips. Both the PR's lateral and the `ARRAY()`
  InitPlan remove that plan-choice dependency, so neither can cliff. The real win over master is
  variance, not just the mean.
- On PG18 the fixture naturally picks the good hash plan; reproducing the 286ms cliff needs
  `SET enable_hashjoin=off; SET enable_mergejoin=off;` (step 0 above), or the exact stats the PR's
  regression test constructs. Pathology is stats/version-dependent, but real.
- Memoize does **not** engage for the lateral shape — verified with fresh and stale stats; the
  per-row probe is cheap enough that the planner declines to cache it. The fix's efficiency comes
  from the bounded `LIMIT`ed index probe, not caching.
- `<> ALL(array)` is O(saturated-groups=S) per candidate row (runtime Param → not hashed), vs the
  lateral's O(limit) bounded probe. Measured crossover on ~12k backlog: S=1 → ARRAY 3.8ms / lateral
  33ms; S=500 → ~even (54 / 50ms); S=3000 → ARRAY 285ms / lateral 59ms. S = groups saturated at that
  instant, each holding ≥ limit active (worker-occupied) jobs — so the ~500 crossover is ~500–1,000
  concurrent in-flight jobs on one queue, bounded by live worker throughput, not tenant count. Most
  deployments sit far below it. Practical read: ARRAY ~10× faster for realistic concurrency; lateral
  only wins at extreme scale (~1k+ concurrent jobs) and pays ~10× on the common case for that insurance.
- Two-phase key: saturation filter in the hot path (pre-`LIMIT`) as an uncorrelated `ARRAY()`
  InitPlan; exact-count enforcement in the cold path (post-`LIMIT`) where rows are ≤ `limit`.
  The bug lives entirely in doing the count join pre-`LIMIT`.

bcomnes:
The array is an interesting idea, let me fiddle around with that and see what it looks like. 

There were some other alternatives as like using additional stats tables to shortcut/cache some of this work, but that would require migrations and additional table complexity.

bcomnes:
Thanks for the detailed `ARRAY()` investigation. I implemented that approach first, then continued probing the resulting plans under stale statistics, larger batches, higher active-group cardinality, and tiered concurrency.

That led to the current version of the PR, which uses a JSONB active-count map for `groupConcurrency > 1` and tiered concurrency. The `groupConcurrency: 1` path still uses `NOT EXISTS`, since that already produces a good plan.

## Why move beyond the array

The array approach successfully removes the original pre-`LIMIT` active-count rescan. It converts the set of saturated groups into an uncorrelated InitPlan, so PostgreSQL computes it once instead of pushing a correlated count below the candidate scan.

Further testing found two remaining concerns:

1. `group_id <> ALL(array)` compares each pending candidate with the saturated-group array. Its cost therefore grows with both the number of candidate rows examined and the number of saturated groups.
2. Tiered concurrency requires a different saturation threshold for each tier. The implementation generated one array InitPlan per configured tier, so query size, planning work, and CTE traversal grew with the number of tiers.

There was also a separate planner issue after `LIMIT`: the final `UPDATE ... FROM group_filtered` could inline the ranking query on the inner side of a nested loop. Under the stale-statistics fixture, the array fixed the active-count rescan but could still execute the ranking thousands of times.

The current design addresses both planner-sensitive joins:

- Active counts are grouped once and aggregated into a single JSONB object mapping `group_id` to `active_cnt`.
- Candidate rows use a keyed lookup from that object through an uncorrelated scalar InitPlan.
- The active count is carried through `next` and `group_ranking`; there is no post-`LIMIT` join back to the count relation.
- The final IDs are passed to the update through `id = ANY (ARRAY(SELECT id FROM group_filtered))`, making the filtered ID set a one-time InitPlan as well.
- Tier limits remain data in the existing JSONB parameter, so generated SQL is independent of the number of configured tiers.

## Benchmark setup

I compared three actual implementation snapshots:

- **Original:** the pre-fix materialized count CTE joined before and after `LIMIT`.
- **Array:** the intermediate `ARRAY()` saturation filter with its original final `UPDATE ... FROM group_filtered`.
- **Map:** the current JSONB count map plus the final filtered-ID InitPlan.

The benchmark used real pg-boss schemas and each version's generated `fetchNextJob()` update query on PostgreSQL 17.10. Each execution used `EXPLAIN (ANALYZE, BUFFERS, FORMAT JSON)` and verified the number of jobs claimed.

The incident-shaped fixture used:

- `groupConcurrency: 2`
- One active job followed by `ANALYZE`
- The second active job and remaining fixture rows inserted after `ANALYZE`, leaving planner statistics stale
- 114 additional active groups
- 12,000 older pending jobs belonging to a saturated group
- Eligible jobs behind that saturated backlog

Additional scenarios varied:

- Current versus stale statistics
- Batch sizes of 1, 100, and 1,000
- Saturated active-group counts of 1, 500, 3,000, 10,000, and 50,000
- Tier counts of 1, 100, and 250

The core scenarios were run five times with variant order rotated; the table reports median execution time. The 10,000- and 50,000-group cases were single exploratory runs.

| Scenario | Original | Array | Map |
|---|---:|---:|---:|
| Incident-shaped, stale stats | 158.84 ms | 10.96 ms | **4.03 ms** |
| Incident-shaped, current stats | 3.13 ms | **2.05 ms** | 2.86 ms |
| Batch size 100, stale stats | 554.75 ms | 428.30 ms | **4.77 ms** |
| Batch size 1,000, stale stats | 4.53 s | 4.44 s | **11.11 ms** |
| 500 saturated groups | 637.94 ms | 52.17 ms | **8.36 ms** |
| 3,000 saturated groups | 3.87 s | 248.64 ms | **15.04 ms** |
| 10,000 saturated groups | 13.19 s | 773.48 ms | **34.08 ms** |
| 50,000 saturated groups | 63.79 s | 4.03 s | **147.40 ms** |
| 1 tier | 14.06 ms | 11.69 ms | **3.86 ms** |
| 100 tiers | 140.96 ms | 30.80 ms | **5.99 ms** |
| 250 tiers | 342.59 ms | 55.46 ms | **9.31 ms** |

The current-statistics result is useful context: the array has the lowest floor in the small case. The map adds a small fixed cost to build and access the JSONB value. The benefit is substantially more predictable behavior when statistics are stale or cardinality increases.

## Plan evidence

For the incident-shaped stale-statistics case:

| Implementation | Active-count scan loops | Ranking loops |
|---|---:|---:|
| Original | 12,036 | 12,152 |
| Array | 1 | 12,152 |
| Map/current | 1 | 1 |

For the stale-statistics case with a batch size of 1,000:

| Implementation | Active-count scan loops | Ranking loops |
|---|---:|---:|
| Original | 13,000 | 13,116 |
| Array | 1,000 | 13,116 |
| Map/current | 1 | 1 |

The array InitPlan does its intended job: the pre-`LIMIT` saturation set is built once. The remaining batch cliff comes from the separate final update/ranking plan. The current implementation prevents both forms of repeated execution.

This table compares the three implementation snapshots rather than isolating one variable at a time. In particular, the array snapshot does not include the current final-ID InitPlan. Adding that to an array implementation should remove the ranking cliff, but it would still retain linear `<> ALL(array)` candidate checks and tier-specific array generation. I have not included an unmeasured hybrid result in the table.

At 100 tiers, the generated array query was about 29.6 KB with 101 InitPlans, 100 of which executed. At 250 tiers, it was about 71.3 KB with 251 InitPlans, 250 of which executed. The map query remained about 2 KB with three InitPlans regardless of tier count.

## Representative original query

This is the relevant shape of the original query. The count CTE is materialized once, but PostgreSQL is still free to rescan the materialized result for either join.

```sql
WITH active_group_counts AS MATERIALIZED (
  SELECT group_id, COUNT(*)::int AS active_cnt
  FROM job
  WHERE name = $queue
    AND state = 'active'
    AND group_id IS NOT NULL
  GROUP BY group_id
),
next AS (
  SELECT j.id, j.group_id, j.group_tier
  FROM job j
  LEFT JOIN active_group_counts agc ON agc.group_id = j.group_id
  WHERE j.name = $queue
    AND j.state < 'active'
    AND NOT j.blocked
    AND j.start_after <= now()
    AND (
      j.group_id IS NULL
      OR agc.active_cnt IS NULL
      OR agc.active_cnt < $default_limit
    )
  ORDER BY j.created_on, j.id
  LIMIT $batch_size
  FOR UPDATE OF j SKIP LOCKED
),
group_ranking AS (
  SELECT next.id, next.group_id, next.group_tier,
    ROW_NUMBER() OVER (PARTITION BY next.group_id ORDER BY next.id) AS group_rn,
    COALESCE(agc.active_cnt, 0) AS active_cnt
  FROM next
  LEFT JOIN active_group_counts agc ON agc.group_id = next.group_id
),
group_filtered AS (
  SELECT id
  FROM group_ranking
  WHERE group_id IS NULL
     OR active_cnt + group_rn <= $default_limit
)
UPDATE job j
SET state = 'active', started_on = now()
FROM group_filtered
WHERE j.id = group_filtered.id;
```

## Representative array query

The array version makes the pre-`LIMIT` saturation set an uncorrelated InitPlan. The tiered implementation generated a `CASE` branch and a separate array subquery for each configured tier.

```sql
WITH active_group_counts AS MATERIALIZED (
  SELECT group_id, COUNT(*)::int AS active_cnt
  FROM job
  WHERE name = $queue
    AND state = 'active'
    AND group_id IS NOT NULL
  GROUP BY group_id
),
next AS (
  SELECT j.id, j.group_id, j.group_tier
  FROM job j
  WHERE j.name = $queue
    AND j.state < 'active'
    AND NOT j.blocked
    AND j.start_after <= now()
    AND (
      j.group_id IS NULL
      OR j.group_id <> ALL (ARRAY(
        SELECT group_id
        FROM active_group_counts
        WHERE active_cnt >= $default_limit
      ))
    )
  ORDER BY j.created_on, j.id
  LIMIT $batch_size
  FOR UPDATE OF j SKIP LOCKED
),
group_ranking AS (
  SELECT next.id, next.group_id, next.group_tier,
    ROW_NUMBER() OVER (PARTITION BY next.group_id ORDER BY next.id) AS group_rn,
    COALESCE(agc.active_cnt, 0) AS active_cnt
  FROM next
  LEFT JOIN active_group_counts agc ON agc.group_id = next.group_id
),
group_filtered AS (
  SELECT id
  FROM group_ranking
  WHERE group_id IS NULL
     OR active_cnt + group_rn <= $default_limit
)
UPDATE job j
SET state = 'active', started_on = now()
FROM group_filtered
WHERE j.id = group_filtered.id;
```

## Representative current map query

The map is built once. Both uses of the scalar subquery become uncorrelated InitPlans that read the single materialized JSONB value, rather than joining or scanning an active-group relation once per candidate.

```sql
WITH active_group_count_map AS MATERIALIZED (
  SELECT COALESCE(
    jsonb_object_agg(group_id, active_cnt),
    '{}'::jsonb
  ) AS counts
  FROM (
    SELECT group_id, COUNT(*)::int AS active_cnt
    FROM job
    WHERE name = $queue
      AND state = 'active'
      AND group_id IS NOT NULL
    GROUP BY group_id
  ) active_groups
),
next AS (
  SELECT j.id, j.group_id, j.group_tier,
    COALESCE(
      ((SELECT counts FROM active_group_count_map) ->> j.group_id)::int,
      0
    ) AS active_cnt
  FROM job j
  WHERE j.name = $queue
    AND j.state < 'active'
    AND NOT j.blocked
    AND j.start_after <= now()
    AND (
      j.group_id IS NULL
      OR COALESCE(
        ((SELECT counts FROM active_group_count_map) ->> j.group_id)::int,
        0
      ) < COALESCE(($tiers ->> j.group_tier)::int, $default_limit)
    )
  ORDER BY j.created_on, j.id
  LIMIT $batch_size
  FOR UPDATE OF j SKIP LOCKED
),
group_ranking AS (
  SELECT next.*,
    ROW_NUMBER() OVER (PARTITION BY group_id ORDER BY id) AS group_rn
  FROM next
),
group_filtered AS (
  SELECT id
  FROM group_ranking
  WHERE group_id IS NULL
     OR active_cnt + group_rn
        <= COALESCE(($tiers ->> group_tier)::int, $default_limit)
)
UPDATE job j
SET state = 'active', started_on = now()
WHERE j.id = ANY (ARRAY(SELECT id FROM group_filtered));
```

## Regression coverage

The committed tests intentionally assert deterministic behavior and plan structure rather than elapsed time. They cover:

- Current and stale statistics
- `groupConcurrency: 1` and `groupConcurrency: 2`
- A saturated backlog with eligible work behind it
- A large batch
- Correct claimed-job counts
- No repeated active-count CTE scan
- No repeated ranking execution
- No repeated job-table scan caused by the group-concurrency subplan
- Generated SQL remaining independent of configured tier count

The existing behavioral group-concurrency tests also continue to cover normal and tiered claiming semantics.

## Remaining tradeoff

The JSONB map still has an `O(active groups)` construction cost and allocates one JSONB value per fetch. It is not free at extreme cardinalities: the deliberately large 50,000-group fixture took about 147 ms locally. It also does not remove the cost of scanning past a large pending backlog, and it does not change the existing concurrency semantics between simultaneous workers.

What it does provide is a query shape whose active-count lookup, tier handling, ranking, and final update do not depend on planner-selected joins against a multi-row count relation.

I can commit the complete three-way benchmark harness, standalone SQL reproductions, or both if having runnable examples in the repository would help review. I left them out initially because pg-boss does not currently appear to have a benchmark convention, and elapsed-time thresholds would not be appropriate for normal CI.

bcomnes:
BTW the original version of the PR DID NOT WORK OUT! 😭 Benchmarks on the current version look promising though. I will continue to probe it a bit.

bcomnes:
Im capturing the benchmark work over in this branch: https://github.com/bcomnes/pg-boss/tree/bret/group-concurrency-benchmarks

timgit:
This looks good.  I'll wait for you to issue the all clear before I merge.

bcomnes:
My biggest concern with this is if this will make the situation where you have empty workers, a large pool of jobs for a group id beyond the group concurrency limit and they all over fetch jobs beyond the concurrency limit at once.

bcomnes:
Overfetch is the same.

bcomnes:
I haven't had a chance to throw prod traffic at this yet, but the benchmarking looks good. Hard for me to predict though.

PR #846, opened 2026-07-12
Group availability hooks

bcomnes:
Group concurrency is great for limiting per Tennant capacity. However if the jobs consume some kind of externally rate limited resource then you have to centrally coordinate that across all workers (even if you have enough per-tenant capacity on your worker side). When the resources needed inside of a job are exhausted, it would be really nice to have a queues side hook to disable jobs that will run into that condition, until they become runnable again.

Hence: Group availability hooks

This is a hook that runs against candidate job group ids, and can further narrow the group id selection of the final job candidate selection. 

This way an external resource can signal to the queue "Don't work on these jobs! they won't work yet".

Still just exploring the idea, feedback and criticism welcome. I'm still iterating on the implementation, code review not necessary yet.


bcomnes:
I'm going to close this until I convince myself its a good idea.

PR #848, opened 2026-07-16
Fix PGlite flakiness from coarse now() timestamps

ziemekobel:
### Problem

`npm run test:pglite` failed, because on PGlite, jobs were often not fetchable right after `send()`, even though the same tests passed against real Postgres.

PGlite’s now() is coarse: back-to-back statements frequently see the same timestamp. Jobs are inserted with `start_after = now()`, but fetch required `start_after < now()`. When those times tied, the job looked “not ready yet,” so `fetch` returned nothing (or only part of a batch). That showed up as flaky failures in `npm run test:pglite`.

A related issue: when several jobs shared the same `created_on`, ordering fell back to UUID `id`, so FIFO / insertion-order assertions could pick the wrong job.

### Solution

Fetch with `start_after <= now()` (same rule NOTIFY already uses), so a job scheduled for “now” is immediately claimable.
In PGlite tests only, nudge the clock after each `send/insert` so `created_on` values stay distinct and order-sensitive tests stay stable.

### DIFF
--- a/docs/database-backends.md
+++ b/docs/database-backends.md
@@ -102,7 +102,7 @@ CockroachDB but are independent in principle:
     SELECT id FROM jobs
     WHERE name = $name
       AND state < 'active'
-      AND start_after < now()
+      AND start_after <= now()
     ORDER BY priority DESC, created_on, id
     LIMIT $batchSize
   )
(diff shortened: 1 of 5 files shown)


timgit:
Thanks!

PR #849, opened 2026-07-16
Fix key_strict_fifo blocking and priority ordering

ziemekobel:
### Problem

`key_strict_fifo` previously enforced ordering only through the job_i8 unique index. On fetch, pg-boss would try to activate a blocked key's queued job, hit a unique violation (23505), and `manager.fetch()` swallowed it as an empty result. For batch fetches this meant **a single stuck, retrying, or permanently failed key starved every other key in the queue.**

### Solution

`fetchNextJob` now excludes candidates whose `singleton_key` already has a job in `active/retry/failed` and dedups to the oldest queued job per key, ordered by `(created_on, id)`. A blocked key only holds back its own successors; every other key keeps flowing, and successors resume in FIFO order once the blocking job leaves those states. The `NOT EXISTS` carries the policy predicate so the partial job_i8 index services the per-key lookup instead of a sequential scan.

Additionally, per-key ordering was still violated by priority: the per-key dedup only ran for batch fetches and, in both paths, happened after the global `priority desc, created_on, id` ordering and LIMIT, so a newer higher-priority job of a key could be fetched ahead of an older one. `fetchNextJob` now restricts candidates to each key's oldest eligible job (its FIFO head) before priority and LIMIT apply, so within a key the oldest always wins while priority still orders across keys, upholding the "processed in the order they were created" contract.

Adds tests for:
* a failed key not blocking other keys,
* an active key not blocking other keys,
* successors recovering in order after retry,
* batch fetch deduping same-key jobs,
* priority not reordering within a key,
* and batch fetch not surfacing a newer high-priority same-key job.

Docs updated to describe per-key (not per-queue) blocking.

### DIFF
--- a/docs/api/queues.md
+++ b/docs/api/queues.md
@@ -40,18 +40,20 @@ Allowed policy values:
 | `singleton` | Only allows 1 job to be active, unlimited queued. Can be extended with `singletonKey` |
 | `stately` | Combination of short and singleton: Only allows 1 job per state, queued and/or active. Can be extended with `singletonKey` |
 | `exclusive` | Only allows 1 job to be queued or active. Can be extended with `singletonKey` |
-| `key_strict_fifo` | Strict FIFO ordering per `singletonKey`. Requires `singletonKey` on every job. Blocks processing of jobs with the same key while any job with that key is active, in retry, or failed. |
+| `key_strict_fifo` | Strict FIFO ordering per `singletonKey`. Requires `singletonKey` on every job. Holds back the next job for a key while a job with that same key is active, in retry, or failed. Blocking is scoped per key, so a stuck key never stalls other keys. |
 
 > [!WARNING]
 > `stately` queues are special in how retries are handled. By definition, stately queues will not allow multiple jobs to occupy `retry` state. Once a job exists in `retry`, failing another `active` job will bypass the retry mechanism and force the job to `failed`. If this job requires retries, consider a custom retry implementation using a dead letter queue.
 
 > [!NOTE]
-> `key_strict_fifo` queues enforce strict FIFO (First-In-First-Out) ordering per `singletonKey`. This is useful when you need to ensure jobs for the same entity (e.g., the same order, customer, or resource) are processed sequentially in the order they were created. The queue will block processing of subsequent jobs with the same `singletonKey` while any job with that key is:
+> `key_strict_fifo` queues enforce strict FIFO (First-In-First-Out) ordering per `singletonKey`. This is useful when you need to ensure jobs for the same entity (e.g., the same order, customer, or resource) are processed sequentially in the order they were created. The queue will hold back subsequent jobs with the same `singletonKey` while any job with that key is:
 > - **active**: currently being processed
 > - **retry**: waiting to be retried after a failure
 > - **failed**: permanently failed (exhausted all retries)
 >
-> To unblock a key after a permanent failure, you can either delete the failed job using `deleteJob()` or retry it using `retry()`. Use `getBlockedKeys()` to discover which keys are currently blocked due to failed jobs.
+> Blocking is scoped to the individual key. A key that is stuck (for example on a permanently failed job) only holds back its own successors; jobs for every other `singletonKey` continue to be fetched normally, so a single failure never stalls the whole queue.
+>
+> To unblock a key after a permanent failure, you can either delete the failed job using `deleteJob()` or retry it using `retry()`. Once the blocking job leaves the active/retry/failed state, that key's queued successors become fetchable again, still in FIFO order. Use `getBlockedKeys()` to discover which keys are currently blocked due to failed jobs.
 
 * **partition**, boolean, default false
 
--- a/src/plans.ts
+++ b/src/plans.ts
@@ -1337,7 +1337,8 @@ function buildFetchParams (options: FetchJobOptions): FetchQueryParams {
 export function fetchNextJob (options: FetchJobOptions, noSkipLocked = false): SqlQuery {
   const { schema, table, name, policy, limit, includeMetadata, priority = true, orderByCreatedOn = true, ignoreStartAfter = false, groupConcurrency, minPriority, maxPriority } = options
 
-  const singletonFetch = limit > 1 && (policy === QUEUE_POLICIES.singleton || policy === QUEUE_POLICIES.stately)
+  const keyStrictFifo = policy === QUEUE_POLICIES.key_strict_fifo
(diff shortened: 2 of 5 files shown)


ziemekobel:
`npm run cover` works fine for me, so dunno why it failed in CI

ziemekobel:
Hey @timgit! Any chance to get it reviewed soon?

timgit:
Thanks for opening and the fix.  I ran this through a review with Claude to look for any indexing-related side effects, and it flagged a couple of things.  I'm including its write up here for your review.  

---

> **Blocker: the fetch cost is quadratic in per-key depth.** The FIFO-head `NOT EXISTS` (the `h`
> subquery) has no supporting index, so the planner compiles it to a hash anti-join whose join
> filter compares every same-key pair — a key with depth *d* costs *d²* comparisons. Deep single
> keys (one hot entity with a backlog) are exactly the `key_strict_fifo` workload, and it degrades
> right when the queue is already under load. Measured at 200k queued rows, PG18:
>
> | shape | master | this PR | index + `DISTINCT ON` |
> |---|---|---|---|
> | 8,000 keys × 25/key | 142 ms | 422 ms (3.0×) | **137 ms** |
> | 2,000 keys × 100/key | 140 ms | 1,069 ms (7.6×) | **110 ms** |
> | 200 keys × 1,000/key | 127 ms | **7,930 ms (62×)** | **59 ms** |
>
> **The fix needs both a new index and a query reshape** — I measured each half alone and neither is
> enough (index alone can't be consumed by the anti-join; `DISTINCT ON` alone trades the hash for a
> full sort and is *worse* than this PR). Together they beat master at every shape and get *faster*
> as per-key depth grows, because the priority sort input shrinks to one head per key.
>
> Partial index, built in the same `createQueue` policy branch that already emits `job_i8`, so only
> `key_strict_fifo` queues pay for it (needs a schema bump + migration — next is v38):
>
> ```sql
> CREATE INDEX job_i10 ON job (name, singleton_key, created_on, id)
>   WHERE state < 'active' AND NOT blocked AND policy = 'key_strict_fifo'
> ```
>
> Then replace the `h` `NOT EXISTS` with a `heads` CTE that picks one head per key. The index
> supplies `(singleton_key, created_on, id)` order, so `DISTINCT ON` needs no sort:
>
> ```sql
> WITH heads AS (
>   SELECT DISTINCT ON (singleton_key) id
>   FROM job
>   WHERE name = $1 AND state < 'active' AND NOT blocked AND start_after < now()
>     AND policy = 'key_strict_fifo'
>   ORDER BY singleton_key, created_on, id
> ),
> next AS (
>   SELECT j.id FROM job j
>   WHERE j.name = $1 AND j.id IN (SELECT id FROM heads)
>     AND <existing blocking NOT EXISTS, unchanged>
>     AND <ignoreSingletons / minPriority / maxPriority / group filters, unchanged>
>   ORDER BY j.priority desc, j.created_on, j.id
>   LIMIT $2
>   FOR UPDATE OF j SKIP LOCKED
> )
> ```
>
> `FOR UPDATE` stays on `next` so locking is unchanged — the two-CTE split is because `DISTINCT ON`
> is illegal with `FOR UPDATE`. Keep the `ignoreStartAfter` variant of the `heads` predicate in sync
> with `j`. This also makes `singleton_ranking` redundant for `key_strict_fifo` (see below), so you
> can drop `keyStrictFifo` from the `singletonFetch` condition.
>
> **Two correctness details:**
>
> 1. **`startAfter` reorders within a key — the same class of bug this PR fixes for priority.** The
>    head-check scopes "older sibling" to currently-eligible rows (`start_after < now()`), so a
>    delayed older job isn't treated as the head and a newer same-key job gets fetched ahead of it:
>    ```js
>    await boss.send(q, { o: 1 }, { singletonKey: 'A', startAfter: 60 })  // older
>    const newer = await boss.send(q, { o: 2 }, { singletonKey: 'A' })    // newer
>    const [job] = await boss.fetch(q)   // job.id === newer -> FIFO violated (confirmed)
>    ```
>    This is pre-existing on master and your inline comment calls it out, but the PR also strengthens
>    the docs to promise jobs run "in the order they were created" and resume "still in FIFO order" —
>    that claim and this behavior can't both hold. My preference is to document the `startAfter`
>    carve-out rather than make delays block a key (that's a bigger behavior change than this PR
>    should carry), but the stronger doc language needs to come back down either way.
>
> 2. **`minPriority`/`maxPriority` workers can stall a key permanently.** The head-check doesn't
>    apply `j`'s priority filters, so if a key's head is below a worker's `minPriority`, that worker
>    gets nothing for the key:
>    ```js
>    await boss.send(q, { o: 1 }, { singletonKey: 'A', priority: 0 })   // head
>    await boss.send(q, { o: 2 }, { singletonKey: 'A', priority: 10 })
>    await boss.fetch(q, { batchSize: 5, minPriority: 5 })  // master: returns the priority-10 job
>                                                            // this PR: returns 0 (measured)
>    ```
>    Arguably correct (skipping the head would break FIFO), but it's a real change from master and a
>    fleet where every worker sets `minPriority` never drains that key. Same story for `ignoreGroups`
>    and a saturated `groupConcurrency` group. A test pinning the intent plus a doc line would cover
>    it.
>
> **Minor:** after the reshape, `singleton_ranking` is dead weight for `key_strict_fifo` (the
> head-check already guarantees ≤1 candidate per key). Also worth noting in the description that the
> `row_number() ... ORDER BY id` change quietly makes batch fetch on `singleton`/`stately`
> deterministic (keeps lowest id) — a nice improvement, just outside the stated scope.

ken8203:
Hi @ziemekobel — thank you for the original fix, analysis, and test cases. I also need this fix downstream, so I picked up the maintainer-requested `DISTINCT ON` + `job_i10` follow-up in draft #871, rebased on current master. I've credited your work in the PR description. If you'd prefer to complete the work in #849 yourself, I'm happy to close #871 and contribute here instead.

ziemekobel:
> Hi @ziemekobel — thank you for the original fix, analysis, and test cases. I also need this fix downstream, so I picked up the maintainer-requested `DISTINCT ON` + `job_i10` follow-up in draft #871, rebased on current master. I've credited your work in the PR description. If you'd prefer to complete the work in #849 yourself, I'm happy to close #871 and contribute here instead.

Hi @ken8203, thank you for fixing it for both of us :) Let's continue working on it in your PR

ziemekobel:
https://github.com/timgit/pg-boss/pull/871 addresses issues spotted in this PR, so the further work will continue there.

PR #859, opened 2026-07-29
allow schema names that require quoting

### LINKED ISSUE #700: Schema name validation incorrectly rejects dashes (hyphens) which are valid in PostgreSQL
`pg-boss` currently validates schema names to only allow alphanumeric characters and underscores, but PostgreSQL actually supports dashes/hyphens in schema names when properly quoted (e.g., `"my-schema"`)

**Error:**
```
AssertionError [ERR_ASSERTION]: Name can only contain alphanumeric characters or underscores
    at assertPostgresObjectName (file:///pg-boss/dist/attorney.js:163:5)
```

**Steps to reproduce:**

```
const boss = new PgBoss({
  connectionString: 'postgres://...',
  schema: 'my-schema'  // Valid PostgreSQL schema name
});
```

**Expected behavior:**
Schema names with dashes should be accepted and properly quoted in SQL queries

**Actual behavior:**
Throws assertion error rejecting valid PostgreSQL identifiers

**Suggested solution:**

1. Allow dashes in schema names and ensure proper quoting in generated SQL, or
2. Document this limitation clearly in the configuration docs

PostgreSQL allows any character in identifiers when quoted, so the current validation is overly restrictive

MoazEmad1:
Closes #700.

Allows any name PostgreSQL accepts as a quoted identifier to be used as `schema`, by quoting the identifier at the point of use rather than restricting the input.

### Approach

`assertPostgresObjectName` now rejects only what PostgreSQL genuinely cannot store: empty names, names over 50 characters, and null bytes. The schema is then rendered according to the position it appears in, which turned out to need three distinct treatments rather than one:

| Position | Rendering | Reason |
| --- | --- | --- |
| identifier, e.g. `${schema}.job` | `quoteIdent` | quoted only where required |
| `nspname = '...'`, `quote_ident('...')`, `sha224('...')` | `escapeLiteral` | the catalog stores the resolved name, and `quote_ident` is applied by postgres at runtime |
| `to_regclass('....version')` | `literalIdent` | parsed back into an identifier path, so it must arrive pre-quoted |

The last one is worth calling out. `to_regclass` splits on `.` and down cases unquoted parts rather than fully lexing them, so `to_regclass('My-Schema.version')` returns **null** instead of raising. Passing the raw name there would have made version detection silently report an empty schema on a populated database.

### Backward compatibility

A name is emitted bare exactly when the previous validation would have accepted it (`^[A-Za-z_][A-Za-z0-9_]*$`), reserved words excepted. This is deliberate rather than cosmetic: an installation configured with `MySchema` physically owns a schema named `myschema`, because the bare name was folded on the way in. Quoting it now would point pg-boss at a different, empty schema. Restricting quoting to names that were previously rejected keeps every existing deployment on exactly the SQL it runs today, and the schema manifest check confirms the generated SQL is unchanged.

Reserved words are the one exception. They pass the current validation but generate DDL that fails to parse, so no deployment can be using one today:

```js
new PgBoss({ schema: 'user' }) // accepted
// CREATE SCHEMA IF NOT EXISTS user;
// ERROR: syntax error at or near "user"
```

Those are now quoted, which resolves that case as a side effect. Happy to split it out if you'd rather keep this PR to the reported issue.

### Tests

`test/schemaIdentifierTest.ts` covers name validation, the bare vs quoted rules including the mixed case compatibility case, literal rendering, the generated SQL for each of the three positions, and a full send/fetch/complete lifecycle in a schema named `pg-boss test.schema`.

`test/testHelper.ts` interpolated schema and table names unquoted in `dropSchema`, `findJobs`, and `countJobs`, so those are quoted as well.

Docs updated for the `schema` option.

### DIFF
--- a/docs/api/constructor.md
+++ b/docs/api/constructor.md
@@ -72,6 +72,14 @@ The following options can be set as properties in an object for additional confi
 
     Database schema that contains all required storage objects. Only alphanumeric and underscore allowed, length: <= 50 characters
 
+    To use a name that isn't a legal bare identifier — one containing dashes, or a reserved word — quote it yourself:
+
+    ```js
+    new PgBoss({ schema: '"My-Schema"' })
+    ```
+
+    The value is used verbatim as an identifier, so the quotes are preserved as written. Note that `MySchema` and `"MySchema"` are different schemas: PostgreSQL folds the unquoted form to `myschema`. Double quotes, single quotes, periods, dollar signs, backslashes and control characters are rejected inside a quoted name.
+
 
 **Operations options**
 
--- a/src/attorney.ts
+++ b/src/attorney.ts
@@ -514,9 +514,37 @@ function resolveBackend (config: any) {
 
 function assertPostgresObjectName (name: string) {
   assert(typeof name === 'string', 'Name must be a string')
-  assert(name.length <= 50, 'Name cannot exceed 50 characters')
-  assert(!/\W/.test(name), 'Name can only contain alphanumeric characters or underscores')
-  assert(!/^\d/.test(name), 'Name cannot start with a number')
+
+  const quoted = name.startsWith('"') && name.endsWith('"') && name.length > 1
+  const resolved = quoted ? name.slice(1, -1) : name
+
+  assert(resolved.length > 0, 'Name cannot be empty')
+  // the limit applies to the resolved name, since that is what postgres stores and what the
+  // derived object names are built from.
+  assert(resolved.length <= 50, 'Name cannot exceed 50 characters')
+
+  if (quoted) {
+    // the value is interpolated into identifier positions verbatim, so a double quote inside the
+    // outer pair would close the identifier early and allow arbitrary SQL to follow.
+    assert(!resolved.includes('"'), 'Quoted name cannot contain double quotes')
+  } else {
+    assert(!/\W/.test(name), 'Name can only contain alphanumeric characters or underscores')
+    assert(!/^\d/.test(name), 'Name cannot start with a number')
+    return
+  }
+
+  // a single quote would terminate the literal in the string-comparison positions, and would also
+  // break the `EXECUTE format('… ')` bodies the schema is interpolated into.
+  assert(!resolved.includes("'"), 'Quoted name cannot contain single quotes')
+  // $ can collide with the dollar-quoted function bodies the schema appears inside.
+  assert(!resolved.includes('$'), 'Quoted name cannot contain dollar signs')
+  // partition and table naming splits on '.', so a dot would be misread as a schema separator.
+  assert(!resolved.includes('.'), 'Quoted name cannot contain periods')
+  // backslash is not valid in the bytea inputs the name is hashed through, and control
+  // characters can break out of the provenance comments in exported migration SQL.
+  assert(!resolved.includes('\\'), 'Quoted name cannot contain backslashes')
+  // eslint-disable-next-line no-control-regex
+  assert(!/[\x00-\x1f\x7f]/.test(resolved), 'Quoted name cannot contain control characters')
 }
 
 function assertQueueName (name: string) {
--- a/src/drifter.ts
+++ b/src/drifter.ts
@@ -1,3 +1,4 @@
+import { resolveSchemaName } from './tools.ts'
 import type { ManagedIndex, InvalidIndex, MismatchedIndex, ManagedFunction, MismatchedFunction, TableColumnDrift, EnumDrift, ConstraintDrift, SchemaDriftReport } from './types.ts'
 
 const SINGLE_QUOTE_REGEX = /'/g
@@ -166,7 +167,7 @@ export function getSchemaIndexes (schema: string) {
     JOIN pg_class c ON c.oid = i.indexrelid
     JOIN pg_class t ON t.oid = i.indrelid
     JOIN pg_namespace n ON n.oid = c.relnamespace
-    WHERE n.nspname = '${schema.replace(SINGLE_QUOTE_REGEX, "''")}'
+    WHERE n.nspname = '${resolveSchemaName(schema).replace(SINGLE_QUOTE_REGEX, "''")}'
   `
 }
 
@@ -179,7 +180,7 @@ export function getSchemaFunctions (schema: string) {
     SELECT p.proname AS name, pg_get_functiondef(p.oid) AS def
     FROM pg_proc p
     JOIN pg_namespace n ON n.oid = p.pronamespace
-    WHERE n.nspname = '${schema.replace(SINGLE_QUOTE_REGEX, "''")}'
+    WHERE n.nspname = '${resolveSchemaName(schema).replace(SINGLE_QUOTE_REGEX, "''")}'
   `
 }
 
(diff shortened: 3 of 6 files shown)


timgit:
Thanks again for opening. The `to_regclass` catch is a good find.  Also, this helped me think more about the back-compat use case, which makes this a bit more complicated than I had hoped. I did a claude review, which found some blockers, but I think it may be worth switching the approach a bit, which I have attached at the bottom.

## Blockers

Validation is now "anything but NUL", but the generator can't carry that much.
Running `plans.create(schema, 37, {createSchema: true})` then `create_queue`:

| name | result |
|---|---|
| `my-schema`, `My.Schema`, `user`, `1schema`, `schéma`, `a b`, `a"b` | ok |
| `o'brien` | `syntax error at or near "brien"` |
| `a$$b` | `syntax error at or near "b"` |
| `x.job` | `schema "x.job_common" does not exist` |
| `a\b` | `invalid input syntax for type bytea` |

`quoteIdent`'s output lands inside single-quoted literals in a lot of places (every `EXECUTE format('… ${quoteIdent(schema)} …')`, plus `advisoryLock`, which should be using `escapeLiteral`  compare `notifyChannelSql`). `$$` and `$cmd$` break the dollar-quoted function bodies. Dots are newly legal, so a name containing `.job` gets mangled by `job_table_format`'s rewrite.

One more, worth calling out separately: a newline in the name escapes the provenance comment in exported migration SQL.  (Including for completeness, but would anyone intentionally do this, right?)

```sql
-- inlined from "a
DROP SCHEMA public CASCADE;--".job_table_run_async (migration v28, …)
```

That's the `pg-boss migrate --output` path, which someone pipes into psql.

## Alternative proposal

#700's original request specifically mentioned wanting to use dashes, but the blockers above are related to parsing all potential inputs, which touches a lot of the code in `plans.ts` and `migrationStore.ts`, the latter which would ideally be left alone.

Instead, what if we simply allowed the caller to specify the quoting, such as this?

```js
new PgBoss({ schema: '"My-Schema"' })
```

Identifier positions already work with that on master today, unmodified: create, create_queue, getVersion, correct `nspname`. So `plans.ts` and `migrationStore.ts` need no changes at all. Only the literal positions break (10 sites: `nspname =` in plans/drifter, plus one `quote_ident`), and those need one helper resolving the config string to the name the catalog actually holds.

Two things that need care if you want to continue working on this PR:

- Validation is security-critical here, not optional. `schema: '"a"; DROP SCHEMA public CASCADE; --"'` injects on master today. Rejecting `"` inside the outer pair makes it safe. `$`, `.job` and control chars still need rejecting for the reasons above.  Caller-quoting removes the quoting logic, not the hazards.
- The resolver has to strip quotes only when they're redundant (`"pgboss"` -> `pgboss`), not unconditionally. Otherwise `MySchema` and `"MySchema"` end up sharing a notify channel and advisory lock.

Separately, I found a pre-existing bug while going through this: `nspname = '${schema}'` compares the configured name against the resolved one, so mixed-case installs are already broken (`ensureQueueStatsPartitions('MySchema')`
throws every run, and old stats partitions never get dropped). Not yours, so this could be addressed in this PR or it can be a separate issue. The resolver above happens to fix it.

Let me know your thoughts, or if you'd rather I pick it up from here. Either way the effort here wasn't wasted. Thanks again.

MoazEmad1:
Thanks for the thorough review. I reproduced all four failures locally and they're exactly as you describe.

The root cause is a position I didn't account for. I classified the schema's uses as identifier, literal, and identifier-inside-a-literal, but missed a fourth: `quoteIdent` output landing inside a single quoted string that postgres re-parses later, which is every `EXECUTE format('… ${quoteIdent(schema)} …')` body. That's ~197 sites, and it's why `o'brien` terminates the enclosing literal. `advisoryLock` is a plain mistake on my part; the schema is hashed there exactly as in the notify channel, so it should have been `escapeLiteral` and I converted one and not the other.

I'd like to continue with the caller-quoting approach if you're happy for me to. I checked the premise and it holds: with `schema: '"My-Schema"'` the identifier positions and `to_regclass` both resolve correctly against unmodified generation, and only the catalog comparison breaks. `nspname = '"My-Schema"'` matches nothing where `nspname = 'My-Schema'` matches. Confining the change to the resolver and the ~10 literal sites is a much better trade than rewriting `migrationStore.ts`, and it drops the mixed-case folding problem entirely since the caller is now stating which one they mean.

On the two points you flagged:

- Validation runs before anything else, on the raw config string. A `"` inside the outer pair is rejected, which closes the injection; `$`, control characters including newline, and `.` are rejected as well, for the dollar-quoting, provenance-comment and `job_table_format` reasons you gave. I'll treat the length limit as applying to the resolved name rather than the quoted string.
- The resolver strips the outer quotes only when they're redundant, so `"pgboss"` resolves to `pgboss` and shares its channel and lock, while `MySchema` and `"MySchema"` stay distinct.

Happy to fold in the `nspname` mixed-case bug here since the resolver fixes it as a side effect. Say the word if you'd rather it went in its own issue and I'll leave it out.

I'll rework the branch and force-push. Since this replaces the current diff wholesale, let me know if you'd prefer a fresh PR instead.

MoazEmad1:
Reworked to the caller-quoting approach and force-pushed, so the diff is entirely different from before. `plans.ts` and `migrationStore.ts` are back to master apart from the literal sites; `migrationStore.ts`, `cli.ts` and `testHelper.ts` are untouched. 370 lines instead of 2770.

**Validation** (`attorney.ts`) runs on the raw config string. Bare names keep exactly the old rules. Quoted names reject `"` inside the outer pair, plus `'`, `$`, `.`, `\` and control characters. Your four cases and the injection string are all rejected up front:

```
"a"; DROP SCHEMA public CASCADE; --"  -> Quoted name cannot contain double quotes
"o'brien"                            -> Quoted name cannot contain single quotes
"a$$b"                               -> Quoted name cannot contain dollar signs
"x.job"                              -> Quoted name cannot contain periods
"a\b"                                -> Quoted name cannot contain backslashes
```

The length limit now applies to the resolved name rather than the quoted string.

**Resolver** (`tools.ts`): one note where I departed from your description. You suggested stripping quotes only when redundant, but a quoted name is always stored verbatim, so the redundancy test doesn't change the resolved value, only whether two configs collide. The rule reduces to quoted → contents verbatim, bare → lower case, which still gives both properties you wanted:

```
resolveSchemaName('"pgboss"')   === resolveSchemaName('pgboss')     //true, shared channel and lock
resolveSchemaName('"MySchema"') !== resolveSchemaName('MySchema')   // distinct, the latter is myschema
```

Happy to switch to an explicit redundancy check if you'd rather it read that way.

**Literal sites**: 12: six `nspname` in `drifter.ts`, three in `plans.ts`, one `quote_ident`, plus `advisoryLock` and the notify channel hash. `advisoryLock` was the one I'd got wrong last time; it hashes the name exactly as the notify channel does, so it takes the resolved name too. The `nspname` inside the v28 migration body is left alone since that output is frozen.

**The pre-existing mixed-case bug** is fixed as a side effect: `ensureQueueStatsPartitions('MySchema')` succeeds now where it threw on every run before. There's a regression test for it. Say the word if you'd rather that came out and went in its own issue.

One caveat worth stating plainly: I've verified this against postgres locally, but the validation here is security-critical and I'd rather you looked hard at the reject list than took my word that it's complete.

PR #860, opened 2026-07-31
add exact dead-letter redrive

mikery:
## What changed

- add `redriveJob(name, deadLetterJobId, options)` for one exact dead-letter row
- return the fresh replacement job ID, or `null` when the selected row is not eligible
- accept the existing `db` option so the redrive composes inside a caller-owned transaction
- preserve the source row when the replacement insert cannot succeed

## Why

The existing batch `redrive()` API selects jobs oldest-first through `limit`; it cannot safely target an operator-selected dead-letter job. Applications that require audited, single-job recovery need a native exact-ID primitive without reaching into pg-boss tables themselves.

The `id` argument is explicitly the selected row's ID in the dead-letter queue, not its earlier `sourceId` metadata.

## Implementation

The PostgreSQL plan locks only the selected eligible row, creates a fresh replacement using the destination queue's current policy, deletes the dead-letter row only after insertion succeeds, and returns the replacement UUID. Passing a custom `db` keeps the operation within that adapter's existing transaction.

## Validation

- `npm run test:pglite -- test/failureTest.ts` — 19 passed, 1 skipped
- `npm test -- test/adapterTest.ts -t "should roll back an exact redrive through a knex transaction"` — passed against PostgreSQL
- TypeScript, ESLint, and schema-manifest checks pass as part of those commands

### DIFF
--- a/docs/api/jobs.md
+++ b/docs/api/jobs.md
@@ -529,6 +529,25 @@ do {
 } while (moved > 0)
 ```
 
+### `redriveJob(name, id, options)`
+
+Moves one exact, un-started job out of a dead letter queue and re-creates it as a
+fresh job on its original source queue. Returns the replacement job id, or `null`
+when the selected dead-letter job is absent, currently being processed, has no
+recorded source queue, or cannot be inserted under that queue's current policy.
+
+The source job is deleted only after its replacement is inserted. Pass a custom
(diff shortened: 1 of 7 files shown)


mikery:
Opened on wrong upstream by mistake. Apologies for spam.

PR #871, opened 2026-08-11
fix: prevent key_strict_fifo jobs from blocking other keys

ken8203:
This is a follow-up to #849 by @ziemekobel, whose original report, implementation, and test cases established the problem. I also need this fix downstream, so I picked up the maintainer-requested query/index reshape and schema migration. If @ziemekobel would prefer to finish the work in #849, I'm happy to close this PR and contribute there instead.

## Problem

`key_strict_fifo` can return no work for the entire queue when the oldest candidate belongs to a key that already has an `active`, `retry`, or `failed` job. A single blocked key therefore starves unrelated keys.

This builds on #849 and implements the indexed `DISTINCT ON` shape suggested in the maintainer review there, including the requested schema v38 migration.

## Changes

- Select one eligible FIFO head per `singleton_key` in a materialized CTE before applying cross-key priority and fetch limits.
- Keep the blocking probe scoped to the same key, backed by the existing unique `job_i8` partial index.
- Add `job_i10 (name, singleton_key, created_on, id)` for strict FIFO head selection.
- Build `job_i10` on `job_common` and existing/future `key_strict_fifo` partitions only.
- Make CLI/exported migrations carry partition policy metadata so their target set matches the live BAM path.
- Make v38 rollback remove pending BAM commands and refuse rollback while the index is actively being built.
- Document and test the intentional semantics:
  - `startAfter` jobs join FIFO ordering when they become eligible.
  - priority orders across keys but cannot reorder within a key.
  - `minPriority`/`maxPriority` cannot skip a key's head.

## Correctness coverage

Tests cover both shared and partitioned queues:

- `active`, `retry`, and `failed` jobs block only their own key.
- batch fetch returns at most one head per key.
- priority does not reorder jobs within a key.
- deferred jobs and priority filters follow the documented semantics.
- v37 → v38 migration targets only shared and strict FIFO job tables.
- rollback removes pending builds and rejects an in-progress build.
- live BAM and CLI/export migration paths produce the same target scope.

## Performance

PostgreSQL 18, 200,000 queued rows:

| Shape | Execution time |
| --- | ---: |
| 8,000 keys × 25 jobs | 64.7 ms |
| 2,000 keys × 100 jobs | 44.8 ms |
| 200 keys × 1,000 jobs | 47.7 ms |
| 200,000 keys × 1 job | 381.6 ms |

The depth cases no longer exhibit the O(d²) correlated anti-join behavior reported on #849. The 200,000-distinct-key case is the opposite breadth extreme: it materializes and externally sorts 200,000 heads, with about 12 MB temp reads and 43 MB temp writes in this run. It is included here as an explicit scaling ceiling rather than hidden by the depth-focused benchmark.

`INCLUDE (start_after)` was also tested against 199,999 deferred rows. PostgreSQL selected the existing `job_i5` in both cases (0.169 ms without INCLUDE, 0.116 ms with it), so this PR keeps `job_i10` narrow to avoid write amplification.

## Verification

- PostgreSQL: 193 affected tests passed
- PGlite: 180 affected tests passed
- Distributed strict FIFO: 39 tests passed
- TypeScript, ESLint, build, and schema manifest checks passed


ziemekobel:
Hi @ken8203! While testing this branch I found a scenario where a key stalls permanently: an older job deferred with `startAfter`, plus a newer same-key job fetched via the deferred-job carve-out that lands in `retry`. Once the older job becomes eligible it claims the key's head slot but is blocked by the retry sibling, and the retry job can no longer be re-fetched because it's not the head — fetch returns empty for that key forever. (master stalls identically today via the swallowed `job_i8` unique violation, so this branch is a good place to close the gap.)

I've opened ken8203/pg-boss#1 against this PR's branch with a fix (order head candidates by `state DESC` so a retry job keeps its key's head position — `job_i8` guarantees it's the only fetchable job for the key anyway) plus a regression test covering both partition modes.

ken8203:
> Hi @ken8203! While testing this branch I found a scenario where a key stalls permanently: an older job deferred with `startAfter`, plus a newer same-key job fetched via the deferred-job carve-out that lands in `retry`. Once the older job becomes eligible it claims the key's head slot but is blocked by the retry sibling, and the retry job can no longer be re-fetched because it's not the head — fetch returns empty for that key forever. (master stalls identically today via the swallowed `job_i8` unique violation, so this branch is a good place to close the gap.)
> 
> I've opened [ken8203#1](https://github.com/ken8203/pg-boss/pull/1) against this PR's branch with a fix (order head candidates by `state DESC` so a retry job keeps its key's head position — `job_i8` guarantees it's the only fetchable job for the key anyway) plus a regression test covering both partition modes.

Hi @ziemekobel , thanks for the fixing, the pr has been merged! ❤️

ziemekobel:
Hi @timgit! Would you have some time to take a look at this PR, please?
Without this fix running `key_strict_fifo` in production risks blocking everything with one failing job.

timgit:
I'll try and merge this in a few days, yes.

PR #874, opened 2026-08-16
fix: stamp dead_letter on redrive from destination queue

Nazaire:
redrive() copied retry/retention/policy from the destination queue but omitted dead_letter, so redriven jobs got null. fail() routes to the DLQ via job.dead_letter, so a second terminal failure never dead-lettered again even when the queue still had deadLetter configured.

Match send()'s queue stamp and document deadLetter as part of the inherited destination config.

### DIFF
--- a/src/plans.ts
+++ b/src/plans.ts
@@ -2254,7 +2254,8 @@ export function insertDeadLetterJob (schema: string): string {
 // oldest-first, capped at $4. The JOIN in `candidates` only matches jobs whose destination queue
 // exists, so legacy/orphaned jobs (NULL source_name, no override) are never deleted — they stay
 // in the DLQ rather than being lost. Re-created jobs get a new id, `created` state, retry_count 0,
-// cleared output, NULL source_*, and the destination queue's current retry/retention/policy config.
+// cleared output, NULL source_*, and the destination queue's current retry/retention/policy/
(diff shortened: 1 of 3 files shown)


src/plans.ts
```diff
@@ -2276,11 +2277,12 @@ export function redriveJobs (schema: string, table: string): string {
...
       INSERT INTO ${schema}.job
         (name, data, priority, retry_limit, retry_backoff, retry_delay, retry_delay_max,
-         expire_seconds, keep_until, deletion_seconds, policy, singleton_key, heartbeat_seconds)
+         expire_seconds, keep_until, deletion_seconds, policy, singleton_key, heartbeat_seconds,
+         dead_letter)
       SELECT COALESCE($2, m.source_name), m.data, m.priority, q.retry_limit, q.retry_backoff,
         q.retry_delay, q.retry_delay_max, q.expire_seconds,
         now() + q.retention_seconds * interval '1s', q.deletion_seconds, q.policy,
-        m.singleton_key, m.heartbeat_seconds
+        m.singleton_key, m.heartbeat_seconds, q.dead_letter
```
Nazaire, line 2285: we should also keep `group_id` and `group_tier` ?

fritzstauff (approved):
LGTM!

PR #886, opened 2026-09-02
feat: pluggable recurrence kinds on schedule()

kibertoad:
## Problem

`schedule()` accepts a cron string and nothing else. The timekeeper decides what is due by evaluating `cron-parser`'s `prev()` against database time every 30 seconds and sending when `prevDiff < 60`, deduplicating across instances with `singletonSeconds: 60`. There is nowhere to plug in another expression language, and the dedup throttle also caps a schedule at one job a minute.

## What this does

Makes the expression kind explicit and lets the process register parsers, the same way it registers `work()` handlers:

```ts
const boss = new PgBoss({
  connectionString,
  recurrences: {
    rrule: {
      // pure function; pg-boss never stores or serialises it
      next: (expression: string, after: Date, tz: string) => Date | null,
      validate: (expression: string, tz: string) => void   // throw to reject at schedule() time
    }
  }
})

await boss.schedule('run-workflow',
  { kind: 'rrule', expression: 'DTSTART;TZID=Europe/Berlin:20260901T090000\nRRULE:FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR' },
  { workflowId },
  { key: workflowId, missed: 'once' })
```

A plain string stays `{ kind: 'cron' }`, and `cron` is the one built-in kind. No new dependency.

## Storage

Migration v41 adds `kind text NOT NULL DEFAULT 'cron'`, `next_run_at timestamptz` and `last_run_at timestamptz` to `schedule`. `schedule()` computes `next_run_at` with the registered parser at insert time, and the pass claims what has come due in one statement:

```sql
WITH due AS (
  SELECT name, key, next_run_at, last_run_at AS prior_run_at
  FROM pgboss.schedule
  WHERE next_run_at IS NOT NULL AND next_run_at <= now() AND kind = ANY($1::text[])
  ORDER BY next_run_at
  FOR UPDATE SKIP LOCKED
)
UPDATE pgboss.schedule s
SET last_run_at = due.next_run_at, next_run_at = NULL
FROM due
WHERE s.name = due.name AND s.key = due.key AND s.next_run_at = due.next_run_at
RETURNING ...
```

Then, per row, the jobs are forwarded and `next_run_at` is written back from `parser.next(...)`. Cross-instance dedup is the row claim rather than `singletonSeconds: 60`, which also lifts the minute granularity for kinds that want finer resolution. The `next_run_at = due.next_run_at` re-check is what actually makes the claim exclusive, so `noSkipLocked` backends (CockroachDB) run the same statement with no locking clause.

Instances without a parser for a stored kind skip those rows and warn (`unsupported_recurrence`), mirroring how a queue with no `work()` handler is simply not fetched.

Schedules with no pending occurrence are re-anchored by a repair step at the top of each pass: rows upgraded from v40 (`next_run_at` null), and rows whose process died between claiming and rescheduling. Repair never replays the occurrence it finds parked, since a claim that never reached the queue may equally well have reached it and died after.

## `missed` policy

With `next_run_at` persisted, the pass knows exactly which occurrences were skipped while no instance ran, so `missed: 'skip' | 'once' | 'all'` is exposed on `schedule()`. Default `skip` preserves today's behaviour. `all` is capped at 1000 occurrences per schedule, with a `missed_occurrences_capped` warning when it truncates.

## Backwards compatibility

- `schedule(name, cronString, ...)` is unchanged, including that an occurrence which has only just passed fires immediately: `schedule()` anchors the first occurrence one grace window back, which is what per-pass `prev()` evaluation did. An expression that recurs faster than that window is anchored ahead instead, so a second-resolution kind does not start life owing a backlog.
- The `cron` column still holds the expression whatever the kind. `getSchedules()` reports it as both `cron` and `expression`, and adds `kind`, `nextRunAt`, `lastRunAt`.
- v39 rows migrate with `next_run_at` null and are anchored by the first pass.


timgit:
Thanks for opening. I'm considering making RRULE a first-class dependency instead of doing the constructor-time custom code. rrule-temporal looks like a better fit than rrule atm.

kibertoad:
@timgit I'd suggest considering https://www.npmjs.com/package/rrule-rust?activeTab=readme in that case, it's highly performant and robust.

I'm happy to open first-class support PR as well in case you confirm that this is a preferred direction.

timgit:
I'd prefer to avoid native abstractions unless there's a need. They can be complicated to install sometimes.

kibertoad:
@timgit I've opened https://github.com/timgit/pg-boss/pull/896 adding support for rrule specifically, could you review the approach?

kibertoad:
Superseded by #898, which answers the same request with a fraction of the code: no registration API, no second `schedule()` signature, and no persisted `next_run_at`, which is where all three of the defects reproduced on this PR came from. Closing in favour of that one.

Rules are a first-class expression format there rather than a plugin point, so `schedule()` takes `'FREQ=MONTHLY;BYDAY=-1FR;BYHOUR=17'` wherever it takes a cron expression, and the row records which of the two formats it holds in a `kind` column that the pass corrects when it disagrees with the expression beside it.

The one piece worth carrying forward from here on its own merits is the `missed: 'skip' | 'once' | 'all'` catch-up policy. It is separable from the recurrence format and it is not in #898, so it can land as its own change.

PR #887, opened 2026-09-03
feat(schedule): add getSchedule() and previewSchedule()

kibertoad:
Two small additions to the scheduling API, both aimed at code that manages schedules on behalf of someone else: admin tooling, and libraries that wrap `schedule()` behind their own facade.

### `getSchedule(name, key)`

`(name, key)` is the primary key of the `schedule` table, so `getSchedules(name, key)` can only ever return zero or one row, but it still returns an array. Every caller that wants "the schedule for this key" writes the same destructure-then-check dance:

```js
const [schedule] = await boss.getSchedules('report', 'eu')
if (!schedule) { /* ... */ }
```

`getSchedule(name, key)` returns the row or `null`. `key` defaults to `''`, matching `schedule()`.

### `previewSchedule(cron, options)`

Answering "when does this next fire?" currently means reaching for `cron-parser` directly in the calling code. That has two problems:

1. It duplicates the validation `schedule()` performs, and can drift from it. Getting it wrong is easy: the time zone check has to force a `CronDate` construction, otherwise `cron-parser` accepts `America/New_Yrok` and only fails later.
2. It evaluates against the local clock, while the cron pass evaluates against database time. On an instance with real skew the two disagree, which is exactly the case a preview is meant to explain.

`previewSchedule()` returns the next occurrences as `Date[]`:

```js
boss.previewSchedule('0 3 * * *', { tz: 'America/Chicago', count: 3 })
```

- Validation is shared with `schedule()`. The expression and time zone checks moved into `assertRecurrence()`, which both now call, so anything `previewSchedule()` accepts can also be stored.
- `from` defaults to database time (instance clock plus the cached skew), the same reading `shouldSendIt()` uses.
- Occurrences are strictly after `from`, so handing the last one back in yields the next page.
- `count` defaults to 5 and is capped at 1000, so a per-second expression cannot be asked for an unbounded walk.
- No query and no started instance required. It is pure arithmetic.

Pairs naturally with the read above:

```js
const schedule = await boss.getSchedule('report', 'eu')
const upcoming = boss.previewSchedule(schedule.cron, { tz: schedule.timezone })
```

### Notes

- No schema change, no migration, no behaviour change to existing methods. The only edit to `schedule()` is the extraction of its two validation calls into the shared helper.
- Tests in `test/scheduleReadTest.ts`: `getSchedule` key selection and the null case, and `previewSchedule` covering time zone evaluation, the skew-derived default reference point, strict-exclusive paging, and each validation failure.
- Docs added to `docs/api/scheduling.md`, including the caveat that a preview describes the expression rather than the delivery, since the cron pass runs on `cronMonitorIntervalSeconds`.


kibertoad:
Folded into #898, which carries this branch as its first commit and combines it with #888 and #896. Closing in favour of that one; nothing here is dropped.

`previewSchedule()` there also reads a recurrence rule, told apart the way `schedule()` tells the two formats apart, so a stored schedule can be previewed from its `cron` column whichever format it holds. A finite rule answers with what is left of it and a spent one with nothing, so the liveness check `schedule()` makes lives apart from the readability checks the two share.

Issue #880, opened 2026-08-25
docs: fromDrizzle doesn't work with the drizzle-orm/bun-sql driver (Bun.SQL double-encodes ::json string params)


shayaansultan:
**Versions:** pg-boss 12.28.0 · drizzle-orm 0.45.2 · Bun 1.4.0 · PostgreSQL 16.15

The [adapters docs](https://github.com/timgit/pg-boss/blob/master/docs/api/adapters.md) note node-postgres and postgres-js support for the Drizzle adapter. To save others some debugging: the combination with Drizzle's **bun-sql** driver fails on every `send()`:

```
PostgresError: cannot call json_to_recordset on a scalar (22023)
```

**This is not a pg-boss bug.** pg-boss correctly pre-stringifies the job batch for `json_to_recordset($1::json)`; Bun.SQL then JSON-encodes the string parameter *again* in json context, so Postgres sees a scalar. Upstream report: oven-sh/bun#28819. There's also no clean adapter-side fix — binding the raw array instead would break node-postgres, which serializes JS arrays as Postgres array literals.

**Suggestion:** a one-line note in the Drizzle section of the adapters docs ("the bun-sql driver is currently incompatible — see oven-sh/bun#28819") until the Bun fix lands. Happy to PR that note if you'd take it.

<details>
<summary>Repro script (fails on the <code>send</code>; the same script on <code>drizzle-orm/node-postgres</code> works)</summary>

```ts
// bun add pg-boss drizzle-orm && DATABASE_URL=postgres://... bun repro.ts
import { sql } from 'drizzle-orm'
import { drizzle } from 'drizzle-orm/bun-sql'
import { fromDrizzle, PgBoss } from 'pg-boss'

const url = process.env.DATABASE_URL!
const db = drizzle(url)
const boss = new PgBoss(url)
boss.on('error', console.error)

await boss.start()
await boss.createQueue('demo')

await db.transaction(async (tx) => {
  // throws: PostgresError: cannot call json_to_recordset on a scalar (22023)
  await boss.send('demo', { hello: 'world' }, { db: fromDrizzle(tx, sql) })
})
```

</details>

PS: I am new to open-source and used Claude for doing all of this. If there is anything wrong with how I wrote the issue or anything lmk! Would love any feedback 🙏

timgit:
[...]

PR #888, opened 2026-09-03
feat(schedule): record the last job a schedule created

kibertoad:
A schedule row records what should happen and when, but nothing about what did happen. Going from a schedule to the job it produced means guessing: scan the queue for a job whose payload matches `schedule.data`, or correlate on time and hope the queue has no other traffic. Neither works once two keys on the same queue send the same payload.

This adds `last_job_id` to the `schedule` table and exposes it as `lastJobId`, so a schedule points at the job it most recently created.

```js
const [schedule] = await boss.getSchedules('report', 'eu')

if (schedule.lastJobId) {
  const [job] = await boss.findJobs('report', { id: schedule.lastJobId })
  console.log(job?.state)
}
```

An operator looking at a schedule can reach its last run without correlating on payload or timestamp.

### How it is written

The cron pass already inserts one send-it job per due schedule. It now carries the schedule's `key` alongside `name` in that payload, so the send-it handler knows which row an occurrence came from, and records the ids for a whole batch in one statement via a JSON recordset. Batch size does not drive the statement count.

- `send()` resolving `null` (a throttle or queue policy dropped the job) records nothing, since there is no job to point at.
- A batch holding two occurrences of one schedule is reduced to the latest before it reaches SQL. `UPDATE ... FROM` leaves it unspecified which source row wins when several join one target, so passing both would make the stored id arbitrary rather than latest. A worker that fell behind can pick up two minute buckets at once, so this is reachable.
- The update is best effort. The schedule fired and the job exists; failing to annotate the row must not fail the send-it job and replay the occurrence, so the failure is reported on `error`, named with the schedules it covers. The report itself cannot escape either: an `error` event with no listener is a throw by Node's contract, and that throw would replay the very batch it was reporting on.
- `updated_on` is deliberately untouched. It tracks edits to the definition, and a firing schedule has not been edited.

Since the annotation is a second statement rather than part of the send, `null` means "no run recorded", not "never fired". A schedule that last fired before this upgrade reads `null` until its next run, and so does one whose annotation failed after its job was already created. The docs and the `Schedule` JSDoc say so, and the queue's job history stays the authority on what actually ran.

### Send-it singleton key

The cron pass deduplicated occurrences on `` `${name}__${key}` ``. Underscores are legal in both a queue name and a schedule key, so `('report_', 'daily')` and `('report', '_daily')` produced the same key, and the 60-second singleton window dropped whichever occurrence lost the race: one of the two schedules silently never fired. The key is now `JSON.stringify([name, key])`, which no pair can collide on.

An instance still on the old format writes the old key, so a mixed-version deployment can fire a schedule twice in the minute a rollout straddles.

### Rolling upgrade

`key` is read as optional rather than required, so the payload change is safe in both directions:

- A send-it payload written by 12.29.0 has no `key`, so a 12.30.0 handler forwards the job and skips the annotation.
- A payload written by 12.30.0 has an extra `key`, and `checkSendArgs()` reads only `name`, `data`, and `options` off a request object, so a 12.29.0 handler ignores it.

### Schema

Migration v40: `ALTER TABLE schedule ADD COLUMN IF NOT EXISTS last_job_id uuid`, with the matching drop as the rollback. Nullable and unconstrained on purpose: the referenced job is subject to the queue's retention policy and will eventually be deleted, so a foreign key would either block retention or null the column back out. That caveat is documented.

### One behaviour change to call out

`getSchedules()` previously ran `SELECT *`, so `created_on` and `updated_on` reached callers in snake_case while every other read in the API returns camelCase. Adding a column that way would have produced `last_job_id`, so the schedule reads now use an explicit aliased column list and return `createdOn`, `updatedOn`, and `lastJobId`. Code reading `schedule.created_on` or `schedule.updated_on` needs the camelCase name. Neither was in the `Schedule` type before, so TypeScript callers were not depending on them.

`@pg-boss/proxy` follows: its `scheduleSchema` carries the three new fields, so the generated OpenAPI contract exposes them and the type-drift check stays in step with `types.Schedule`.


toruiwasa:
Sorry for the noise from #897, which referenced this PR and pinged you. That reference was wrong and I have removed it.

`setScheduleLastJobIds` does bind a stringified value in front of `json_to_recordset($1::json)`, which is the #880 pattern, and the statement does fail under Bun.SQL. What I missed is that the driver cannot get there. It runs on the constructor's `db`, and constructing pg-boss with a bun-sql adapter fails at `start()` on `BEGIN` ("Only use sql.begin, sql.reserved or max: 1"). #880 reaches it because `send()` takes a `db` per call.

Please disregard it.

timgit:
Thanks for opening.  Regarding sequencing, I don't mind if you want to merge these 3 PRs into 1 since they are all related to scheduling.  It seems the most natural order would be to add rrule first, then layer these after that work, but that would produce 2 successive schema migrations.  How do you feel about merging them?

kibertoad:
Folded into #898, as proposed here. Closing in favour of that one; nothing here is dropped.

It carries this branch as its second commit and puts `last_job_id` and #896's `kind` in a single v41 migration, so the collision between the two is gone and a database takes one pass over the `schedule` table instead of two. The send-it payload there carries this PR's `key` and JSON singleton key on cron and rule occurrences alike.

PR #890, opened 2026-09-03
feat(work): add transactional workers

kibertoad:
## The problem

A handler that writes to the database has two things to persist: whatever it wrote, and the fact that the job is done. Today those are two separate commits with a gap in between, and a crash in that gap leaves the wrong outcome either way round:

- the handler's rows committed, the job still `active`: the job is retried and the work happens twice (a second charge, a duplicate ledger row, another email)
- the job completed, the handler's rows lost: the work never happened and nothing will retry it

The usual answer is to make every handler idempotent. That is real work, and often not possible when the side effect lands in a table you do not own.

pg-boss already lets you put *its* writes in a transaction you control: `send`, `complete`, `fail` and friends all take a `{ db }`. `work()` was the gap: it never handed the handler a transaction, so getting this one property meant giving up `work()` and rebuilding it around a manual `fetch()` / `complete()` pair, with your own polling loop, batching, backoff, error handling and shutdown behaviour.

## What this adds

`transactional: true` on `work()`. The handler gets a second argument: a `db` bound to a transaction pg-boss opened for this batch, and pg-boss commits the job's completion in that same transaction.

```js
await boss.work('charge-customer', { transactional: true }, async (jobs, tx) => {
  const [job] = jobs

  await tx.executeSql('INSERT INTO ledger (customer, amount) VALUES ($1, $2)', [job.data.customer, job.data.amount])

  // commits with the ledger row and with this job's completion
  await boss.send('send-receipt', { customer: job.data.customer }, { db: tx })
})
```

Either all of it commits or none of it does. There is no window where the side effect exists and the job does not, or the reverse.

`tx` is the ordinary pg-boss `db` interface, so it composes with every call that already takes `{ db }`, and with the ORM adapters if the handler wants to drive its own writes.

## How it works

The job is claimed the normal way, on a pooled connection, before any transaction is opened. The transaction wraps the handler and the completion, and nothing else.

That single decision is what keeps the rest of a worker unchanged:

- **The job is `active` and visible to every other connection while the handler runs.** `expireInSeconds`, heartbeats, `failWip` on shutdown, and another instance's supervisor all see it and can act on it, exactly as they do for any other worker.
- **A handler that throws rolls back its own writes and the completion**, and the job is then failed on a pooled connection. Retry counts, retry delays and dead lettering are the ordinary code path, not a second implementation of it.
- **A process that dies mid-handler leaves an `active` row with its attempt recorded**, and the expiration reclaims it. Nothing is erased, and a payload that kills the process still exhausts its retries and dead-letters.
- **Fetching, batching, `localConcurrency`, group limits and heartbeats are untouched**, because none of them are inside the transaction. An idle poll is still a single statement.

`perJobResults` is the one option rejected: it settles each job in a batch separately, and one transaction has a single outcome.

## The new primitive: `Db.beginTransaction()`

`withTransaction(fn)` serves callers that can express the whole transaction as one function. A worker cannot: it decides how the transaction ends after the handler has already returned or thrown. `beginTransaction()` returns a handle (`db`, `commit()`, `rollback()`), and `withTransaction` is now a wrapper over it, unchanged in behaviour.

The handle owns its connection and releases it exactly once:

- `rollback()` is idempotent and never rejects. It is already the error path, and the caller holds an error worth more than "ROLLBACK failed".
- `rollback()` is bounded. Statements queue per connection, so a ROLLBACK issued while a statement is still in flight (an abandoned handler that blew its deadline, waiting on someone else's row lock) would wait as long as that statement does. Past the deadline the connection is dropped instead, and the server aborts the transaction with it.
- A failed `COMMIT` releases with the error, so the pool discards a connection left in an unknown state.
- The handle attaches its own `error` listener to the pinned connection. pg-pool takes its idle listener off a client for as long as it is checked out, so a connection that drops mid-transaction would otherwise emit `error` with nothing listening, which ends the process.
- `commit()` and `rollback()` refuse to run once settled, and the handle's `db` refuses statements, so a late call cannot reach a connection the pool has already handed to someone else.

Declared on `IDatabase` as an optional capability, the way `listen` is, so an adapter can opt in. `work()` asserts on it, so a `db` without it fails on the call rather than on the first fetch.

## Costs, in the docs

- **Each handler in flight holds a pool connection for its own duration.** Size `max` above `localConcurrency`, summed over the transactional queues. pg-boss emits a `transactional_pool_headroom` warning from `work()` when the built-in pool has no room left, rather than letting fetches and failures quietly wait out `connectionTimeoutMillis`.
- **A long transaction holds its snapshot and holds vacuum back**, so this suits handlers that finish in seconds. For long work, keep the default worker and pass `{ db }` to `complete()`.
- **A SQL error the handler catches leaves the transaction aborted**, and Postgres then rejects the completion too, so the job fails even though the handler returned. Either let such errors propagate, or isolate the statement behind a `SAVEPOINT` of your own. The recorded failure says which of the two happened.

## Also in here

`failWip()`, the last step of a shutdown, no longer lets one worker's failure skip the remaining workers, the pool close, the stopped flag, and the `stopped` event.


timgit:
Review findings:

---

## 1. (blocker) An aborted worker commits a half-executed handler

No external actor, no race to win: `stop({ graceful: false })` with a transactional handler in
flight hits this every time, as does a graceful stop whose timeout expires.

`resolveWithinSeconds` **resolves** when its `AbortController` fires — `delay()` ends with
`.catch(resolve)`, so an abort is indistinguishable from the handler returning:

```ts
const result = await resolveWithinSeconds(handling, maxExpiration, `handler execution exceeded ${maxExpiration}s`, ac)
// abort lands here with result === undefined, handler still running
```

`failWip()` calls `worker.abort()`. So a non-graceful `stop()` — and a graceful one whose timeout
expires — drops `#processJobs` into the **success** branch while the handler is mid-flight: it runs
the completion (which settles nothing, `failWip` just failed the job) and then
`await transaction.commit()`.

Reproduced (`repro3.mts`): handler inserts `first-half`, sleeps 3s, inserts `second-half`;
`stop({ graceful: false })` fires 300ms in.

```
committed side effects: [ 'first-half' ]
second write error: Transaction is already settled
job: { state: 'failed', output: { value: 'pg-boss shut down while active' } }
```

Half the handler's work is committed and the job is failed. With `close: true` the commit lands 57ms
into the stop. This is the PR's own opening example — the side effect exists, the job does not.

**Fix.** Treat an abort as a failure on the transactional path, before the completion:

```ts
const result = await resolveWithinSeconds(handling, maxExpiration, `handler execution exceeded ${maxExpiration}s`, ac)

if (transaction && ac.signal.aborted) {
  throw new Error('worker was aborted before the handler finished')
}
```

Non-transactional behaviour is unchanged; the transactional path rolls back and takes the ordinary
`fail()` route it already takes for a throw.

## 2. (medium) A lost claim still commits the handler's writes

Anomaly-only: something outside the worker has to settle the job while the handler runs. In
likelihood order that is `cancel()` / `fail()` / `deleteJob()` from an operator or the dashboard; the
heartbeat path, when a transient `touch()` error lets `failJobsByHeartbeat` reclaim a live job; and
expiry, which is rarer than it looks — `resolveWithinSeconds` counts the same `expireInSeconds` from
handler start while the supervisor only notices on its next pass, so the local timeout normally fires
first and *rejects*, which rolls back correctly. It loses that race only when the event loop is
starved or a pass lands early.

Rare, but what it costs when it lands is precisely the duplicate the feature is sold to prevent, and
the fix is free on healthy batches — so documenting it as a known limitation is a legitimate call,
just the weaker one.

`#processJobs` never looks at `completion.affected`. When the claim is lost while the handler runs —
`expireInSeconds` expiry, heartbeat failure, another instance's supervisor, a peer's `failWip` — the
in-transaction `complete()` updates **0 rows** and the transaction commits the handler's writes
anyway. The job is then retried and the side effect happens twice.

Reproduced (`repro2.mts`): handler inserts a ledger row; a peer fails the job mid-handler (what
expiration does); the handler returns.

```
state while handler still running: retry
side effects committed: 2 [ 1, 2 ]
final job state: completed retryCount 1
```

Two ledger rows for one job — the duplicate charge the PR opens with.

The existing test *"should let the supervisor reclaim a transactional job whose handler never
returns"* stops one step short of this: it reclaims the job, then calls `releaseHandler()` and never
asks what the transaction did with the handler's writes.

**Fix.** `complete()` already reports `affected`, so the healthy batch needs no new SQL at all:
`affected === jobIds.length` commits as it does today. Only a *short* completion is ambiguous, and it
is ambiguous in exactly two ways — the claim was lost, or the handler settled the jobs itself through
`tx` (the documented pattern, which also leaves pg-boss's own completion with 0 affected). One read,
inside the transaction and behind that gate, separates them:

```ts
const completion = await this.complete(name, jobIds, ..., { db: transaction.db })

if (transaction && completion.affected < jobIds.length) {
  const { rows } = await transaction.db.executeSql(
    `SELECT id, state FROM ${schema}.${table} WHERE name = $1 AND id = ANY($2::uuid[])`, [name, jobIds])

  if (rows.some(r => r.state !== 'completed')) throw new Error('the claim was lost while the handler ran')
}

await transaction.commit()
```

The throw takes the rollback and the ordinary `fail()` path that a handler throw already takes; a
`fail()` on a job that is no longer active is a no-op, so nothing else needs changing. The extra
statement never runs on a healthy batch — only where something has already gone wrong.

Two alternatives if even the anomaly-path read is unwanted:

- **Track the handler's own settle.** `tx.db` is pg-boss's own object and `Manager.complete/fail/
  cancel` already receive `options.db`; comparing identity against the open transaction records which
  ids the handler settled, making the rule `affected + handlerSettled === jobIds.length` with zero
  round trips. Costs a small per-transaction map, read only on the transactional path.
- **Narrow the contract.** Declare that the worker owns settlement and drop the "handler completes the
  job itself through `tx`" pattern and its test, the way `perJobResults` is already rejected. Then
  `affected === jobIds.length` is the whole check — no query, no state — at the price of a documented
  capability.

## 3. (minor) `failWip`'s error emit can still skip the remaining workers

The new `catch` reports through `this.emit(events.error, …)`. `#promoteEvents` re-emits it on the
PgBoss instance, and with no user `error` listener Node throws `ERR_UNHANDLED_ERROR` — out of the
catch, out of `failWip`, skipping the remaining workers, the pool close, the `stopped` flag and the
`stopped` event. That is the exact regression the change is meant to prevent.

Reproduced (`repro4.mts`, no `error` listener attached): `stop()` rejects with `ERR_UNHANDLED_ERROR`
and the second queue's job is still `active` afterwards. The PR's test attaches a listener, so it
passes. On master the raw error escaped too, so this is an incomplete fix rather than a regression —
but it is one line:

```ts
} catch (err: any) {
  if (this.listenerCount(events.error)) {
    this.emit(events.error, { ...err, message: err.message, stack: err.stack, queue: worker.name, worker: worker.id })
  }
}
```

## 4. (minor) The new warning is outside the warning system

`transactional_pool_headroom` is documented in `docs/api/events.md` but:

- it is **not** in `types.WarningType`, so the documented type union and the docs now disagree, and a
  typed consumer (`PersistedWarning`, the dashboard) cannot name it;
- it is emitted with a bare `this.emit` rather than `emitAndPersistWarning`, so it never reaches the
  `warning` table under `persistWarnings`, unlike every other warning type.

Manager has no `WarningContext` today, so persisting means passing one in — reasonable to skip, but
then say so; the union entry should land either way.

kibertoad:
All four addressed in 2910517. Two of the three behavioural fixes landed differently from the suggestion, both because the suggested version does not work as written; details below.

## 1. Aborted worker commits a half-executed handler

Fixed, but not on `ac.signal.aborted`. That check fires on **every** completion: `resolveWithinSeconds` cancels its own timeout through the same controller in its `finally`, so the signal is already aborted by the time control returns. With the suggested line in place, 8 of the 20 existing transactional tests failed, every job carrying the abort error as its output.

`Worker.abort()` records the abort instead, which is precise because `failWip` is its only caller, and the manager resets the flag per batch beside `abortController`:

```ts
if (transaction && worker?.aborted) {
  throw new Error('pg-boss shut down while the handler was still running')
}
```

Test: `should roll back a handler abandoned by a shutdown`. The handler writes `first-half`, sleeps, then writes `second-half`; the stop lands in between. It waits for the transaction to start refusing statements before reading the table, since an uncommitted insert is invisible from another connection either way and an empty table proves nothing until the transaction is over. Reverting the guard reproduces your repro3: `expected 0 to be 1`.

## 2. Lost claim commits the handler's writes

Fixed with your first alternative, tracking the handler's own settles, rather than the in-transaction read. The read cannot actually separate the two cases: an external `fail()` leaves the row in `retry`, and so does a handler calling `fail(..., { db: tx })`. `rows.some(r => r.state !== 'completed')` reads both as a lost claim, so a handler that fails or cancels its own jobs through `tx` gets rolled back and re-failed with pg-boss's error instead of its own.

Identity tracking has no such case. `complete`/`fail`/`cancel`/`deleteJob` record their ids when `options.db` is the open transaction, `#processJobs` takes the set before running its own completion (which goes through the same `complete()` and would otherwise record itself), and the rule is `affected + handlerSettled === jobIds.length`. Zero round trips on both the healthy path and the anomaly, and the documented handler-settles-it-itself pattern keeps working unchanged.

Test: `should roll back when the claim is lost while the handler runs`, your repro2 as an assertion. The handler writes a ledger row and, on the first attempt only, fails the job on a pooled connection. One row for two attempts. Reverting the check gives `expected 2 to be 1`.

Worth recording: the ordinary `fail()` that follows is not always a no-op. `failJobsById` matches `state < 'completed'`, and the state enum orders `created < retry < active`, so a job a peer moved to `retry` gets failed again and spends another retry. Better than the duplicate, and it is what any failing worker does, so I left it and dropped the "no-op" claim from the comment.

## 3. `failWip`'s error emit skips the remaining workers

Fixed, but `listenerCount` does not detect this one. `#promoteEvents` attaches a listener to the Manager for every event, so the count here is always at least 1; the `ERR_UNHANDLED_ERROR` comes from the **PgBoss** re-emit, one frame further out than this emitter can measure. Guarding on the count leaves the bug exactly as it was.

The emit is wrapped instead. Test: `should finish the shutdown when a wip failure has no error listener`, the existing wip test with the listener removed. Without the fix it fails on `stop()` rejecting, with the stack running through `src/index.ts:119` (`#promoteEvents`).

## 4. Warning outside the warning system

`transactional_pool_headroom` is in `types.WarningType`, and the emit goes through `emitAndPersistWarning` with a `WarningContext` getter on Manager, shaped like Timekeeper's. `persistWarnings` covers it like every other type. Test: `should persist the pool headroom warning under persistWarnings` reads the row back out of the `warning` table.

timgit:
Re-reviewed at `2910517`. All four findings are properly addressed and I can reproduce the fixes —
shutdown mid-handler now leaves nothing committed, a stolen claim rolls back, and `stop()` finishes
without an `error` listener attached. 1215/1215 on PG18, 23/23 transactional including
`DISTRIBUTED=true`, lint and tsc clean. The `Worker.aborted` flag is the right call over the signal,
and the comment explaining why `resolveWithinSeconds` makes the signal useless here is the kind of
thing I want in this file.

Three things before merge.

### 1. The claim guard counts a settle the handler asked for, not one that landed

`#trackHandlerSettle` records the ids before the statement runs and never looks at what it affected:

```ts
async complete (name, id, data?, options = {}) {
  ...
  this.#trackHandlerSettle(options, ids)   // records ids, then runs the SQL
```

So a handler settle that updates **0 rows** — which is exactly what a lost claim leaves behind — is
recorded as if it had settled the job, and `#assertClaimHeld` accounts for it. Both terms go to 0 and
1 for a job nobody settled, and the batch commits.

That puts the duplicate right back, through the pattern this same commit documents. Handler writes a
ledger row, a peer fails the job mid-handler (what expiry does), handler then self-settles through
`tx`:

```
state while handler still running: retry
handler self-complete affected: 0
side effects committed: 2 [ 1, 2 ]
final job state: completed retryCount 1
```

Two ledger rows for one job. The new "should roll back when the claim is lost" test covers the
claim-taken-away case only — it never has the handler settle its own job, which is what hides it.

`mapCommandResponse` gives a count rather than the ids, so an all-or-nothing gate is the cheapest
sound rule:

```ts
const result = await db.executeSql(sql, [name, ids, outputData])
const response = this.mapCommandResponse(ids, result)

if (response.affected === ids.length) this.#trackHandlerSettle(options, ids)

return response
```

A short handler settle then records nothing and the guard rolls back, which is the safe direction.
Same move in `fail`, `cancel` and `deleteJob` — all four track ahead of their statement today.

### 2. Back out the warning persistence — that ask was mine and it was wrong

I asked for two things on the warning and hedged the second. Persisting it is the half to drop.

`work()` calls `#warnOnTransactionalPoolHeadroom` once per registration, and `pinned` counts the
transactional workers already registered, so once it crosses `max` every *later* transactional
`work()` call trips too, each with a larger `pinned` and so a different message. The comparison is
against `config.max` — static configuration, nothing at runtime clears it, so the same rows are
written again on every process start. With `max: 2`, four transactional queues, two boots:

```
rows written: { n: 6, distinct_messages: 3 }
```

Multiply by queues, instances and deploys inside `warningRetentionDays`. `deleteOldWarnings` keeps it
from growing without bound, but it buries the warnings someone actually needs.

It is also the wrong kind of thing for that table. Every other persisted warning records an episode
and is deduped by a flag that clears when the condition does — `#warnedXminHorizon`,
`#warnedAutovacuumDisabled`, `#warnedBloat`, `#warnedFetchOptions`. Pool headroom is a configuration
assertion with no episode to record.

Reverting to the `this.emit` you had at `b3c2724` also fixes something the persistence introduced:
`emitAndPersistWarning` reports its own failure through an `error` event, which re-throws when
nothing is listening, so `work()` itself can now reject:

```
work() THREW: 42P01 - relation "...warning" does not exist
```

Keep the `types.WarningType` entry and the `docs/api/events.md` row — that half was right.

If you'd rather keep the persistence, it needs the same dedupe as every other one (a
`#warnedPoolHeadroom` flag before the emit), and the guard belongs inside the helper rather than at
the call site — `#tryWarnOnTransactionalPoolHeadroom` with a `try`/`catch` in the body, matching
`tryAdvisoryLock` / `trySetQueueMonitorTime`. Don't drop the `await` to fix it: a floating rejection
takes the process down, which is worse than the throw.

### 3. Please add the database-side bound here rather than in a follow-up

Nothing bounds the handler's transaction from the database. `plans.transaction()` pins `lock_timeout`
and `idle_in_transaction_session_timeout` for migrations, but `beginTransaction()` sets no GUCs, so a
transactional handler is bounded only by `resolveWithinSeconds` and by `rollback()`'s
`TRANSACTION_ROLLBACK_TIMEOUT_MS` race — both timers in the process that has already gone wrong. When
it is the process that failed (starved event loop, a driver wedged below the promise), the
transaction stays open and keeps advertising `backend_xmin`, which is the thing `monitorVacuum` and
the `xmin_horizon` warning exist to report. I'd rather not ship the feature that makes a long-lived
transaction reachable from user code without it.

I measured the options so this isn't an open design question — PostgreSQL 18.0 and CockroachDB
v26.2.2:

| GUC | PostgreSQL 18 | CockroachDB 26.2 | Bounds |
|---|---|---|---|
| `transaction_timeout` | `25P04`, connection killed | `57014`, connection killed | the whole transaction |
| `idle_in_transaction_session_timeout` | `25P03`, connection killed | connection killed, **no SQLSTATE** | the gap between statements |
| `statement_timeout` | `57014`, next statement `25P02` | same | one statement |

`transaction_timeout` is the bound we want, and it exists on CockroachDB and PostgreSQL 17+ but not
on 13–16, which we still support — `set_config` on it there is a hard error. Good news is that
picking between them needs no version parsing: `current_setting(name, true)` returns NULL for an
unrecognised parameter on both backends, so one statement at startup settles it. Prefer
`transaction_timeout` where it exists, fall back to `idle_in_transaction_session_timeout`. Skip
`statement_timeout` — it bounds one statement, and its `57014` leaves the next one at `25P02`, which
would make your new message tell the user they swallowed a SQL error.

For the default, derive it from `expireInSeconds` rather than matching it:

```ts
maxExpiration * 1000 + TRANSACTION_ROLLBACK_TIMEOUT_MS   // 5000, src/db.ts:17
```

`#processJobs` already computes `maxExpiration`. Matching exactly would be a coin flip on the
diagnosis — whichever timer lands first decides whether the user gets `handler execution exceeded
30s` and a clean rollback, or a killed connection — and the transaction is open wider than the
handler anyway, since BEGIN precedes it and the completion and COMMIT follow it, so an equal bound
kills a batch that succeeded at its commit. `expireInSeconds` is also already the supervisor's
reclaim bound, and I don't want a third actor on that instant. Derived this way it only ever fires
when the whole Node path failed, which is exactly when the server should be the one to give up. A
`work()` option can override it.

Set it with `SELECT set_config(name, $1, true)` rather than a built `SET LOCAL` string — one
parameterized statement, portable to drivers that reject multi-statement queries, and verified
transaction-local with no session leak on both backends. Issue it from `#processJobs` on `tx.db`
right after the begin rather than as a `beginTransaction()` argument: `beginTransaction` is public
`IDatabase` surface, so an adapter that ignores a new parameter would lose the bound silently.

One thing worth a test: on CockroachDB the idle timeout arrives as a bare `Connection terminated
unexpectedly` with no SQLSTATE, so the client `error` listener you added has nothing to report there.

### Smaller

- Settling the job with raw SQL on `tx` now rolls back with `the claim on 1 of 1 job(s) was lost`,
  where it committed at `b3c2724`. Fair narrowing — direct writes to the job table were never
  supported — but the new docs line reads as endorsing it. Worth naming what the guard actually
  recognises (`complete()`/`fail()`/`cancel()`/`deleteJob()` with `{ db: tx }`), and widening that
  error message, since "the claim was lost" is the wrong story for a handler that settled its own job
  by hand.

kibertoad:
All three addressed in ec8d44a, plus the smaller item. The warning ask I took as written (drop the persistence); the other two landed as suggested.

## 1. The claim guard counted a settle the handler asked for

Fixed as you wrote it. All four of `complete`, `fail`, `cancel` and `deleteJob` now record after their statement and only when `affected === ids.length`, so a short settle records nothing and `#assertClaimHeld` rolls back. `complete` and `fail` needed the distributed branch folded into the same exit for it, since both returned early before the response existed.

Test: `should roll back when the handler settles a claim it no longer holds`, your repro as an assertion. A peer fails the job mid-handler, the handler then self-completes through `tx`, and the attempt's own `affected` is asserted to be 0. One ledger row for two attempts. Reverting the gate gives your `expected 2 to be 1`.

## 2. Back out the warning persistence

Done. Back to the `this.emit` at `b3c2724`, and the helper is synchronous again, which takes `work()` off the path where a failed table write rejects the registration. `types.WarningType` and the `docs/api/events.md` row stay, and both the row and the persistence section now say the type is event-only and why.

Test: `should keep the pool headroom warning out of the warning table`, the old persistence test inverted. It asserts the event fires and the row does not appear under `persistWarnings: true`.

## 3. The database-side bound

In, as designed. `#processJobs` issues `SELECT set_config($1, $2, true)` on `tx.db` right after the begin, default `maxExpiration * 1000 + TRANSACTION_ROLLBACK_TIMEOUT_MS`, overridable with a new `transactionTimeoutSeconds` work option (`0` removes the bound; rejected without `transactional`).

Which GUC comes from one probe, `current_setting(name, true)` over both names, resolved on the first transactional batch and cached for the process. It runs on the transaction's own connection rather than at startup so `work()` stays off the database, and a failed probe is not remembered, so the next batch asks again instead of running unbounded forever.

`statement_timeout` stayed out for the reason you gave: its `25P02` on the next statement would have the aborted-transaction message blame a handler that swallowed nothing.

Tests:

- `should bound the handler transaction from the database side` reads `current_setting` from inside the handler and asserts `35s` for a 30s `expireInSeconds`, on whichever GUC the server has.
- `should honour an explicit transactionTimeoutSeconds` and `should leave the transaction unbounded at transactionTimeoutSeconds 0`.
- `should roll the handler back when the database gives up on its transaction` sets 1s under a handler that keeps issuing statements for 1.5s, and asserts the job failed with nothing committed. Skipped on a server without `transaction_timeout`, since the idle timeout would never fire against a handler that keeps talking.
- `should fail a transactional batch whose connection dies without a sqlstate` covers the CockroachDB shape you flagged: `pg_terminate_backend` under a live handler lands as the same bare disconnect, and the test asserts the batch fails, the writes roll back, and `stop()` still finishes.

## Smaller

The message is widened. It now reads `N of M job(s) were no longer claimed by this worker when the handler returned`, then names both readings: something took the claim away, or the handler settled the job with SQL of its own instead of `complete()`/`fail()`/`cancel()`/`deleteJob()` on `{ db: tx }`. The docs paragraph names the four calls the guard recognises and says plainly that a direct write to the job table is read as a lost claim.

Test: `should name the handler in the rollback when it settled the job with its own sql`.

timgit:
Thanks for the resubmisison.

---------------

Re-reviewed at `84debe3`. All three asks are addressed, and the shape of each is the one I'd have
picked: the settle gate reads the count instead of the ask, the headroom warning is back to an
event with the reasoning written into the code and the docs, and the database bound probes for its
GUC in one statement with no version parsing. 33/33 transactional on PostgreSQL 18, plus 80 on the
settle paths the `complete`/`fail` restructure touches (`complete`, `failure`, `cancel`, `delete`,
`workLifecycle`, `expire`, `flow`); tsc, eslint and `docs:readme --check` clean.

Four things, two of which I'd hold the merge for.

### 1. The count gate drops the ids that *did* land, so a partial settle rolls back a batch that worked

`#trackHandlerSettle` is all-or-nothing per call, so a settle that affects some of its ids records
**none** of them:

```ts
if (response.affected < response.requested) return
```

That is the right reading when the short count means a lost claim, but it is also what a handler
produces when it settles part of its own batch first. Cancel one job, then settle the batch:

```js
await boss.work(q, { transactional: true, batchSize: 2 }, async (jobs, tx) => {
  await tx.executeSql(`INSERT INTO ledger DEFAULT VALUES`)
  await boss.cancel(q, jobs[0].id, { db: tx })                                  // 1 of 1, recorded
  await boss.complete(q, jobs.map(j => j.id), undefined, { db: tx })            // 1 of 2, recorded as nothing
})
```

The one job that really was completed is never recorded, so the guard is one short and the whole
batch goes:

```
settles: [{ cancel: 1, complete: 1, requested: 2 }]
job states: failed failed
output: 1 of 2 job(s) were no longer claimed by this worker when the handler returned...
ledger rows: 0
```

Both jobs fail and the handler's writes are thrown away, with an error that says a claim was lost
when none was. That test passes at `2910517` and fails at `84debe3`, so it came in with this fix.

The information the gate is missing is which ids the statement actually touched, and the statement
is already in a position to say. Returning them costs no round trip:

```sql
-- plans.completeJobs / plans.cancelJobs
RETURNING id
...
SELECT COUNT(*), coalesce(array_agg(id), '{}'::uuid[]) AS ids FROM results
```

```ts
// mapCommandResponse
settled: result?.rows?.[0]?.ids,

// #trackHandlerSettle
const landed = response.settled ?? (response.affected < response.requested ? [] : response.jobs)
```

I ran that as a patch: the case above commits with one ledger row, and the 33 transactional plus 80
settle-path tests stay green. A new `settled` field rather than repointing `jobs`, since `jobs` is
the requested list and `completeTest`/`failureTest` assert it that way. `deleteJobsById` takes the
same edit; `failJobsById` goes through the shared `failJobs` builder, which the supervisor also
uses, so that one probably wants the ids behind a flag rather than for every caller.

### 2. On CockroachDB, a transactional worker on a queue with heartbeats fails every job

This one is not from this round's commits — it reproduces at `2910517` too — but it is the feature
failing wholesale on a backend we support, and the new `set_config` widens it, so I'd rather it not
merge unfixed.

A queue with `heartbeatSeconds`, a transactional worker, a handler that runs long enough for one
heartbeat to land:

```js
await boss.createQueue(q, { heartbeatSeconds: 10 })
await boss.work(q, { transactional: true, heartbeatRefreshSeconds: 0.5 }, async () => delay(1500))
```

```
attempts: 1
final state: failed
output: 40001 restart transaction: TransactionRetryWithProtoRefreshError: WriteTooOldError
```

`touch()` writes `heartbeat_on` on the claimed row from a pooled connection, by design, so the job
stays visibly `active`. On CockroachDB that write lands at a timestamp above the handler
transaction's, and the completion pg-boss runs inside it then cannot write the same row: serializable
has already returned rows to the client, so there is nothing to refresh and the batch dies with a raw
retry error rather than any of the messages this PR writes.

What changed at `84debe3` is the reach. Before, the transaction only had a timestamp to lose once the
handler itself ran something through `tx`; a handler that never touched `tx` was fine. `set_config`
is now the first statement in every transactional batch, so *every* handler on a heartbeat queue is
exposed. Confirmed by short-circuiting `#setTransactionTimeout`: the same test passes.

I don't think there's a cheap fix that keeps both — the heartbeat has to write that row from outside,
and CockroachDB will not let the transaction write it afterwards. The honest options are to skip the
in-flight heartbeat refresh for transactional workers on a distributed backend and let
`expireInSeconds` be the liveness bound there, or to reject `transactional` on those backends
outright the way it is already rejected without `beginTransaction`. Either way it needs a test and a
line in the requirements list, since right now the docs promise heartbeats work "as usual".

### 3. The bound replaces an operator's own, including a tighter one

`#setTransactionTimeout` writes its derived value unconditionally. A role that already carries a
stricter bound gets it widened for exactly the transactions that run longest:

```js
new PgBoss({ ...config, options: '-c transaction_timeout=3000' })
// handler observes: transaction_timeout = 35s
```

3 seconds the DBA set, 35 seconds the handler gets. `idle_in_transaction_session_timeout` is the more
likely one to find already set — managed providers and poolers set it as a matter of course — and it
is the GUC we fall back to on 13–16, so that is the common case, not the exotic one.

Taking the tighter of the two needs no extra round trip and no extra probe, because `pg_settings`
reports these unitless in milliseconds on both backends (PostgreSQL 18 `3000`, CockroachDB `0`),
which `current_setting` does not:

```sql
SELECT set_config($1,
  (SELECT CASE WHEN s.setting::bigint > 0 AND s.setting::bigint < $2::bigint THEN s.setting ELSE $2 END
     FROM pg_settings s WHERE s.name = $1), true)
```

Reading it per transaction rather than caching the value also keeps it correct per connection, which
a process-lifetime cache would not be.

### 4. A probe that cannot run fails every batch, forever

The re-probe is the right call over remembering a rejection, but it is paired with letting the
rejection out of `#setTransactionTimeout`, which fails the batch. Together they mean a backend or a
`db` adapter that cannot answer `current_setting(name, true)` doesn't lose the bound — it loses
transactional workers entirely, one failed job per batch with `probe refused` as the output, for the
life of the process. Your own new test is that behaviour, minus the repetition.

The bound is a backstop against a process that has already failed. I'd rather it degrade than take
the batch with it: catch the probe failure, emit a `warning` once, run the batch unbounded, and let
the next batch probe again exactly as it does now.

### Smaller

- `npm run test:cockroachdb:full` is 4 failed / 26 passed / 3 skipped on this file, and CI only runs
  `distributedDatabaseTest`, so none of it shows up on the PR. Two are the GUC assertions —
  CockroachDB answers `current_setting('transaction_timeout', true)` with `35000`, not `35s`, so
  `expect(bound).toBe('35s')` is a PostgreSQL-only spelling. The third is the new lost-claim test:
  on CockroachDB the handler's own `complete()` raises `40001` instead of returning `affected: 0`,
  so `selfSettled` is `[1]`, not `[0, 1]`. The batch still rolls back, so the guarantee holds, but
  the diagnosis the job records is a raw retry error rather than the message you wrote for it —
  worth a sentence in the docs' distributed caveats either way. The fourth is the heartbeat failure
  above.

kibertoad:
All four addressed in c1836b1, plus the smaller items. Two landed as written; two took a different
shape, and the reasoning is below.

## 1. The count gate dropped the ids that landed

Fixed as you wrote it. `completeJobs`, `cancelJobs` and `deleteJobsById` return `id` from their
`results` CTE and aggregate it beside the count, `mapCommandResponse` carries it as a new `settled`
field, and `#trackHandlerSettle` records `response.settled ?? (short ? [] : response.jobs)`. The
fallback is for the mutators that share the mapper and do not report ids (`resume`, `restore`,
`retry`, `touch`); none of them is a settle, so it never fires in practice.

`failJobsById` takes the ids behind a flag on the shared `failJobs` builder, as you suggested. The
supervisor's `failJobsByTimeout`/`failJobsByHeartbeat` go through the same builder and only ever read
the count, so they stay on `SELECT COUNT(*)` rather than shipping a whole sweep's ids back. The
distributed `completeDistributed` and `failDistributed` already had the rows in hand and now report
them too.

Test: `should commit when the handler settles part of its own batch first`, your repro as an
assertion. The handler cancels one of two jobs, then completes both; the batch commits, the states
come out `cancelled`/`completed`, and the ledger holds one row. Reverting the change gives your
`ledger rows: 0` with both jobs failed.

## 2. Transactional workers and heartbeats on CockroachDB

Reproduced at `84debe3` exactly as you describe, and I confirmed the reach: the same test passes on
YugabyteDB, whose snapshot isolation makes the second write wait rather than abort, and on Postgres
under `DISTRIBUTED=true`. So this is CockroachDB's isolation, not the distributed toggles, and it
gets a compatibility flag of its own rather than riding on `noMultiMutationCte`.

I took the second of your two options, narrowed. Skipping the refresh does not actually leave
`expireInSeconds` as the bound: `failJobsByHeartbeat` matches on `heartbeat_on + heartbeat_seconds`,
so a `heartbeat_on` nobody refreshes goes stale under a live handler and the supervisor reclaims
precisely the long-running jobs the docs point heartbeats at (the 2-hour video job with
`heartbeatSeconds: 60`). The batch would roll back, retry, and do it again, with nothing in the
output to say why. That is worse than today's raw `40001`, which at least fails immediately.

So `noTransactionalHeartbeat` refuses the combination instead. `work()` reads the queue and throws
when it configures `heartbeatSeconds`, and the check runs only on a backend that sets the flag, so
`work()` stays off the database everywhere else. A per-job `heartbeatSeconds` can still put a
heartbeat on a batch whose queue has none, which `work()` cannot see, so `#processJobs` asserts the
same thing from the batch's own jobs before the begin. Rejecting the pair rather than `transactional`
wholesale keeps the feature on CockroachDB, where the other 33 transactional tests pass.

Tests: `should reject a transactional worker on a heartbeat queue where the backend cannot carry
both` and `should fail a batch a per-job heartbeat put on a queue that has none`, both driven by a
`__test__noTransactionalHeartbeat` hook so they run on plain Postgres in CI rather than only in the
CockroachDB job. The existing refresh test is skipped under `isCockroachDb`. Requirements list,
`docs/database-backends.md`, and the compatibility flag table all carry it.

## 3. The bound replaced an operator's own

Fixed, using `pg_settings` as you suggested, though `LEAST` rather than the `CASE`: your version as
written is `42804 CASE types bigint and text cannot be matched`, because `$2::bigint` fixes the
parameter's inferred type and `THEN s.setting` is text.

```sql
SELECT set_config($1, LEAST($2::bigint, NULLIF((SELECT s.setting::bigint FROM pg_settings s WHERE s.name = $1), 0))::text, true)
```

`LEAST` ignoring nulls covers both cases where there is nothing to be tighter than, which is what the
`COALESCE` was doing in your version: `0` is how either GUC spells "no bound" and `NULLIF` turns it
into a null, and a backend that recognises the parameter without listing it in `pg_settings` returns
one already. Verified on PostgreSQL 18 and CockroachDB v26.2 that it keeps a 3s operator bound and
applies the derived 35s over a 600s one.

Tests: `should keep an operator bound that is tighter than the derived one` and `should apply the
derived bound over an operator bound that is looser`, both starting the pool with `-c` on both GUCs
so whichever one the server has carries the operator's value. That needed `options` on
`DatabaseOptions`; `pg` has always accepted it and pg-boss passes the config straight through, so
this only names it.

## 4. A probe that cannot run failed every batch

Fixed, and the probe moved off the handler's transaction to get there. Catching it where it was would
not have been enough: the probe ran on `tx.db`, so a statement the backend refuses leaves that
transaction aborted, and swallowing the error would hand the handler a transaction answering `25P02`
to everything, which the aborted-transaction message would then blame the handler for. On the pooled
connection a failed probe costs the bound and nothing else.

A failure now emits a `transaction_timeout_probe` warning once per process, runs the batch unbounded,
and leaves the memo empty so the next batch asks again. Event only, like the headroom warning, and
for a sharper reason: it fires because a statement failed, so a write to the `warning` table is the
least likely thing to succeed at that moment. Both the type union and the `docs/api/events.md` row
are updated, including the persistence paragraph that named `transactional_pool_headroom` as the only
exception.

Test: `should run unbounded and probe again after a probe the server refused` replaces the old one.
It asserts both jobs complete, the first with no bound and the second at 35s, two probes, and exactly
one warning.

## Smaller

The GUC assertions read `pg_settings` through a shared `readTransactionBound()` helper, which is
unitless milliseconds on both backends, rather than `current_setting`, which spells the same bound
`35s` on PostgreSQL and `35000` on CockroachDB. That was three tests.

The lost-claim test asserts `[1]` on CockroachDB against `[0, 1]` elsewhere, since the handler's own
`complete()` on a row a peer just moved raises `40001` rather than returning `affected: 0`. The
ledger assertion is unchanged and holds on both, so the guarantee is the same and only the recorded
diagnosis differs. Written up in the CockroachDB section of `docs/database-backends.md` alongside the
heartbeat limit.

timgit:
Thanks for the back and forth with the agents.  We all have our sidekicks working so hard for us, don't we?  :)  Another review posted below.

----

Re-reviewed at `9f2cce1`. All four are addressed, and the two that took a different shape took a
better one: `LEAST` really is the right expression (the `CASE` as written was
`42804 CASE types bigint and text cannot be matched`), and moving the probe off `tx.db` onto the
pooled connection is a stronger fix than catching it where it was, for the reason given. Refusing the
heartbeat pair rather than skipping the refresh is also the right call — `failJobsByHeartbeat`
matching on `heartbeat_on + heartbeat_seconds` is the part that makes the other option worse than the
`40001`.

Numbers: 1230/1230 on PostgreSQL 18; 38/38 transactional on PG18 and under `DISTRIBUTED=true`;
**34 passed / 4 skipped / 0 failed on CockroachDB v26.2.2**, where `84debe3` was 4 failed; 98 on the
settle paths the `settled` change touches; `tsc`, `eslint` and `docs:readme --check` clean.

N1 is pinned to the line that caused it: with everything else on `9f2cce1` and only the reading
reverted to `response.affected < response.requested ? [] : response.jobs`, *"should commit when the
handler settles part of its own batch first"* fails, and passes again with it back. Two side effects
of the fix worth noting as improvements rather than accidents — `settled` carries the ids the
database canonicalised where `jobs` carried whatever the caller passed, so the `Set` lookup no longer
depends on uuid spelling; and the `??` fallback keeps the four settles from drifting apart from the
mutators that share the mapper. The `array_agg` costs nothing measurable: a 10,000-id `complete()`
went 179.2ms → 184.1ms including the uuids on the wire.

The YugabyteDB claim holds and it is the one that decides where the flag belongs: the heartbeat
refresh test passes there, so not setting `noTransactionalHeartbeat` for `yugabytedb` is right.

Two things before merge, both with a patch attached, neither a correctness bug in the feature.

### 1. The apply now has the failure mode the probe was just taught to survive

The probe degrades — warn once, run unbounded, ask again. The apply does not, and the apply is the
half that grew a new way to fail. It went from `SELECT set_config($1, $2, true)`, which touches no
catalog, to a statement that reads `pg_settings` and casts `setting::bigint`, and it runs as the
**first statement of the handler's transaction** — the position `#resolveTransactionTimeoutGuc`'s own
comment explains is the dangerous one.

With a `db` adapter that answers `current_setting` and refuses anything reading `pg_settings` (a
backend or adapter with the GUC but not the catalog view):

```
handler ran: false
job state: failed
output: {"code":"42P01",...,"message":"pg_settings unavailable"}
warnings: []
```

Every batch, and silent as far as the warning system is concerned. That is the last round's finding,
one statement over.

Splitting the read half out and rehearsing it on the pooled connection inside the probe puts it back
on the warn-once path:

```ts
const TRANSACTION_TIMEOUT_TIGHTEST = `LEAST($2::bigint, NULLIF((SELECT s.setting::bigint FROM pg_settings s WHERE s.name = $1), 0))`
const TRANSACTION_TIMEOUT_APPLY = `SELECT set_config($1, ${TRANSACTION_TIMEOUT_TIGHTEST}::text, true)`
const TRANSACTION_TIMEOUT_REHEARSE = `SELECT ${TRANSACTION_TIMEOUT_TIGHTEST}`
```

with `if (guc) await this.db.executeSql(TRANSACTION_TIMEOUT_REHEARSE, [guc, '0'])` chained onto the
probe. The rehearsal is the read only: `set_config` is the part that cannot fail, and leaving it out
means the rehearsal cannot set anything, which matters for a custom `db` adapter whose `executeSql`
may be inside the caller's own open transaction. Sharing the expression is what keeps the two from
drifting. Same reproducer with it in place gives `handler ran: true`, the job completed, and one
`transaction_timeout_probe` warning; 38/38 transactional on PG18, and the existing probe test still
counts its two probes since the rehearsal does not match the string it refuses.

### 2. The two new operator-bound tests cannot run on a server without `transaction_timeout`

```ts
options: `-c transaction_timeout=${ms} -c idle_in_transaction_session_timeout=${ms}`
```

`-c` on a parameter the server does not recognise is refused **at connect**, not at use, so every
connection in the pool fails and the test dies in `PgBoss.start()`. On YugabyteDB 2025.2.3:

```
FAIL  should keep an operator bound that is tighter than the derived one
error: unrecognized configuration parameter "transaction_timeout"
 ❯ Contractor.isInstalled src/contractor.ts:80:20
FAIL  should apply the derived bound over an operator bound that is looser
error: unrecognized configuration parameter "transaction_timeout"
```

YugabyteDB is PostgreSQL 15, so this is not a YugabyteDB quirk — it is every PostgreSQL 13-16, which
is the whole reason the fallback GUC exists. The helper's comment sets both "since which one pg-boss
picks depends on the server", but the server that makes the choice interesting is the one that cannot
be handed the option at all. CI runs neither YugabyteDB nor a pre-17 PostgreSQL, so nothing catches
it.

Asking the same question the code asks, and setting only that GUC, fixes it:

```ts
const withOperatorBound = async (ms: number) => {
  const db = await helper.getDb()
  const { rows } = await db.executeSql("SELECT current_setting('transaction_timeout', true) AS tt")
  await db.close()

  const guc = rows[0].tt !== null ? 'transaction_timeout' : 'idle_in_transaction_session_timeout'

  return { ...ctx.bossConfig, options: `-c ${guc}=${ms}` }
}
```

Both tests then pass on PostgreSQL 18 and on YugabyteDB.

### Smaller

- **The probe retry has no cooldown.** Not remembering the rejection is right, but the memo is
  cleared unconditionally, so a probe that will never be answered is re-issued on every transactional
  batch for the life of the process and, after the first warning, says nothing about it. Remembering
  the failure for a minute rather than for one batch keeps the retry and drops the per-batch cost.

- **The rejection message names the queue for a condition a job can carry.** The `#processJobs`
  assert shares the `work()` string, but it exists precisely for the case where the queue has no
  heartbeat and a job brought its own — which is what the branch's own test sets up. The user is then
  pointed at a queue that is configured correctly, and told to drop a `heartbeatSeconds` that is not
  there. Worth a second string. Related: `updateQueue({ heartbeatSeconds })` on a queue that already
  has a transactional worker starts failing its batches with the same message, which is the backstop
  working as designed but deserves a line in `docs/database-backends.md`.

- **`transaction_timeout_probe` is kept out of the warning table for the wrong reason.** The
  conclusion is right. The stated reason — a statement failed, so a write to the table is unlikely to
  succeed — does not cover the likeliest cause, a backend or adapter that does not understand the
  statement, where the database is healthy and the table writable. What keeps it out is the same
  thing that keeps `transactional_pool_headroom` out: it is not an episode, and a backend that cannot
  answer the probe will not answer it tomorrow either.

- **`HttpCommandResponse = CommandResponse`**, so `settled` — `@internal` in core — became part of the
  proxy's published OpenAPI schema the moment core grew it, and the key-drift typecheck makes
  documenting it the path of least resistance. `9f2cce1` describes what was already going over the
  wire rather than changing it; the open question is whether the proxy should alias the core type or
  pick the fields it publishes.

- **`test/scheduleMissedTest.ts` still carries +43** of timekeeper catch-up tests unrelated to
  transactional workers. Good tests, wrong PR — fourth round.

One note on the CockroachDB numbers: the first runs showed 3-4 failures, all `Contractor.create`
raising `RETRY_SERIALIZABLE` inside `helper.start`. That was two suites sharing one cluster, not the
branch. Run alone the file is clean. Recording it because schema creation is what gives way first on
a contended cluster and it reads like a feature failure.

kibertoad:
Both addressed in 3950a69, plus the smaller items. The two patches landed as written.

## 1. The apply had the failure mode the probe was taught to survive

Fixed as you wrote it. The `LEAST(...)` expression is one shared constant; `TRANSACTION_TIMEOUT_APPLY` wraps it in `set_config` and `TRANSACTION_TIMEOUT_REHEARSE` is the bare `SELECT` of it, run by the probe on the pooled connection with `'0'` for the bound once it has picked a GUC. A refused rehearsal takes the same path as a refused probe: warn once, run unbounded, ask again later. `set_config` stays out of the rehearsal for the reason you gave.

Test: `should run unbounded rather than fail the batch when pg_settings cannot be read`, your reproducer as an assertion. The adapter answers `current_setting` and refuses anything reading `pg_settings`; the handler runs, the job completes with no bound, and exactly one `transaction_timeout_probe` warning carries the adapter's error. With the rehearsal short-circuited to `SELECT 1` it fails on `expected +0 to be 1` (no rehearsal reached the adapter, and the batch would have died on the apply). The existing probe test still counts two probes, since the rehearsal does not contain the string it refuses.

## 2. The operator-bound tests could not run without `transaction_timeout`

Fixed as you wrote it. `withOperatorBound` asks `current_setting('transaction_timeout', true)` on a helper connection and sets only the GUC the server has, so a PostgreSQL 13-16 or YugabyteDB pool no longer fails at connect. The comment explains why `-c` on the unrecognised parameter is the wrong shape rather than a harmless extra.

## Smaller

- **Probe cooldown.** A refused probe is remembered for a minute (`TRANSACTION_TIMEOUT_PROBE_COOLDOWN_MS`); batches inside that window run unbounded without asking, and the first batch after it asks again. The warning stays once per process. Test: `should not repeat a refused probe until the cooldown has passed`, an adapter that never answers, two batches, one probe. Disabling the cooldown check gives `expected 2 to be 1`. The existing retry test sets the new internal `__test__transactionTimeoutProbeCooldownMs: 0` hook so its second batch is the one that re-probes.

- **The batch rejection names the job.** `#processJobs` has its own string now. It says the queue carried no heartbeat when the worker registered, so the heartbeat came from `heartbeatSeconds` on `send()` or from `updateQueue()` after registration, and points at dropping it there. The per-job test asserts the new message. `docs/database-backends.md` names both routes and says each such batch fails before its transaction opens.

- **Why `transaction_timeout_probe` stays out of the warning table.** The reason in `docs/api/events.md` and the field comment is the one you gave: it is not an episode, and a backend or adapter that cannot answer the probe today will not answer it tomorrow. The "a write is the least likely thing to succeed" sentence is gone.

- **`HttpCommandResponse = CommandResponse`.** Left as aliased, deliberately. All four fields of `CommandResponse` are `@internal` in core, including the three the proxy has published since it existed, so `settled` did not cross a line the alias was holding; the proxy has always mirrored whatever the boss call returns, and the schema describes the wire truthfully. Picking the published fields would be a change to the proxy's contract on its own terms, and I would rather not fold that into this PR. Happy to open it separately if you want the proxy to stop publishing the internal shape.

- **`test/scheduleMissedTest.ts`** is back to `master`. The two timekeeper tests are in the branch history (`b9c4fc7`) and will go in a PR of their own.

kibertoad:
@timgit yeah, code review cycles are a bit bizarre now, but can't complain - they definitely catch more stuff than humans ever did before.

PR #892, opened 2026-09-04
[docs] Add screenshots to pg-boss dashboard

jperelli:
I think it is important to show screenshots from a ui package. It is awesome and there is no screenshot anywhere to see!

Thanks for this!


timgit:
Thanks!

PR #896, opened 2026-09-07
feat(schedule): accept RRULE expressions

kibertoad:
Adds RRULE support to `schedule()`, as a first-class expression format rather than a plugin point.

```js
// 5pm on the last Friday of the month, Chicago time
await boss.schedule('report', 'FREQ=MONTHLY;BYDAY=-1FR;BYHOUR=17', null, { tz: 'America/Chicago' })

// or the iCalendar block a calendar exports
await boss.schedule('standup', [
  'DTSTART;TZID=Europe/Berlin:20260901T090000',
  'RRULE:FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR',
  'EXDATE;TZID=Europe/Berlin:20261224T090000'
].join('\n'))
```

Rules cover what cron cannot express: the last Friday of the month, every second Monday, a schedule that stops on a date or after a number of runs. Occurrences are computed by [rrule-temporal](https://www.npmjs.com/package/rrule-temporal) (one dependency, ~440 KB, bundles its own Temporal implementation and needs no polyfill on Node 22+).

## How the format is chosen

An expression carrying a `FREQ=` part, or starting with `DTSTART`, `RRULE`, `RDATE` or `EXDATE`, is read as a rule. Everything else stays a cron expression, which cannot be mistaken for a rule since no cron field contains `=`.

`schedule()` makes that call once and stores the answer in `schedule.kind` (`cron` or `rrule`), which `getSchedules()` now returns. Every reader takes the format from the column, which buys three things: the cron pass stops parsing each expression to find out what it is on every pass, a schedule cannot be validated as one format and evaluated as the other, and a row written straight into the table with SQL can say what it is without anyone guessing.

Schema v41 adds the column with a `CHECK` on the two kinds and a `cron` default. The default labels every row a database already has, which is what all of them are: cron was the only format a schedule could hold.

The `cron` column keeps its name and holds the expression whatever the format. Renaming it would break every consumer reading the table, from the dashboard to a hand-written query.

## Semantics

* **Time zone.** `DTSTART` decides it when it names one. Otherwise the `tz` option does, so `FREQ=DAILY;BYHOUR=9` with `tz: 'America/Chicago'` runs at nine in Chicago, across DST transitions.
* **DTSTART.** A rule that carries none is anchored on 1970-01-01T00:00:00 in the schedule's zone. That anchor is what an `INTERVAL` counts from, so `FREQ=HOURLY;INTERVAL=6` runs at 00:00, 06:00, 12:00 and 18:00 like the equivalent cron expression, and it is the same anchor in every instance and every release (an anchor of "now" would move the phase on every restart). `COUNT` is rejected without a `DTSTART`, since counting from the epoch leaves a rule with nothing left to send, and a schedule that silently does nothing is the one failure a caller cannot see.
* **Finite rules.** `UNTIL` and `COUNT` are honored; once spent, the schedule sits in the table and sends nothing further.
* **Resolution.** Same 60-second due window as cron, so a rule finer than a minute (`FREQ=SECONDLY`, a `BYSECOND` list) sends at most one job a minute, exactly like the 6-placeholder cron format.
* **Validation in `schedule()`.** An unknown part (`BYHOURS=9`), an unknown property (a mistyped `DTSRAT` would otherwise be dropped and leave the rule on the epoch anchor), a second `DTSTART` or `RRULE`, and the combinations RFC 5545 forbids outright, via rrule-temporal's `strict` mode.

## Throttle slot

A rule occurrence need not land on a minute boundary. The forwarded job is therefore filed in the slot the occurrence falls in (`singletonOffset`) rather than the one insert time falls in: two passes on either side of a slot boundary would both find the occurrence inside the due window and send it twice. Cron occurrences keep the slot every release has always filed them in, so a rolling upgrade still collapses them.

## Notes

* Schema v41: one column on `schedule`, no index work, no BAM command, no new constructor option.
* During a rolling upgrade, an instance on an older release reads a stored rule as cron and reports `invalid_schedule` until it is replaced. Documented.
* The dashboard and the proxy declare their own schedule shapes against a published pg-boss, so they pick `kind` up when they bump that dependency and nothing in either breaks without it. The dashboard still labels every schedule by parsing it as cron, so a rule row gets no frequency label. Left alone here.
* `test/rruleTest.ts` covers format detection, zone and anchor semantics, DST, finite rules, every rejection above, the due window under clock skew, the slot pinning across a boundary, and the kind a row is stored with, including the rewrite when an expression is replaced by one of the other format. `test/migrationTest.ts` covers the v41 backfill and the CHECK on a migrated table.


timgit:
Thanks! I think adding a kind column is fine on the table. It would be more pure SQL friendly if needed and might simplify things in core as well.

kibertoad:
@timgit addressed

timgit:
Posting Claude's review below, containing a little bit of overlap from my comments from #888.

---

Claude reviewed this branch at `1aec417` against schema 39 on PG 18, plus CockroachDB v26.2.2 for
the migration. Full suite is 1043/1043, eslint clean, `gen:manifest:check` reports schema v41 up to
date, and the v41 DDL applies on CockroachDB as a background schema change with the existing row
backfilled. Detection held on all 13 cron shapes swept — no cron expression was misread as a rule —
validation was correct on all 32 rule shapes probed, DST walks agree with the cron path, and the
epoch anchor is not a performance trap (`FREQ=MINUTELY` answers in 0.8 ms rather than iterating from
1970). The findings below are what survived verification.

**Finding 1 — an occurrence can be dropped entirely (medium).** `dueOccurrence()` reads a single
occurrence per pass — the earliest one after the window start — so an occurrence is visible only
while no earlier one is still inside the 60-second window. When two occurrences fall closer together
than `cronMonitorIntervalSeconds`, no pass lands in the span and the later one never fires. An
hourly rule with a one-off `RDATE` five seconds before the hour (the shape a calendar export
produces) loses the hourly run at the default 30-second interval at every phase tried:

```
DTSTART:20260908T120000
RRULE:FREQ=HOURLY
RDATE:20260908T125955

interval=30s phase= 0s  slots: 12:59:00        <-- 13:00 hourly run dropped
interval=30s phase=13s  slots: 12:59:00        <-- 13:00 hourly run dropped
interval=15s phase=13s  slots: 12:59:00 13:00:00
```

This is not the documented throttle: the docs promise at most one job a minute, and here a minute
that has an occurrence produces none. Reversing the read direction only moves the loss to the other
occurrence. Suggested fix is to read the window instead of a point, since rrule-temporal has
`between()`:

```ts
const occurrences = rule.between(new Date(databaseTime - OCCURRENCE_WINDOW_SECONDS * 1000), new Date(databaseTime))
```

then send one job per distinct throttle slot in the list. One call, bounded at 60 seconds, and it
makes the documented contract exact. Worth a regression test for two occurrences in one window;
`rruleTest` covers slot pinning across a boundary but not this.

**Finding A — an older instance leaves a stale kind behind (medium).** `plans.schedule()`'s upsert
sets `kind` from `EXCLUDED`, but the statement a 12.30.x instance runs does not name the column, so
it rewrites `cron` and leaves `kind` at whatever a newer instance last wrote. Replaying the exact
pre-`kind` statement against a real v41 schema:

```
after new-instance schedule(): [ { kind: 'rrule', cron: 'FREQ=MINUTELY' } ]
after old-instance schedule(): [ { kind: 'rrule', cron: '* * * * *' } ]
pass result: sent = 0 | warnings = [ '... rrule part "* * * * *" has no value' ]
```

That is the ordinary rolling-upgrade shape — apps call `schedule()` on boot — and the row stays
broken for old and new instances alike until `schedule()` is called again from a new instance, which
for most deployments is the next boot rather than the next pass. An application rollback without a
schema rollback does the same. The head before `5f4bc3e` could not reach this state, because the
format came from the expression and every row was read the way it was written.

**Finding B — a v41 rollback and re-upgrade relabels every rule as cron (medium-low).** The v41
`uninstall` drops the column and nothing rebuilds it, so the re-run install labels every row `cron`
from the default:

```
on v41:                  [ { kind: 'rrule', cron: 'FREQ=MINUTELY' } ]
after rollback to 40:    [ { cron: 'FREQ=MINUTELY' } ]
after re-upgrade to 41:  [ { kind: 'cron', cron: 'FREQ=MINUTELY' } ]
```

Both A and B stop a schedule silently rather than firing it at a wrong time, which is the right
direction to fail in, and `warnedSchedules` keys the warning so it is emitted once with the queue and
key named. But the row survives looking valid and never fires again. Two changes close the class:

1. Treat the column as a hint, not a verdict. On the error path only — where the pass already
   catches — re-run `isRrule()`, and if detection disagrees with the stored kind, evaluate it the
   detected way and optionally repair the row. One regex on a path that is already exceptional.
2. Fill the column from the expression in the migration rather than from a blanket default:

```sql
UPDATE ${schema}.schedule SET kind = 'rrule'
 WHERE cron ~* '(^|[[:space:]]|;)FREQ=' OR cron ~* '^[ \t]*(DTSTART|RRULE|RDATE|EXDATE)[;:]'
```

No cron expression matches either pattern, it costs one scan of a small table, and it also repairs
any row a v40 instance wrote during the upgrade window — Finding A's other half.

Worth noting the column is not load-bearing for performance: `isRrule` measures at 55 ns, so 0.055 ms
per pass at 1000 schedules against a pass that runs every 30 seconds. What it actually buys is a
labeled format in `getSchedules()` and the ability for a hand-written row to declare itself. Both are
real, and reverting `5f4bc3e` — which would take the schema bump, the #888 collision and Findings A
and B with it — is a legitimate alternative to fixing them.

**Finding 2 — `singletonSlot` is now honored on the public `insert()` path (low).** `insertJobs`
declares `"singletonSlot" text` in the `json_to_recordset` column list, and `insert()` stringifies
caller objects straight into it, so an undeclared, undocumented field is live from user input and a
bad value surfaces as a raw Postgres error rather than a pg-boss one:

```
insert with garbage singletonSlot: THROW invalid input syntax for type timestamp: "not-a-timestamp"
```

On master the field was silently ignored. Prefixing it (`__singletonSlot`) or dropping it from the
recordset on the `insert()` path is enough.

Lower severity, none blocking:

- `cachedRule` clears the whole map at `CACHE_MAX` instead of evicting one, so past 1000 distinct
  `(tz, expression)` pairs the hit rate settles near zero rather than "one pass rebuilding" as the
  comment says. A miss is cheap (build 0.04–0.06 ms against evaluate 0.11–0.28 ms), so an LRU or a
  reworded comment both work. Entries for deleted schedules are never evicted until the cap trips.
- `packages/proxy` restates the kind set as `z.enum(['cron','rrule'])` instead of deriving it from
  `plans.SCHEDULE_KINDS`, so a third kind would typecheck on both ends and be dropped at the
  boundary. Its `"pg-boss": "^12.30.0"` also needs `^12.31.0` now that `HttpSchedule` requires
  `kind`.
- The cockroachdb/yugabytedb timeout move from 60s to 120s is a good refactor — one place, and the
  "only lift, never cap" rule stated where both readers see it — but it is unrelated to RRULE and
  changes how long a genuinely hung distributed test takes to report.
- `rrule-temporal` is worth pinning tighter than a caret for the first release, given how much
  correctness rides on it and that the DST-gap fix in 2.2.4 is recent. This takes pg-boss from three
  runtime dependencies to four.
- `assertRrule` checks "has an occurrence left" against `new Date()` while the pass evaluates against
  database time. Only matters for a rule expiring inside the skew window.
- The dashboard degrades quietly rather than badly: `cronHuman()` renders `Custom schedule` and the
  next-fire column renders `—`, nothing throws. Now that the row carries `kind`, the schedules page
  can at least label a rule as a rule. Follow-up, not this PR.
- The docs' rolling-upgrade paragraph is accurate about an older instance reading a rule as cron and
  warning, but does not mention that an older instance's `schedule()` can leave a stale `kind` and
  break the row for new instances too. If Finding A is fixed in code this stands as written.

On sequencing: #888 also claims schema v41 and adds a column to the same table, so whichever of the
two lands second on its own has to move to v42. Combining #896, #887 and #888 as Tim proposed on #888
puts `kind` and `lastJobId` in a single v41 migration and removes the collision entirely.

This PR supersedes #886 — same request, a fraction of the code, no registration API and no second
`schedule()` signature — and all three of #886's reproduced defects were properties of its
persisted-`next_run_at` design, so they do not appear here. The one thing worth carrying forward from
it is the `missed: 'skip' | 'once' | 'all'` catch-up policy, which is separable and can land on its
own merits.

kibertoad:
Findings addressed in `d1f5ae1`. The suite is 953/953 on the branch, eslint clean, `gen:manifest:check` reports v41 up to date.

**Finding 1, an occurrence can be dropped.** Fixed with `between()`, as suggested. The pass reads the whole 60-second window and sends one job per distinct throttle slot, so the documented resolution is now exact: one job for each minute that holds an occurrence, and two occurrences in separate minutes both send however close together they are. `rruleTest` gained two cases: the hourly-plus-`RDATE` shape from the report, which now forwards both occurrences in their own slots, and a per-15-second rule, whose four occurrences in one window collapse to one job per minute rather than four.

**Finding A, a stale kind, and Finding B, a rollback relabelling every rule.** Both closed the two ways you laid out. The column is a hint rather than a verdict: when an expression cannot be read the way the column says, and is written the other way, the pass reads it the way it is written, and it also corrects the column, so the row leaves the fallback path and `getSchedules()` stops reporting a format the expression is not in. The v41 migration fills the column from the expression rather than leaning on the default alone, with your two patterns, written so postgres and CockroachDB's regexp engine read them the same way (`^` anchors the whole string in one and not the other, so a property on a line below the first is matched on the whitespace before it). `migrationTest` covers a v40 table holding a bare rule and an iCalendar block; `rruleTest` covers the fallback and the repair in both directions, and that an expression broken in both formats still warns and relabels nothing.

**Finding 2, `singletonSlot` on the public path.** Prefixed to `__singletonSlot`, so `insert()` does what it did before rules existed: ignores the name. Covered by a test that inserts a garbage `singletonSlot` and gets a job with a null `singleton_on` instead of a postgres error.

**The rest**

- The cache is a `toad-cache` `LruMap` now, so the cap evicts what nobody evaluates rather than clearing the map. Comment reworded to match.
- `assertRrule` takes the pass's reading of the database clock.
- `rrule-temporal` pinned to `~2.2.4`.
- `SCHEDULE_KINDS` is exported as `scheduleKinds`, beside `states` and `policies`. The proxy has to derive it at its next dependency bump rather than here: it builds against a published pg-boss (12.28.0 in its lockfile today), so `^12.31.0` and a runtime import of the constant would both fail `npm install`/typecheck until 12.31.0 is on the registry. Every previous proxy bump landed in a release commit for the same reason.
- Dashboard label: agreed, follow-up.
- The distributed timeout move is unrelated, as you say. It stays on the branch since it is one place instead of two, but it is easy to drop if you would rather it went in on its own.

**Sequencing.** Combined as you proposed on #888: **#898** carries #887, #888 and #896 as three commits with `kind` and `last_job_id` in a single v41 migration, so the collision is gone. That PR also does the integration the combination needs, mainly `previewSchedule()` reading either format so a stored rule can be previewed from its `cron` column without consulting its `kind`, with a finite rule answering with what is left of it and a spent one with nothing.

kibertoad:
Folded into #898, which carries the RRULE work as its third commit alongside #887 and #888, with `kind` and `last_job_id` in one v41 migration. Closing in favour of that one; nothing here is dropped.

The review findings are addressed in `d1f5ae1`, which is in #898: the pass reads the whole due window rather than a point, the `kind` column is a hint the pass corrects rather than a verdict, the migration fills it from the expression, `singletonSlot` is prefixed off the public `insert()` path, and the rule cache is an LRU. Details in the [reply above](https://github.com/timgit/pg-boss/pull/896#issuecomment-5590661488).

PR #898, opened 2026-09-08
feat(schedule): RRULE expressions, lastJobId, getSchedule() and previewSchedule()

kibertoad:
Replaces #887, #888 and #896, now closed in favour of this one, and supersedes #886.

Combined as proposed on #888. All three land on the `schedule` table, and #888 and #896 each claimed schema v41 on their own, so whichever went second would have had to move to v42 and every database would have paid a second pass over the table. Here `kind` and `last_job_id` arrive in one v41 migration and the collision is gone.

Four feature commits, then the review fixups. The first three are one per PR this replaces, in that order: the first two are those branches as they stood, and the third carries the RRULE work, the #896 review findings, and the integration the combination needs. The fourth is the one piece of #886 worth keeping, the catch-up policy.

## What lands

**`getSchedule(name, key)` and `previewSchedule(cron, options)`** (#887)

```js
const schedule = await boss.getSchedule('report', 'eu')
const upcoming = boss.previewSchedule(schedule.cron, { tz: schedule.timezone })
```

`(name, key)` is the primary key of the `schedule` table, so `getSchedules(name, key)` could only ever return zero or one row while still returning an array. `previewSchedule()` answers "when does this next fire?" with the validation `schedule()` performs and the clock the cron pass reads, rather than leaving the calling code to reach for cron-parser and get both slightly wrong. No schema change.

**`lastJobId`** (#888)

```js
if (schedule.lastJobId) {
  const [job] = await boss.findJobs('report', { id: schedule.lastJobId })
}
```

A schedule row said what should happen and when, but nothing about what did. `last_job_id` points a schedule at the job it most recently created, so an operator can reach the last run without correlating on payload or timestamp. Written from the send-it handler, one statement per batch, best effort, with the caveats documented: `null` means "no run recorded", not "never fired".

**RRULE expressions** (#896)

```js
// 5pm on the last Friday of the month, Chicago time
await boss.schedule('report', 'FREQ=MONTHLY;BYDAY=-1FR;BYHOUR=17', null, { tz: 'America/Chicago' })

// or the iCalendar block a calendar exports
await boss.schedule('standup', [
  'DTSTART;TZID=Europe/Berlin:20260901T090000',
  'RRULE:FREQ=WEEKLY;BYDAY=MO,TU,WE,TH,FR',
  'EXDATE;TZID=Europe/Berlin:20261224T090000'
].join('\n'))
```

Rules cover what cron cannot express: the last Friday of the month, every second Monday, a schedule that stops on a date or after a number of runs. Occurrences come from [rrule-temporal](https://www.npmjs.com/package/rrule-temporal), which bundles its own Temporal implementation and needs no polyfill on Node 22+.

An expression carrying a `FREQ=` part, or a line that opens with `DTSTART`, `RRULE`, `RDATE` or `EXDATE`, is read as a rule. Everything else stays a cron expression, which cannot be mistaken for a rule since no cron field contains `=`, `:` or `;`. `schedule()` makes that call once and stores the answer in `schedule.kind`, which `getSchedules()` returns.

**Catch-up policy** (#886)

```js
// a nightly report reads the current state of the world, so three missed nights are one report
await boss.schedule('report', '0 3 * * *', null, { missed: 'once' })
```

A pass sends the occurrences of the preceding 60 seconds and nothing else, so a deployment that was down, or between deploys, for an hour never sent what that hour held. `missed` decides what a schedule owes for that: `skip` is today's behavior and stays the default, and `once` sends a single job for the most recent missed occurrence, however many were missed.

Those two are the whole option. A caught-up job carries the schedule's `data` and nothing else, so a policy sending one job per missed occurrence would produce twelve identical jobs after a twelve-hour outage with no way for a handler to tell which hour each was for. `once` needs no such identity, and it is bounded by construction rather than by a cap.

The gap comes off `version.cron_on`, which every pass already advances. The claim answers with the timestamp it replaced now, so the pass knows how long scheduling was off without a `next_run_at` persisted per schedule, which is where each of the defects on #886 came from. No new column and no new constructor option: the policy rides in the `options` blob the row already carries.

Each schedule's range is bounded below by its own `created_on`, so a schedule written during the gap does not start life owing occurrences of an expression that was not in the table yet. Not by `updated_on`, which a deployment calling `schedule()` on every boot rewrites on the way up, and which would leave the policy with nothing to catch up on in precisely the case it exists for.

A missed occurrence names the throttle slot it falls in, the mechanism a rule occurrence already used, so the catch-up job dedups across instances and cannot collide with the occurrence that is due now. The occurrence is read backwards from the window, so what the read costs follows the one job it sends rather than the length of the outage.

`lastJobId` resolves by slot as a result: a catch-up job and a due job can reach the send-it handler in one batch, in an order the fetch does not promise, so keeping whichever settled last named an arbitrary run rather than the later one.

## Schema v41

One migration, two columns on `schedule`, no index work, no BAM command, no new constructor option:

```sql
ALTER TABLE schedule ADD COLUMN IF NOT EXISTS kind text NOT NULL DEFAULT 'cron' CHECK (kind IN ('cron','rrule'));
UPDATE schedule SET kind = 'rrule' WHERE kind = 'cron' AND (cron ~* '(^|[[:space:]]|;)FREQ=' OR cron ~* '(^|[[:space:]])(DTSTART|RRULE|RDATE|EXDATE)[;:]');
UPDATE schedule SET timezone = 'UTC' WHERE timezone IS NULL;
ALTER TABLE schedule ALTER COLUMN timezone SET DEFAULT 'UTC';
ALTER TABLE schedule ADD COLUMN IF NOT EXISTS last_job_id uuid;
```

The kind `UPDATE` is what puts the label back on a table that has been rolled back to v40 and re-upgraded, since `uninstall` drops the column rather than remembering it. The two patterns are the detection `isRrule()` performs, written in the terms postgres and CockroachDB's regexp engines share, and no cron expression matches either.

CockroachDB does not run that statement. It executes `ADD COLUMN` as a schema-change job and refuses a write to the column inside the transaction that added it, and the migration is one transaction, so an existing database failed the upgrade there with 42P10, `column "kind" is being backfilled`. The `noAddColumnBackfill` compatibility flag leaves the backfill out on a backend that cannot take it, and the label falls to the pass: a row whose kind disagrees with its expression is read the way it is written and relabelled on the first pass that reaches it. What it costs is the window before that pass, where `getSchedules()` reports `cron` for a rule row. A fresh v41 install was never affected, and YugabyteDB migrates with the statement in place.

The zone `UPDATE` retires an accident. `schedule({ tz: null })` passed on releases whose destructuring default only covered `undefined`, and a row written with SQL can leave the column out; cron-parser reads a non-string zone as unset, so those rows have been firing in the local zone of whichever instance took the pass. Nothing chose that: `schedule()`'s default is UTC and the docs say UTC. It rides this migration because the same rows are already being read, and it is what makes `Schedule.timezone: string` true rather than a claim about a nullable column. The column default follows it, so the next hand-written insert cannot make another null, and it matches what a fresh install builds.

The `cron` column keeps its name and holds the expression whatever the format. Renaming it would break every consumer reading the table.

The catch-up policy needs no schema of its own. It is read by the pass, which reads every row anyway, and nothing queries the table by policy, so it sits in the `options` blob beside `tz` and `key`.

## Integration the combination needs

`previewSchedule()` reads either format, told apart the same way `schedule()` tells them apart, so `previewSchedule(schedule.cron, { tz: schedule.timezone })` works for a rule row without the caller consulting `kind`. A finite rule answers with fewer occurrences than asked for, and one whose last occurrence has passed answers with none, which is how a caller learns a schedule has finished. `schedule()` still refuses to store a rule with nothing left to send, so that check lives apart from the readability checks the two share.

`schedule()`'s validation is one dispatch: `assertRecurrence()` reads an expression in whichever format it is written, and both formats report an unusable time zone in the same words, through the same `assertTimezone()`.

The send-it payload carries the schedule's `key` and an injective singleton key (#888) on both cron and rule occurrences, and a rule occurrence additionally names the throttle slot it belongs in (#896).

## The #896 review, addressed

- **An occurrence could be dropped entirely.** The pass read one occurrence per schedule, so an occurrence was visible only while no earlier one was still inside the 60-second window. An hourly rule with an `RDATE` five seconds before the hour lost the hourly run at the default interval. The pass now reads the window with `between()` and sends one job per distinct throttle slot, which makes the documented resolution exact: one job for each minute that holds an occurrence, and two occurrences in separate minutes both send.
- **A stale `kind` left a row that never fired.** A 12.30.x instance's `schedule()` does not name the column, so an upsert from one during a rolling upgrade replaced the expression and left the kind a newer instance had written; a v41 rollback and re-upgrade relabelled every rule cron from the default. The column is now a hint rather than a verdict: when an expression cannot be read the way the column says, and is written the other way, the pass reads it the way it is written and corrects the column. The migration's `UPDATE` closes the other half.
- **`singletonSlot` was live on the public `insert()` path.** `insert()` stringifies caller objects straight into `insertJobs`' recordset, so an undocumented field took user input and a bad value surfaced as a raw postgres timestamp error. The column is now declared only in the statement the cron pass asks for, and the field is stripped from the objects `insert()` is handed, so `insert()` does under either spelling what it did before: ignores the name.
- **The rule cache is gone.** Keeping built rules between passes saved a rule schedule about 0.1ms a pass and cost every install a dependency for it, so a rule is built per read. `previewSchedule()` is the one caller it earned anything for, because the rrule branch of its walk rebuilt the rule for every occurrence; that walk builds it once and closes over it now, the way the cron branch beside it keeps one parsed interval.
- **`assertRrule` judged liveness against the local clock** while the pass evaluates against database time. It takes the pass's reading now.
- **`rrule-temporal` is pinned** to `~2.2.4` for the first release.
- `SCHEDULE_KINDS` is exported as `scheduleKinds`, beside `states` and `policies`, so a consumer declaring its own schedule shape can derive the kind set instead of restating it. The proxy will do that at its next dependency bump: it builds against a published pg-boss, and the constant is only public from 12.31.0, so its `z.enum(['cron','rrule'])` and its `"pg-boss": "^12.30.0"` both move with that bump rather than here, where the version does not exist on the registry yet.

Not addressed, deliberately: the dashboard still labels every schedule by parsing it as cron, so a rule row gets no frequency label. It degrades quietly, rendering `Custom schedule` and an empty next-fire column rather than throwing, and now that the row carries `kind` the page can label a rule as a rule. Follow-up.

## The #898 review, addressed

- **`missed: 'all'` is gone.** A caught-up job carries the schedule's `data` and nothing else, so the policy could not name the occurrence each of its jobs stood for, which voids the hourly-export case it was documented with. `skip` and `once` remain, `SCHEDULE_MISSED_POLICIES` and the proxy enum shrink with them, and the docs section is written around `once` rather than trimmed. What a per-occurrence policy would need first is real work: one transaction around the per-queue inserts, `singleton_on` as the occurrence identity through `includeMetadata`, and an answer for a schedule whose own send options set `singletonKey`. Follow-up.
- **The backwards read throws on a rule that is empty near the window and dense behind it.** Widening the step is a guess about density, and that shape defeats it: the step grows until it spans more occurrences than rrule-temporal generates in one call. A per-minute rule whose `UNTIL` passed a few days before the gap opened is that shape, and it lost the whole catch-up to a warning. Limit-independent, so dropping `all` did not dispose of it. A step that overruns now halves and is read again, with the width it overran at as a ceiling and an hour as the floor. The floor is safe by construction: nothing recurs more often than once a second, so an hour holds at most 3600. `test/rruleTest.ts` covers minutely and per-second rules spent an hour, a day, three days and twenty days before the window, and `test/scheduleMissedTest.ts` covers the shape end to end through a pass.
- **`previous()` is gone**, and with it the RDATE bug: it answers with an `RDATE` in place of the rule occurrence that follows it and loses everything in between. Both directions read `between()` now, and the two agree on all ten RDATE/EXDATE shapes.
- **The singleton key escapes the queue name's underscores** rather than moving to a JSON pair. Injective, since neither half can hold a backslash, and byte-identical for every name that carries no underscore. It does move the key for a name that has one, whether or not that name collides today, so the rolling upgrade double-fires those schedules once. That is in the notes below.
- **`schedule({ tz: null })` threw**, and took the documented `previewSchedule()` recipe with it. A falsy zone is "none given" in both methods now, which is strictly more permissive than 12.30.0, and it closes the `tz: 0` case that used to store `'0'` and then throw on every pass. The migration backfill and a `COALESCE` on the schedule read cover the rows already written and whatever an older instance writes mid-rollout. The column now also carries `DEFAULT 'UTC'`, on a fresh install and after the migration, so a hand-written insert cannot make another one. It stays nullable: `NOT NULL` would make an old instance's `schedule({ tz: null })` fail against the migrated schema during a rollout, which is a louder failure than the one being fixed.
- **`__singletonSlot` was still live on `insert()`.** Closed rather than restated: the column is in the statement the pass asks for and nowhere else.
- **The relabel `UPDATE` matches on the expression it judged**, so an upsert landing between the pass's read and its write keeps the kind it was stored with.
- **The proxy mirrors the `count` ceiling**, so 1001 is a 400 naming the parameter and the generated OpenAPI parameter carries the maximum. The core exports the ceiling as `previewScheduleMaxCount`; the proxy's literal moves to it at the same dependency bump as `scheduleKinds` and `scheduleMissedPolicies`.
- **The schedule tests close the pools they open**, and the `updated_on` guarantee is pinned on both write paths.
- **A nullish `missed` reads as the default**, which is where `tz` lands.
- **The split `try` keeps its comment honest.** Both reasons given for it have now been retired: cron-parser does not give up walking a sparse expression, it gets slow, and the narrowing means an expression rrule-temporal refuses at an hour it refuses over a minute too, which the window read catches first and skips the row for. The `try` stays split, because coupling them was a real defect once and two separate calls over different ranges have no reason to share one. The comment says that instead of naming a divergence I cannot reproduce.
- **The v41 migration could not run on CockroachDB.** `ADD COLUMN kind` and the `UPDATE` that labels a stored rule share the migration's one transaction, and the engine refuses a write to a column it is still backfilling. The backfill is dropped there behind `noAddColumnBackfill`, and the relabel the pass already performs puts the label back on the first read of the row. Verified on a three-node cluster through pg-boss's own migration path.
- **`toad-cache` is gone**, and the module-level rule cache with it, for the reasons above.
- **The backwards read says what narrowing costs.** With the ceiling at the floor it walks an hour of range per step, so it costs what it searches rather than what it returns: `FREQ=SECONDLY;BYMONTHDAY=1` answers in around 1.3 seconds inside the pass, bounded by the distance back to the nearest occurrence rather than by the width of the gap, so 30 days and 90 cost the same.
- **`onSendIt`'s comment no longer describes the `all` batch.** A pass files at most two jobs for one schedule, the catch-up and the one due, which is what the test below it pins.

## Notes

- The v41 zone backfill changes behavior for existing rows: a schedule whose `timezone` is NULL has been firing in the host's local zone and fires in UTC after the upgrade. No-op on a UTC host, a shift by the offset otherwise. Worth a release-note line.
- The singleton key the cron pass files a forwarded occurrence under escapes the queue name's underscores, so it changes for every schedule whose queue name contains one, not only for the pairs that collide today. During the rolling upgrade to this release, one occurrence of such a schedule can be sent by an old pass and a new pass under their two different keys and produce two jobs. Bounded to the rollout window, and a queue name with no underscore is unaffected. Worth a release-note line.
- During a rolling upgrade, an instance on an older release reads a stored rule as cron and reports `invalid_schedule` until it is replaced. Documented.
- `getSchedules()` previously ran `SELECT *`, so `created_on` and `updated_on` reached callers in snake_case. The schedule reads now use an explicit aliased column list and return `createdOn`, `updatedOn`, `kind` and `lastJobId`. Code reading `schedule.created_on` or `schedule.updated_on` needs the camelCase name; neither was in the `Schedule` type before.
- The per-backend test timeout moved into `test/timeouts.ts`, so `vitest.config.ts` and the suites that raise the default for a block read one number. It came in on #896 and is unrelated to scheduling, but it is one place instead of two and states the "only lift, never cap" rule where both readers see it.
- On CockroachDB the v41 migration leaves existing rule rows labelled `cron`, so `getSchedules()` reports the wrong format for one until the first pass reads it and relabels. Every other backend labels them in the migration. Worth a release-note line.
- Verified on PostgreSQL 18: full suite 1136 passed / 0 failed, eslint clean, `tsc` clean, proxy typecheck and test suite clean (206 passed), `gen:manifest:check` reports schema v41 up to date. On a three-node CockroachDB cluster: `distributedDatabaseTest` 13 passed / 1 skipped, the rrule, schedule-read and missed-occurrence suites 81 passed run one file at a time, and the v41 migration checked by hand through install at 41, rollback to 40 and migrate. With vitest's file parallelism on, three schedule tests fail inside `Contractor.create` with `RETRY_SERIALIZABLE`, which is schema creation contending with itself on a single-machine cluster rather than anything scheduling does. YugabyteDB was not run locally.
- `test/rruleTest.ts` covers format detection, zone and anchor semantics, DST, finite rules, every rejection, the due window under clock skew, two occurrences inside one window, the slot pinning across a boundary, the backwards read against the window read on ten RDATE/EXDATE shapes, the backwards read over a stretch the rule is empty across, the stored kind and its repair, and the unprefixed slot staying ignored on `insert()`. `test/scheduleMissedTest.ts` covers each policy in both formats, a rule that stopped recurring partway through the gap, the steady state where a pass leaves no gap, the `created_on` bound, a policy the pass does not recognize, a missed occurrence sharing a slot with a due one, the claim's prior timestamp, and an outage caught up end to end. `test/scheduleReadTest.ts` covers `getSchedule` and `previewSchedule` in both formats, a falsy zone, and a legacy row whose zone was never stored. `test/scheduleLastJobTest.ts` covers `lastJobId`. `test/migrationTest.ts` covers the v41 backfill from the expression and the `CHECK` on a migrated table.

#886 asked for the same thing and is closed as well. This is a fraction of the code: no registration API, no second `schedule()` signature, and no persisted `next_run_at`, which is where each of the defects reproduced on that PR came from. The one piece worth carrying forward from it is the `missed: 'skip' | 'once'` catch-up policy, which is here, built on the pass timestamp the version row already carries rather than on per-schedule state.


kibertoad:
@timgit Here is the aggregated PR as requested

timgit:
Posting a slightly cleaned up claude review below with some findings.  

---

**1. `missed: 'all'` silently drops jobs for any rule carrying an `RDATE` (blocking).**

The catch-up walk reads backwards through rrule-temporal's `previous()`, and `previous()` disagrees with the `between()` the due window uses. Reproduced against rrule-temporal directly, no pg-boss code involved:

```
DTSTART:20260909T000000Z
RRULE:FREQ=HOURLY
RDATE:20260909T103012Z

previous(2026-09-09T12:00:00Z, true)  -> 2026-09-09T12:00:00Z
previous(2026-09-09T12:00:00Z, false) -> 2026-09-09T10:30:12Z     <- 11:00:00Z skipped
between(10:00Z, 12:00Z+1ms)           -> 10:30:12Z, 11:00:00Z, 12:00:00Z
```

`previous()` jumps to the RDATE and resumes from there, so every regular occurrence between the cursor and the RDATE is lost. Severity scales with how far back the RDATE sits — over a 4-hour range:

```
no RDATE                between=4  previous=4  agree
1 RDATE mid-range       between=5  previous=4  loses 11:00
1 RDATE on the hour     between=5  previous=4  loses 11:00
2 RDATEs                between=6  previous=5  loses 11:00
RDATE earlier than the range  between=4  previous=1  loses 09:00, 10:00, 11:00
RDATE after the last occurrence / EXDATE only    agree
```

The bug is upstream in rrule-temporal, and there is now a fix open for it — [ggaabe/rrule-temporal#138](https://github.com/ggaabe/rrule-temporal/pull/138). **But it needs fixing here too, and I would treat that as blocking rather than as a follow-up**, for three reasons:

- `rrule-temporal` is pinned `~2.2.4`, so even once that merges, a fix released as 2.3.0 will not be picked up. Waiting on upstream is not actually an available option; the choice is to fix it in this PR or to widen the pin later and gate the release on someone else's cadence.
- This is the shape the feature advertises. `docs/api/scheduling.md` invites callers to paste "the `DTSTART` line, the `RRULE` line, and optional `RDATE` and `EXDATE` lines" of a calendar entry, and the resolution section's worked example is an `RDATE` sitting next to an hourly rule's own occurrence. Calendar exports carry RDATEs routinely.
- It is the same class of defect as the dropped occurrence the window read was introduced to close — a lost occurrence, reached through a different door. That one was worth fixing before merge; this one reads the same.

### The remedy

Swap the `previous()` walk in `occurrencesBefore()` for a `between()` read in chunks that widen backwards from `until`. It needs nothing from upstream, and it is not a trade — it is faster than what is there now as well as correct.

```js
let upper = until
let width = CHUNK_SECONDS * 1000                     // an hour, growing x4 a chunk

while (occurrences.length < limit && upper.getTime() > after.getTime()) {
  const lower = new Date(Math.max(after.getTime(), upper.getTime() - width))
  const chunk = rule.between(lower, new Date(upper.getTime() + 1))

  for (let i = chunk.length - 1; i >= 0 && occurrences.length < limit; i--) {
    occurrences.push(toDate(chunk[i]))
  }

  upper = lower
  width *= CHUNK_GROWTH
}
```

Chunk *k* covers `(lower_k, upper_k]` and the next one's upper bound is `lower_k`, so every occurrence is read exactly once, `until` stays included, and `after` stays excluded — the same bounds the current signature promises. A dense expression fills the cap out of the first chunk; a sparse one reaches `after` in a handful of steps.

I applied it and measured. All eight RDATE/EXDATE shapes now agree with `between()` over the same range, and the cost drops by one to two orders of magnitude, because `previous()` was rebuilding a phase-aligned rule clone on every one of the 1000 steps while `between()` generates a run in a single pass:

| expression | gap | returned | `previous()` walk | chunked `between()` |
|---|---|---|---|---|
| `FREQ=MINUTELY` | 24h | 1001 | 337ms | **2ms** |
| `FREQ=SECONDLY` | 1h | 1001 | 217ms | **7ms** |
| `FREQ=HOURLY` | 90d | 1001 | 225ms | **2ms** |
| `FREQ=DAILY;BYHOUR=9` | 400d | 400 | 145ms | **4ms** |
| `FREQ=MONTHLY;BYDAY=-1FR;BYHOUR=17` | 5y | 60 | 45ms | **1ms** |

Full suite stays **1127/1127** with it applied, and all five schedule suites pass unchanged — it is a drop-in replacement for that one function. That also retires Finding 10 below: a capped catch-up stops being a fifth of a second of blocked event loop per schedule.

The patch is at the bottom of this comment if it saves you the retyping.

Severity does split by policy, for what it is worth: `all` loses jobs outright, while `once` still sends its single job but names the wrong occurrence, so it is wrong rather than lossy. `skip` and the due window are unaffected — the window read uses `between()`.

Related, and the reason it is worth fixing rather than documenting: `missedOccurrences()` shares a `try` with `dueOccurrences()`, so if the backwards walk ever throws where the window read would not, the `continue` drops that schedule's **due** job too, not just its catch-up. The comment argues the two cannot diverge because they read the same expression — the table above is a case where they already read it differently.

**2. Every cron schedule fires twice during the rolling upgrade (medium).** The send-it singleton key moves from `` `${name}__${key}` `` to `JSON.stringify([name, key])`.

The change itself is right, and I confirmed the collision it fixes is real rather than theoretical. `_` is legal in both halves — the queue name and the schedule key go through the same `assertObjectName` charset `[\w.\-/]+` (`src/attorney.ts:88`, `assertKey` at `:620`) — so the concatenation isn't injective:

```
old format  `${name}__${key}`:
  ("report_", "daily")  -> "report___daily"
  ("report",  "_daily") -> "report___daily"     <- same string

new format  JSON.stringify([name, key]):
  ("report_", "daily")  -> ["report_","daily"]
  ("report",  "_daily") -> ["report","_daily"]
```

And it silently drops a job: the forward carries `singletonSeconds: 60` so both land in the same `singleton_on` slot, `job_i4` is `UNIQUE (name, singleton_on, COALESCE(singleton_key, ''))` (`src/plans.ts:772`) and `insertJobs` ends in `ON CONFLICT DO NOTHING` (`:2088`). Two distinct schedules, one occurrence each, same minute:

```
send-it jobs that landed for the two schedules in one minute:
  old underscore format: 1 of 2
  new JSON format:       2 of 2
```

The two schedules are on different queues, so the damage is one queue's scheduled job never being sent, every minute the two coincide, with nothing emitted to say so. `JSON.stringify` over the pair is injective and closes it.

What the fix costs is the transition. The two formats don't collide with *each other*, so during the rollout an old-version pass and a new-version pass land two jobs for one occurrence. Reproduced — same slot, both survive `ON CONFLICT DO NOTHING`:

```
old key: report__ | new key: ["report",""]
send-it jobs queued for one occurrence: 2
   ["report",""]  2026-09-09T19:28:00.000Z
   report__       2026-09-09T19:28:00.000Z
```

Not a narrow window either: `cronMonitorIntervalSeconds` is validated to 1–45 with a default of 30, against a 60s occurrence window, so at least two passes always land inside a window — a rollout only has to put one of them on each version. This affects every existing cron schedule in every multi-instance deployment. It's acknowledged in a comment in `timekeeper.ts` but nowhere a user will read it.

Escaping the separator keeps injectivity and avoids the transition, since it leaves the key byte-identical for every schedule whose name and key contain no `_` — only the pairs that are already broken today change:

```js
const esc = (s) => s.replaceAll('_', '\\_')
singletonKey: `${esc(name)}__${key}`
```

(`\` is outside the charset above, so the escape is injective.) Or keep the JSON form and give it a release-note line plus a sentence in the scheduling docs — the duplicate is one occurrence per schedule, bounded to the rollout window, which may well be the right trade.

**3. `schedule({ tz: null })` now throws, and it takes the documented `previewSchedule` recipe with it (medium).** These were two findings; they have one cause and one fix.

`assertTimezone()` gained an explicit `typeof tz !== 'string'` early-throw (`src/timezone.ts`), and both `schedule()` and `previewSchedule()` inherit it. The `{ tz = 'UTC' }` destructuring default only covers `undefined`, so a nullish value threaded out of config or a database column now throws where 12.30.0 accepted it:

```
schedule tz=null      => THROW: Unknown or unsupported time zone: null     (12.30.0: ACCEPTED)
schedule tz=0         => THROW: Unknown or unsupported time zone: 0        (12.30.0: ACCEPTED)
schedule tz=undefined => ACCEPTED
schedule tz=""        => THROW  (unchanged from 12.30.0)
schedule tz={}        => THROW  (unchanged from 12.30.0)
```

The knock-on is that the recipe in the description and in `docs/api/scheduling.md:283` is exactly the call that throws, because `Schedule.timezone` is typed `string` (`src/types.ts:969`) while the column is nullable (`src/plans.ts:248`) — so TypeScript gives no warning:

```
legacy row timezone read back as: null | kind: cron
docs recipe THREW: Unknown or unsupported time zone: null
```

**Suggested fix: coalesce a falsy `tz` to `'UTC'` — `options.tz || 'UTC'` in both `schedule()` and `previewSchedule()` — and keep the throw for a truthy value the parser rejects.**

That is strictly more permissive than 12.30.0, so nothing that worked before breaks, and it is simpler than it looks because a falsy zone then never reaches a parser at all. Worth having, because the two engines do not agree on one (measured, host `TZ=America/Chicago`):

```
                cron-parser                    rrule-temporal
tz=null      -> host local (Chicago)        -> UTC
tz=""        -> THROW CronDate: unhandled…  -> THROW Invalid time zone specified:
tz=0         -> UTC                         -> THROW Invalid string: 0
```

Coalescing removes that table rather than forcing a choice out of it, and it also closes a latent 12.30.0 bug: `tz: 0` passed validation there (cron-parser reads the *number* `0` as unset), got written into a `text` column as the *string* `'0'`, and then threw on every pass — `"0" -> THROW CronDate: unhandled timestamp` — so the write succeeded and the schedule never fired.

**One more piece to finish it.** Coalescing stops new ambiguous rows, but the ones already in the table are still NULL, and the cron pass hands the column straight to cron-parser — so `previewSchedule()` would say UTC while delivery does host local. The v41 migration is already touching this table, so the cheapest close is one more statement in it:

```sql
UPDATE schedule SET timezone = 'UTC' WHERE timezone IS NULL;
```

After that the column has no NULLs, `Schedule.timezone: string` is true instead of a lie, and store, preview and pass all agree with no runtime coalescing needed in the pass at all. `timezone text NOT NULL DEFAULT 'UTC'` would keep it that way permanently, if that is not too much for this PR.

That backfill does change behavior for existing rows, so it needs a release-note line: a schedule whose `timezone` is NULL has been firing in the host's local zone and will fire in UTC after the upgrade. No-op on a UTC host, a shift by the offset otherwise.

It is worth saying plainly that host local was never a choice anyone made. Nothing in `src/` or `docs/` mentions it — `schedule()`'s default is `'UTC'`, the docs say "if not specified, the default is UTC", and the fallback exists only because cron-parser reads a non-string `tz` as "unset" and drops to the process zone. There are two ways a NULL got into that column: `schedule({ tz: null })` on an older release, where `{ tz = 'UTC' }` only defaults on `undefined` so an explicit null slipped past, and a row written directly with SQL. In both cases the author meant "no time zone specified", which is documented as UTC.

Nor is the current behavior stable enough to depend on: the zone is whatever `TZ` the instance that won `trySetCronTime` happens to carry, so in a cluster with mixed settings the same row fires at a different time depending on which instance took the pass. It is also the only place the process environment reaches into scheduling at all — everything else derives from database time, and `assertTimezone()` exists so that what `schedule()` accepts is exactly what the pass can evaluate. The backfill retires an accident rather than changing a feature.

**4. `__singletonSlot` is still live on the public `insert()` path (low).** The description says it "leaves `insert()` doing what it did before: ignoring the name" — it doesn't. `manager.ts:1169` spreads `...rest` from each caller object straight into the recordset, and `__singletonSlot` is a declared column in it now:

```
1. explicit slot honored on insert(): 2026-01-01T06:00:00.000Z
2. bad slot error: invalid input syntax for type timestamp: "not-a-timestamp"
3. two inserts sharing key+slot -> rows with that singleton key: 1
```

Same three properties the unprefixed name had. The prefix does cut the odds of an accidental collision, which is worth something — the description should just say that rather than say the path is closed. (`send()` is genuinely unaffected; it goes through `#toJobPayload`, which whitelists.) Closing it properly means giving the timekeeper a path `insert()` doesn't share.

**5. The relabel `UPDATE` isn't guarded by the expression it judged (low).** It matches on the primary key only, so a `schedule()` upsert landing between the pass's read and its write gets stamped with the previous expression's kind. Self-healing — the next pass detects and repairs — so the exposure is one pass. Costs one column in the recordset and one clause to close: `AND s.cron = k.cron`.

**6. The proxy doesn't mirror the `count` ceiling (low).** From #887. `previewScheduleQuerySchema.count` has no `.max(1000)`, so `count=1001` arrives as an opaque 500 instead of a 400, and the generated OpenAPI parameter has no `maximum` — a client generated from it will send 5000.

**7. `test/scheduleLastJobTest.ts:40` leaks a pool per call (low).** From #888. `helper.getDb()` opens a fresh `Db`; `test/bamTest.ts:20` is the same shape and ends `await db.close()`. Three call sites here.

**8. Nothing pins the `updated_on` guarantee (low).** From #888. The claim now appears in four places — the `setScheduleLastJobIds` comment, the new `setScheduleKinds` comment, the description, and the docs — and the only test touching `updatedOn` asserts it's a `Date`. A before/after assertion on both write paths is cheap.

**9. `missed: null` throws (low, consistency).** New option, so nothing to break either way — but it is the same shape as the `tz` question in Finding 3, and the two should land on the same answer. `assertMissedPolicy` rejects any non-member including nullish, so `{ missed: null }` from a config object throws while `{ missed: undefined }` defaults to `skip`. If falsy `tz` coalesces, falsy `missed` probably should too.

**10. What a capped catch-up costs, for the docs.** Mostly retired by the remedy in Finding 1 — the 220–340ms per capped schedule I measured against `previous()` becomes 1–7ms with the chunked read, so the `MAX_MISSED_OCCURRENCES` comment's "around a tenth of a second on the event loop" ends up conservative rather than optimistic. Worth re-measuring and restating either way, since as written it is wrong in both directions depending on which implementation ships. Separately, and unaffected by any of that: the 1000 catch-up jobs arrive as 1000 send-it jobs that the handler then forwards at `batchSize: 50` on a 1s poll, so the backlog takes something like twenty seconds to turn into real jobs. Neither is wrong; both are worth a sentence next to the cap, since the operator reading it is by definition in the middle of an incident.


---

<details>
<summary><b>Patch for Finding 1</b> — <code>occurrencesBefore()</code> as a chunked backwards <code>between()</code> (applies on <code>639cb7b</code>)</summary>

```diff
diff --git a/src/rrule.ts b/src/rrule.ts
index 7cdf00d..1ade6cd 100644
--- a/src/rrule.ts
+++ b/src/rrule.ts
@@ -332,41 +332,50 @@ function toDate (occurrence: { epochMilliseconds: number }): Date {
 }
 
 /**
- * The occurrences in `after` to `until`, the lower bound excluded and the upper included, read
- * backwards from `until` and cut off at `limit` of them. Newest first, since that is the end a
- * limit has to keep.
+ * How wide the first backwards chunk of a catch-up read is, and the factor each one after it grows
+ * by. An hour fills the cap outright for anything denser than a job a second, and four steps of
+ * growth reach a month, so a sparse expression over a long outage is read in a handful of calls.
+ */
+const CHUNK_SECONDS = 60 * 60
+const CHUNK_GROWTH = 4
+
+/**
+ * The occurrences in `after` to `until`, the lower bound excluded and the upper included, newest
+ * first and at most `limit` of them.
+ *
+ * Read in chunks that widen backwards from `until` rather than in one between() over the whole
+ * range, because the range a catch-up reads is as wide as the gap it covers: a single between()
+ * over three weeks of a per-second rule materializes every occurrence in those three weeks to hand
+ * back the few the caller will use. A dense expression fills the limit out of the first chunk; a
+ * sparse one reaches `after` in a few steps, since each chunk is four times the last. Either way
+ * the walk costs what it returns rather than what the gap spans.
  *
- * Backwards and bounded, rather than a between() of the whole range, because the range a catch-up
- * reads is as wide as the gap it covers: a between() over three weeks of a per-second rule
- * materializes every occurrence in those three weeks to hand back the few the caller will use. The
- * walk costs what it returns.
+ * Deliberately not rrule-temporal's previous(). That walks backwards from a phase-aligned DTSTART,
+ * and an RDATE is an absolute instant rather than a phase, so a rule carrying one answers with the
+ * RDATE in place of the rule occurrence that follows it and loses everything in between. between()
+ * has no such problem, which is what the due window has always used.
  */
 export function occurrencesBefore (expression: string, after: Date, until: Date, tz: string, limit: number): Date[] {
   const rule = cachedRule(expression, tz)
   const occurrences: Date[] = []
 
-  // Inclusive on the first step and exclusive on the rest, which walks each occurrence once from
-  // the upper bound the range includes.
-  let cursor = until
-  let inclusive = true
+  let upper = until
+  let width = CHUNK_SECONDS * 1000
 
-  while (occurrences.length < limit) {
-    const occurrence = rule.previous(cursor, inclusive)
+  while (occurrences.length < limit && upper.getTime() > after.getTime()) {
+    const lower = new Date(Math.max(after.getTime(), upper.getTime() - width))
 
-    if (occurrence === null) {
-      break
-    }
-
-    const date = toDate(occurrence)
+    // between() excludes both ends. The upper bound is nudged past so an occurrence on it is
+    // included, and the lower one stays excluded because it is the next chunk's upper bound —
+    // which is also what leaves the range's own lower bound, `after`, excluded.
+    const chunk = rule.between(lower, new Date(upper.getTime() + 1))
 
-    if (date.getTime() <= after.getTime()) {
-      break
+    for (let index = chunk.length - 1; index >= 0 && occurrences.length < limit; index--) {
+      occurrences.push(toDate(chunk[index]!))
     }
 
-    occurrences.push(date)
-
-    cursor = date
-    inclusive = false
+    upper = lower
+    width *= CHUNK_GROWTH
   }
 
   return occurrences
```

Applies cleanly on `639cb7b`. Measured on `c7c6884`, whose `src/rrule.ts` is identical: all eight RDATE/EXDATE shapes agree with `between()` over the same range, full suite 1127/1127, all five schedule suites unchanged.

</details>

kibertoad:
Thanks, that RDATE repro was useful. All ten are addressed in 62bff4f.

**1. `previous()` is gone.** I took your chunked `between()` as written and kept your comments, since they say why better than a restatement would. Reproduced your table first, then confirmed the fix: all eight RDATE/EXDATE shapes now agree with the forward window read over the same range, and the capped read measures 1 to 4ms for a rule against your 220 to 340 for the walk. The cron side of the same read measures 5 to 16ms, so the cap comment now states both numbers instead of "around a tenth of a second".

Your closing point about the shared `try` was the part I hadn't seen, and it is a real defect independent of the RDATE bug, so `missedOccurrences()` has a `try` of its own now. The divergence is reachable on the cron side too: the backwards read walks `prev()` up to 1000 times, and cron-parser gives up walking a sparse expression far enough back, which the forward window read never does because it reads one minute. Sharing the `try` meant that throw cost the row's due job and reported a valid schedule as unreadable. Split, the due job still goes out and the catch-up failure warns on its own, which the operator needs anyway: the claim has already moved, so that backlog is gone rather than deferred.

**2. Escaped, not JSON.** You laid out the trade and I think escaping wins outright, so `` `${esc(name)}__${key}` `` it is. Injectivity is easier than the general case, incidentally: `assertObjectName` allows `[\w.\-/]+` on both halves, so neither can hold a backslash, and escaping the name's underscores means the first unescaped `_` in the string can only be the separator. And the key stays byte-identical for every name that carries no underscore, so the only keys that move are the pairs that already collide, which is the transition worth having: nothing to document and no double fire during a rollout.

**3. Falsy is "none given", plus the backfill.** `options.tz || 'UTC'` in `schedule()` and `previewSchedule()`, so `null`, `''` and `0` all store and evaluate as UTC while a truthy zone the parser rejects still throws. That also retires the `tz: 0` case you found, where the write succeeded as `'0'` and every later pass threw.

I did the backfill, and I did not make the column `NOT NULL`. The backfill is unarguable and the same rows are already being read by the kind `UPDATE`. `NOT NULL` I left alone because of the rollout: an old instance's `schedule({ tz: null })` would then fail against the migrated schema, and a hard error on a call that used to succeed is louder than the ambiguity being closed. Instead the schedule read coalesces (`COALESCE(timezone, 'UTC') as timezone`), which covers a null written mid-rollout or by hand and makes `Schedule.timezone: string` honest without depending on a constraint. The behavior change the backfill does carry is in the description as a release-note line.

**4.** Closed rather than reworded. `insertJobs` only declares `"__singletonSlot" text` in the recordset when the caller asks for it, `insert()` strips the field from the objects it is handed, and the pass opts in with `{ __singletonSlots: true }` beside the jobs. A public caller naming it, either spelling, sets nothing.

**5.** `AND s.cron = k.cron`, with the expression carried in the recordset. A row that misses the update because it changed underneath is relabelled by the next pass if it still needs it.

**6.** `.max(1000)` on the proxy's `count`, so 1001 is a 400 and the OpenAPI parameter carries the maximum.

**7.** Closed in a `finally`, in all four places across the new suites, not just the three in `scheduleLastJobTest.ts`.

**8.** Pinned on both paths: a schedule that fires and a row the pass relabels each assert `updatedOn` unchanged. The relabel test also asserts the guard from finding 5, since it is the same statement.

**9.** `missed: null` reads as the default now, same as `tz`. A policy name is never falsy, so nothing a caller could have meant is read past.

**10.** Both numbers are in the cap comment, and the docs gained a bullet next to the cap for the part the read time does not cover: 1000 occurrences arrive as 1000 send-it jobs forwarded 50 a poll, so a capped catch-up reaches its queue over something like twenty seconds.

timgit:
I went back and forth with findings from the latest review.  Claude kept giving me scenarios and timings of how long it would take to recover with the 'all' catch-up policy when there were a lot of "every minute" schedules.  It's arguably an edge case, but I suppose it's not completely out of the realm of possibilities.  The limitation and subsequent backlog it produced comes from the hard-coded batch size of 50 in the timekeeper.  I think this should probably be bypassed for catch up activity regardless using something like insert() behind the scenes since we don't need the same distributed uniqueness guarantees that we currently have by dogfooding a queue for throttling.  It proposed deferral of 'all' for simplicity, but it's up to you on how we can make it a thing along with this release.

----

**1. Please drop `missed: 'all'` for now, and keep `skip` and `once`.**

`onSendIt` discards the occurrence time:

```ts
const { key, slot, ...request } = data
return await this.manager.send(request)
```

So a caught-up job carries the schedule's static `data` and nothing else. The docs say as much, as a caveat:

> **A caught-up job is indistinguishable from an on-time one.** It carries the schedule's `data` unchanged and is created when the pass catches up, so a handler cannot tell it is late, or which occurrence it is running for.

But that caveat voids the use case the policy is introduced with:

```js
// an hourly export has a file to write per hour, so every hour that was missed still has to run
await boss.schedule('export', '0 * * * *', null, { missed: 'all' })
```

Twelve identical jobs after a twelve-hour outage, `data: null`, no timestamp. The handler cannot know which hour each one is for, so it writes the same file twelve times or produces twelve copies of the same current-state export. **This is broken with one schedule and a half-day gap** — it needs no volume, and no cap setting fixes it, because lowering the cap bounds how many jobs appear without making any of them identifiable.

`all` *is* coherent for work that is self-describing at run time — "every minute, process up to 100 pending records," where 1000 catch-up runs genuinely drain a record backlog. It's incoherent for "each run corresponds to a time period," which is the documented example and I think the likelier reading.

`once` has none of this problem: one job whose meaning is "catch up to now" needs no occurrence identity, and it's bounded by construction.

There's a second, smaller thing in the same area. Every catch-up job goes through the one shared `__pgboss__send-it` queue at `batchSize: 50` on a 1s poll, so the drain is global rather than per-schedule. Measured whole-pass, every schedule on `all`:

```
schedules   expression         pass ms    jobs created   drain at 50/poll
     1      cron * * * * *          94           1001         20s
    10      cron * * * * *         243          10010          3.3 min
   100      cron * * * * *        1321         100100         33 min
```

The cap is what binds, not the outage — a per-minute schedule saturates 1000 after 16.7 hours, so a 17-hour gap and a 30-day gap produce identical numbers. This is moderate rather than alarming (at 10 schedules nobody notices), but the docs' "something like twenty seconds" is the per-schedule figure and the fleet figure is unstated.

Both of these are solvable, and your instinct about bypassing send-it for catch-up is the right shape — I checked and it holds up:

- **No new lock and no dedupe needed.** `trySetCronTime` already serializes this. One instance wins the single-row `UPDATE`; the loser re-evaluates `now() - cron_on ≈ 0`, gets zero rows, and the CTE joins `prior` against an empty `claim`, so it runs no pass at all. Successive ranges are disjoint because the next lower bound is the timestamp this pass wrote. The `__singletonSlot` machinery is solving the *due* path's problem — a 60s window read by passes every ≤45s genuinely overlaps — and catch-up has no such overlap.
- **`insert()` honors queue policies identically.** `createJob` builds the same `plans.insertJobs` statement, and `send()` returning `null` is just zero rows from `ON CONFLICT DO NOTHING`.

What it needs is three decisions, which is why I think it's a follow-up rather than something to bolt on here:

1. **One transaction around the per-queue inserts.** `insertJobs` resolves one table from one queue name, so catch-up across ten queues is ten statements. Today the send-it job is a write-ahead record of intent — the claim advances, one atomic insert records the work, and a failed `send()` is retried by the job system. Bypassing removes that, so a crash partway through loses the remaining queues permanently, because `cron_on` has already moved.
2. **Where occurrence identity lives.** `includeMetadata: true` is the right shape, but **not** via `startAfter`: the retry update rewrites it on every attempt but the last, so a handler reading it gets the occurrence on attempt 1 and a backoff timestamp after — a silently wrong answer, and these are exactly the jobs likely to retry. `singleton_on` survives the retry unchanged, is already in `JobWithMetadata`, and is already the occurrence-slot mechanism. Its cost is enrolling the job in `job_i4 UNIQUE (name, singleton_on, COALESCE(singleton_key, ''))`.
3. **What happens when the schedule's own send options set `singletonKey`**, which collides with (2) and also breaks `insert()`'s id alignment, so `lastJobId` can't assume nothing dropped.

Dropping `all` takes this PR back to the three-PR aggregation it was advertised as with no blockers, and it doesn't touch v41 — the catch-up commit carries no migration, so the schema bump belongs entirely to the first three commits. Worth noting `SCHEDULE_MISSED_POLICIES` is exported publicly as `scheduleMissedPolicies`, so shrinking the enum is a visible surface change, which is much better to do now than by withdrawing a shipped policy later. The docs are written around `all` (the cap bullet, the forwarding bullet, the export example), so that section wants rewriting rather than trimming — and it's the natural place to say what `once` does and doesn't promise.

---

**2. The chunked `between()` throws on a rule that is empty for a while and then dense, and this one is needed regardless of what happens to `all`.**

The widening is a guess about density, and a rule whose recent chunks are empty defeats it: the width grows until one chunk covers more occurrences than rrule-temporal will generate in a single call, and it throws rather than truncating.

```
                                              limit=1001, until=2026-02-01
MINUTELY UNTIL 25h before window     OK    n=1001  16.2ms
MINUTELY UNTIL 3d  before window     THROW         Maximum iterations (10000) exceeded in all()
SECONDLY UNTIL 25h before window     THROW         Maximum iterations (10000) exceeded in all()
MINUTELY live                        OK    n=1001   3.5ms
```

It's a regression rather than an inherited limit — the `previous()` walk handled two of those three:

```
OLD OK    MINUTELY UNTIL 3d before window      n=1001 198ms
OLD OK    SECONDLY UNTIL 25h before window     n=1001 139ms
OLD THROW SECONDLY BYMONTHDAY=1                Maximum iterations (10000) exceeded in all()
```

The middle row isn't exotic: a per-minute schedule carrying an `UNTIL`, over a gap that began before it expired. End to end through the pass, with a healthy cron schedule alongside it:

```
=== AS SHIPPED (62bff4f) ===
jobs forwarded per schedule: {"healthy":1}
warnings: ... could not be caught up on the gap since the last cron pass:
          Maximum iterations (10000) exceeded in all()
```

The whole backlog is lost. Your split `try` is what keeps this contained, and it works exactly as intended — the healthy schedule's due job and the failing row's own due job both still go out, and it warns rather than passing silently.

**The important part: this is not disposed of by dropping `all`.** The widening is limit-independent, so it reaches the same fatal chunk whether the caller wants 1001 occurrences or 2. `once` reads `limit + 1 = 2`, and on the shipped code:

```
missed:'once' read (limit=2)
  OK    MINUTELY UNTIL 3d before     n=2
  THROW SECONDLY UNTIL 25h before    Maximum iterations (10000) exceeded in all()
  THROW SECONDLY BYMONTHDAY=1        Maximum iterations (10000) exceeded in all()
  OK    MINUTELY live                n=2
```

Arguably worse there, since `once` sends exactly one job and it goes missing silently.

Lowering `MAX_MISSED_OCCURRENCES` doesn't help either — it still throws at a cap of 10, because the empty chunks widen identically whatever the limit. It's a chunk-width bug, not a cap bug.

Two ways to fix it. Narrow and retry when a chunk overruns, with the initial width as the floor — that floor is safe because no legal expression can overrun the ceiling inside it, since a rule can't recur more often than once a second and an hour holds at most 3600. I have that built and measured: all the shapes above pass, equivalence holds at 700 combinations with 0 mismatches, and it comes with a regression test that fails on `62bff4f` with the exact error. Happy to hand it over.

Or, probably better now: `rrule-temporal` exposes `all(iterator)` taking `(date, i) => boolean`, so a bounded read is available from the library with no chunking at all, and `maxIterations` is a constructor option. With `all` gone and `once` reading two occurrences, the chunking earns a lot less than it was designed for, so the library's own stopping condition may be the cleaner shape.

---

**3. The escaped singleton key moves for every underscored queue name, not only the colliding pairs (non-blocking, and independent of the above).**

The collision fix is real — brute-forced over the allowed alphabet, 66822 name/key pairs:

```
old `${name}__${key}`:      collisions=1806   e.g. ("aa","_") vs ("aa_","") -> "aa___"
new `${esc(name)}__${key}`: collisions=0
```

And injectivity genuinely holds: `assertObjectName` is `/^[\w.\-/]+$/` for both halves, so neither can contain a backslash and the escape can't be forged. But only the *name* is escaped, so:

```
name                    key        old -> new                          changed  collides-today
send_email             ''         send_email__         -> send\_email__          true     false
billing/invoice_run    ''         billing/invoice_run__ -> billing/invoice\_run__ true     false
nightly_rollup         us_east    nightly_rollup__us_east -> nightly\_rollup__us_east true  false
orders                 eu_west    orders__eu_west      -> orders__eu_west        false    false
```

So the rolling-upgrade double-fire is narrowed from every schedule to every schedule on an underscored queue, rather than eliminated. Still the right trade against the JSON form, which moved all of them — it just wants the release-note line after all.

---

**Smaller things, none blocking:**

- `timezone text DEFAULT 'UTC'` on the fresh-install DDL would catch a hand-written insert too, with none of the rollout risk that made you decline `NOT NULL`. The read-side `COALESCE` already covers the rollout case.
- `PREVIEW_MAX_COUNT` still isn't exported, so the proxy's `.max(1000)` and the method's ceiling are two literals that agree today.
- One correction for the record: the reason given for splitting the `try` — that cron-parser gives up walking a sparse expression far enough back — didn't reproduce for me. cron-parser doesn't give up, it just gets slow (`0 0 29 2 *` over 500 years: 608ms). The split is still right; it earns its keep against the rrule read instead.

I also dropped a finding I'd drafted about the cap comment's timing numbers being off on the cron side. They are, but those numbers exist to argue that read cost isn't what sets the cap, and being out by a factor of six doesn't move that conclusion.

kibertoad:
Both of the big ones are in `dae73f8`, along with the three smaller items, and `1f9566a` fixes a regression test that was not testing the regression. Description updated to match.

**1. `all` is dropped.** You are right, and the argument you made is the one that settles it: the policy could not name the occurrence its jobs stood for, and the example the docs introduced it with is exactly the case that breaks. No cap setting rescues that. So `SCHEDULE_MISSED_POLICIES` is `skip` and `once`, `ScheduleMissedPolicy` and the proxy enum shrink with it, and `MAX_MISSED_OCCURRENCES` and the `missed_occurrences_capped` warning are gone rather than left unreachable, which means the dashboard's warning list and `docs/api/events.md` go back to what they were on master. The docs section is rewritten around `once` rather than trimmed: what it does not promise is now the first bullet under it, since that is the thing a reader has to know before choosing it.

Dropping it simplified more than the policy. The catch-up read wants one occurrence, so `occurrencesBefore(..., limit)` is `latestOccurrenceBefore(...)` returning a `Date | null`, the pass reads one occurrence per schedule, the limit+1 probe and the reverse() are gone, and the cron path is a single `prev()` rather than a walk of up to a thousand.

Your three open questions are the reason it is a follow-up rather than something to bolt on, and I agree with your reading of all three. On (2) in particular: `startAfter` being rewritten by the retry update is a trap I would have walked into.

**2. The narrowing.** Reproduced your table first, then built it. A step that overruns halves and is read again; the width it overran at becomes a ceiling no later step goes back over; an hour is the floor. The floor is safe by construction rather than by measurement, which is why I picked it over the library's `all(iterator)`: nothing recurs more often than once a second, so an hour holds at most 3601 occurrences against a 10000 ceiling. `all(iterator)` reads forward from DTSTART, and with the epoch anchor a rule with no DTSTART of its own would iterate 29 million times to reach 2026, so the stopping condition is in the wrong place for a backwards read.

Your four shapes, on the merged code:

```
measured with the read as it now stands, which asks for one occurrence

MINUTELY UNTIL 25h before window     OK   4 steps,  0 narrowings    6ms
MINUTELY UNTIL 3d  before window     OK   4 steps,  0 narrowings    2ms
SECONDLY UNTIL 25h before window     OK  10 steps,  5 narrowings   70ms
SECONDLY UNTIL 20d before window     OK  16 steps,  7 narrowings   84ms
MINUTELY live                        OK   1 step
```

One thing your table does not carry over unchanged, since the read is limit-independent but the shapes are not: at the two occurrences `once` asked for on `62bff4f`, the per-minute rules survive and only the per-second ones throw. Your row for `MINUTELY UNTIL 3d` under `once` says the same, and I had written both regression tests around the per-minute shape before I checked, so they passed on `62bff4f` and proved nothing. Both use a per-second rule now, and I confirmed each throws against the old chunking: `test/rruleTest.ts` reads eight shapes backwards over a 30-day range, of which `secondly, spent a day back` and `secondly, spent twenty days back` are the two that overran, and `test/scheduleMissedTest.ts` runs a per-second rule spent three days before a 30-day gap through a whole pass, asserting the catch-up job arrives and no warning does.

**One correction, and it is against my own comment rather than yours.** With the narrowing in, I could not construct a case where the catch-up read fails and the window read does not, and I tried for a while: the failure modes that survive are `COUNT`-limited rules, where rrule-temporal cannot start near the window at all and walks from the true DTSTART, and those refuse a 60-second `between()` just as flatly as a one-hour one. So the window read throws first and the row is skipped before the catch-up is reached.

That retires both reasons I had given for splitting the `try`: yours about cron-parser, which you already corrected, and mine about the rrule read. The split stays, because coupling the two was a real defect once and two separate calls over different ranges have no business sharing one, but the comment now says that rather than naming a divergence I cannot demonstrate. I dropped the timekeeper test I had written for it, since it was asserting a scenario I could not actually produce, and put the coverage where it is real: a unit test that `latestOccurrenceBefore` reports a rule it cannot read at its narrowest step rather than looping on it, alongside the window read refusing the same expression.

**3. The escaped key wants a release-note line.** Agreed, and your brute force is the right way to have found it. It is in the notes now, stated as what it is: the key moves for every schedule whose queue name contains an underscore, not only for the pairs that collide today, so one occurrence of such a schedule can be sent twice during the rollout. Bounded to the rollout window, and a name with no underscore is byte-identical.

**Smaller things:**

- `timezone text DEFAULT 'UTC'` on the fresh-install DDL, and `ALTER COLUMN timezone SET DEFAULT 'UTC'` in v41 so a migrated database is not left drifting from a fresh one. `uninstall` leaves the default in place, the way v36 did with `bam.created_on`: a v40 instance names the column on every write, so the default it would fall back on never applies. Manifest regenerated.
- `PREVIEW_MAX_COUNT` is exported as `previewScheduleMaxCount`. The proxy's literal stays a literal for now with a comment naming the export, since it builds against `^12.30.0`, so it moves at the same dependency bump as `scheduleKinds` and `scheduleMissedPolicies`.
- The cap comment is gone with the cap, so the timing numbers in it are not wrong in either direction any more.

timgit:
One blocker left, and **it isn't from these two commits** — it came in with `b679963`, and I missed it twice by checking those statements individually instead of in the single transaction the migration actually is.

---

**1. The v41 migration can't run on CockroachDB.**

`ALTER TABLE … ADD COLUMN kind` and `UPDATE … SET kind = 'rrule'` are in the same transaction, and CockroachDB won't let a transaction write a column it just added. Through pg-boss's own migration path — install at v41, roll back to 40, migrate:

```
rolled back to 40
MIGRATE FAILED  42P10  column "kind" is being backfilled
```

Narrowed to the pattern rather than the text:

```
as generated (ADD kind + UPDATE kind)     FAIL 42P10 column "kind" is being backfilled
UPDATE not reading kind in WHERE          FAIL 42703 column "kind" does not exist
nullable ADD then UPDATE                  FAIL 42703 column "kind" does not exist
ADD with computed DEFAULT only, no UPDATE OK
two txns                                  OK
timezone UPDATE + SET DEFAULT only        OK
```

So it's the backfill specifically — not the `CHECK`, not the default, and not the `SET DEFAULT` this round added. Scope is migration only: a **fresh** v41 install on CockroachDB is fine (`schema version = 41`), and `distributedDatabaseTest` is 13 passed / 1 skipped on this head because it installs fresh. **YugabyteDB migrates fine**, same rollback-and-migrate cycle.

The fix I'd suggest, verified: **drop the `kind` backfill for distributed backends.** With only that statement removed the migration runs (`MIGRATE OK -> 41`), and nothing is actually lost — the relabel fallback already covers it. A row whose `kind` disagrees with its expression is read correctly by `dueOccurrences` and rewritten by `setScheduleKinds` on the first pass, and that statement is accepted on CockroachDB. `getAll(schema, noPartitioning, noCovering)` already takes this shape of flag. The alternative is running the backfill in a second transaction for everyone, which is a contractor change rather than a migration-store one. Trade-off worth knowing: without the backfill, `getSchedules()` on CockroachDB reports `cron` for a legacy rule row until a pass relabels it.

One thing that is on me rather than on you: **no CI leg runs a migration against CockroachDB**, which is why this had nowhere to be caught. The CockroachDB job runs `distributedDatabaseTest.ts` only, which installs fresh and never migrates; the `distributed` matrix leg runs against Postgres by design; and `migrationTest.ts` is already red on CockroachDB for harness reasons unrelated to any migration. I'll open that separately and fix it on my side — nothing there is yours to deal with, and it is not a condition of merging this.

---

**Three smaller ones, none blocking:**

- **The narrow-and-retry can cost 1.3s of synchronous pass time.** The floor is an hour, and a rule that's empty for days and then dense forces the walk down to it: `FREQ=SECONDLY;BYMONTHDAY=1` takes **1319ms** to answer, inside the pass. It's bounded by the distance back to the nearest occurrence rather than by the gap (same 1.3s at 30 days and at 90), and a merely spent rule is cheap — `SECONDLY` with an `UNTIL` 31 days back over a 30-day gap answers null in 10ms. Not a regression, since both earlier implementations threw on this shape. But the module comment says the walk "costs what it returns", and at the floor it costs what it searches — worth a sentence there, or a step budget beside the width ceiling.
- **I'd rather not take the `toad-cache` dependency, and I don't think the cache is needed at all.** It is one `LruMap` for one purpose — keeping built `RRuleTemporal` instances between passes — and pg-boss ships three runtime dependencies today, so each new one is install surface for everyone. Measured against a copy of `rrule.ts` whose `cachedRule` always rebuilds, it is worth about a tenth of a millisecond per rule schedule per pass (0.13ms vs 0.21ms for a per-minute rule, due read plus catch-up read), which is 10ms a pass at 100 rule schedules on a 30s interval.

  The one caller where it earns anything is `previewSchedule()`, and only because the rrule branch of the iterator (`timekeeper.ts:764`) calls `nextOccurrence(expression, after, tz)` per occurrence, so `count: 1000` builds the rule a thousand times: 126ms cached against 186ms uncached on `FREQ=MONTHLY;BYDAY=-1FR`. That one is worth fixing on its own terms rather than via a module-level cache, since the budget it burns is `PREVIEW_TIME_BUDGET_MS` and a sparse expression near the ceiling loses returned occurrences, not just milliseconds — build once and close over it, the way the cron branch two lines below already does with `parseRecurrence`. After that the pass builds two rules per schedule and there is nothing left for an LRU to do, and nothing holds up to 1000 rule instances for the life of the process.

  Correctness does not depend on the choice either way: the rules are immutable and `between()` caches nothing that depends on its arguments, which your own comment says. (There is a bigger version of this — the pass evaluates every row separately, so 500 queues on `* * * * *` parse it 500 times a pass — but that predates this PR and I'm filing it separately.)

- **`onSendIt`'s comment still describes the `all` batch** (`src/timekeeper.ts:661`): "a schedule catching up on a gap can put a job per missed occurrence in it, which the fetch hands over in no particular order". At most two now, the catch-up and the due one, which is what the test below it pins. The keyed-latest logic is still right; only the reason is stale.

kibertoad:
All four are in `9d87a00` and `a802e37`, and the description is updated to match.

**1. The v41 migration on CockroachDB.** Fixed the way you suggested. Both directions reproduced on a three-node cluster through pg-boss's own migration path, install at 41, roll back to 40, insert a rule row, migrate:

```
backfill in place   MIGRATE FAILED  42P10  column "kind" is being backfilled
backfill dropped    MIGRATE OK -> 41   row: {"key":"rule","kind":"cron","timezone":"UTC"}
```

The statement is dropped behind a new `noAddColumnBackfill` compatibility flag rather than on the backend's *kind*: YugabyteDB migrates with it in place, as you found, so the flag names the engine limitation instead of the topology, and only CockroachDB's profile sets it. `getAll()` takes it as a third flag beside `noPartitioning` and `noCovering`, and `test/migrationTest.ts` pins that v41 keeps the `ADD COLUMN`, both zone statements and `last_job_id` when it is on, and loses only the backfill.

Nothing else about the migration moves, which the row above shows: the zone backfill still ran on it. The relabel the fallback depends on runs there too, checked on its own since the whole fix rests on it:

```
UPDATE ... FROM json_to_recordset(...)  ->  UPDATE 1, kind = rrule
```

So the cost is exactly the one you named, and it is in the docs now: on CockroachDB `getSchedules()` reports `cron` for a legacy rule row until a pass relabels it. `docs/database-backends.md` carries the flag in the capability matrix and the flag table, and the notes carry the line.

**2. The narrowing cost.** A sentence rather than a step budget. A budget converts a 1.3-second read into a lost catch-up and a warning, and for the one shape that pays it that is the worse of the two. The comment now says that with the ceiling at the floor the read costs what it searches rather than what it returns, names `FREQ=SECONDLY;BYMONTHDAY=1` and the 1.3 seconds, and says the bound is the distance back to the nearest occurrence rather than the width of the gap, so 30 days and 90 cost the same, while a rule already spent when the range opens keeps its widening and answers null in 10ms.

**3. `toad-cache` is gone**, and you were right on both halves. Rules are built per read, and the caller that was rebuilding one a thousand times builds it once: `rruleWalker()` in `src/rrule.ts` holds the cursor and closes over the rule, so `occurrenceWalker()` is one line per format and the rrule branch works the way the cron branch below it already did with `parseRecurrence`. The test that exercised the LRU cap is a plain semantics test now, and a new one walks a rule to `count: 1000` inside the budget.

**4. `onSendIt`'s comment** says at most two, the catch-up and the one due.

PR #899, opened 2026-09-09
feat(db): allow faking DB clock for testing

### LINKED ISSUE #689: Allow controlling Timekeeper in tests for `sendAfter`
Currently, `Timekeeper` only supports physical wall-clock time (it uses `Date.now()` and `setInterval()`), and there's no way to provide a fake version in tests.

Ideally, we'd have some way to provide the current time from outside in a test (`Date.now()`) as well as a way to control when callbacks passed to `setInterval()` fire.

Jest provides [timer mocks](https://jestjs.io/docs/timer-mocks) which can help, but they require knowing the implementation of `Timekeeper` in order to function.

bhamiltoncx:
This is part 1 of 4 PRs to fix #689.

For part 2, see https://github.com/bhamiltoncx/pg-boss/pull/1.
For part 3, see https://github.com/bhamiltoncx/pg-boss/pull/2.
For part 4, see https://github.com/bhamiltoncx/pg-boss/pull/3.

This PR introduces a new DB schema function `pgboss.now()` which can be replaced in test environments with a clock that can be controlled by the test.

A new `eslint` check ensures that all DB time/date function calls invoked by `pg-boss` go through `pgboss.now()` instead of the default Postgres implementation.

Because this changes all `DEFAULT` timestamp values from `now()` to `${schema}.now()`, this requires a DB migration.

I considered updating `search_path` to prepend a new schema, but that is connection-wide and would have potentially affected other queries outside of `pg-boss`. This change only affects SQL queries invoked by `pg-boss` itself.

I developed this PR with help from Claude Code.


bhamiltoncx:
@timgit, I rebased this on master and changed the DB schema bump to v42. I labeled the release as 12.32.0 but we can keep it 12.31.0 if we want to ship them both together.

timgit:
Thanks for the PR.  Would you please consolidate these so I can review the consumer side in addition to the new now() function?

bhamiltoncx:
Of course, will do. (I tried to keep the PRs small so they'd be easier to review, but I'm happy to send one combined one as well.)

bhamiltoncx:
Sent combined PR https://github.com/timgit/pg-boss/pull/901 . Happy to land all at once or piece-by-piece.

PR #901, opened 2026-09-10
feat: controllable clock implementation for JS and DB layers

### LINKED ISSUE #689: Allow controlling Timekeeper in tests for `sendAfter`

bhamiltoncx:
Fixes #689. 

This is a combined version of the four-part PR series:

1) #899 
2) https://github.com/bhamiltoncx/pg-boss/pull/1
3) https://github.com/bhamiltoncx/pg-boss/pull/2
4) https://github.com/bhamiltoncx/pg-boss/pull/3

Tests can now control time for both JS and the DB simultaneously:

```js
import { PgBoss, TestClock } from 'pg-boss'

const clock = new TestClock('2026-01-01T00:00:00Z')
const boss = new PgBoss({ connectionString, clock, __test__enableSpies: true })
await boss.start()

const id = await boss.send('q', {}, { startAfter: 60 })
await clock.tick(60_000)   // Date, Postgres now(), and the worker's poll all move
await boss.getSpy('q').waitForJobWithId(id, 'completed')
```

I wrote this PR series with help from Claude Code.


timgit:
Thanks for consolidating.  Below are the findings.  

--------------

### 1. Drop the v42 repointing of column defaults

This is the big one, and it resolves a CockroachDB blocker as a side effect.

As it stands, v42 rolls back cleanly on Postgres and not on CockroachDB:

```
error: cannot drop function "now" because other objects
       ([pgboss.<schema>.create_queue]) still depend on it            SQLSTATE 2BP01
```

CockroachDB records a dependency from a UDF to every function reachable through the column defaults of
the tables it writes. `create_queue` inserts into `pgboss.queue`, whose `created_on`/`updated_on`
defaults v42 repoints at `pgboss.now()`. Resetting a default doesn't clear that edge. **8 of 41
migration tests fail** — every test that rolls back, not just the v42 one. CI can't see it, because
`npm run test:cockroachdb` only runs `distributedDatabaseTest.ts`, which never migrates backwards.

I can fix that narrowly by re-issuing `create_queue` from a frozen v42 snapshot before the drop, and
I have a patch that does. But the better fix is to not create the dependency at all, because I can't
find anyone the repointing actually serves:

- **Production can't be served by it.** With no clock attached, `${schema}.now()` and
  `pg_catalog.now()` return the same value. The ALTER is a no-op there by construction.
- **The test suite isn't served by it.** I stripped the repointing entirely — the 11 `default
  ${schema}.now()` positions in `plans.ts`, the `ALTER`s in v42, manifest regenerated — and kept
  everything else: the function, the 78 call sites, `TestClock`, `attach()`. Result: **1222/1223**,
  and the single failure is `migrationTest`'s "v42 … points every timestamp default at it", i.e. the
  test whose only job is asserting the repointing happened. Every behavioural test passes, all the
  `TestClock` suites included.

What's left is raw-SQL inserts into `pgboss.job` honouring a fake clock, which I think is outside the scope of the PR.

With the repointing dropped, on CockroachDB:

| `migrationTest.ts` on CRDB | result |
|---|---|
| this PR | 8 failed, 33 passed |
| with my narrow rollback patch | 41 passed |
| **with the repointing dropped** | **1 failed, 40 passed**, and zero occurrences of `2BP01` |

`driftTest.ts` on CRDB is unchanged at 10 failed / 93 passed either way — those are pre-existing.

So: v42 becomes `CREATE FUNCTION` / `DROP FUNCTION`, no ACCESS EXCLUSIVE on `job`, blocker gone,
feature intact.

**This needs one change alongside it.** `job.created_on` is never written by
pg-boss's own SQL — it always comes from the column default, at four insert sites in `plans.ts`: the
shared `send()`/`insert()` path, both dead-letter inserts, and redrive. (`start_after` and
`keep_until` are already explicit; the three sites that do name `created_on` are archive/move paths
copying an existing value, and `source_created_on` on a dead-letter row is the source job's stamp, not
the new row's.) So without the repointing, a job created under a fake clock would get a real-time
`created_on` beside a fake `start_after`. Adding `created_on` to those four column lists as
`${schema}.now()` closes it.

### 2. `stop()` never returns under a `TestClock` while a handler is running

`index.ts` puts the graceful-shutdown deadline on the configured clock, while the poll watching for it
correctly stays on real time:

```ts
const deadlineTimer = this.#config.clock.setTimeout(() => { deadline.reached = true }, timeout)
while (!deadline.reached && this.#manager.hasPendingCleanups()) {
  await delay(500)   // real time, deliberately
}
```

Under a `TestClock` nothing ever sets `deadline.reached`, so it busy-polls forever:

```
job state after tick: active
stop({timeout:1000}) -> TIMED OUT after 8s real time
```

That turns the most common call in a teardown into an unbounded hang with no diagnostic, and a handler
waiting on a clock timer can't be ticked free because the test is already blocked inside `stop()`. The
docs mention it, but the deadline bounds shutdown I/O rather than queue time, so by the rule the stack
already applies to `PREVIEW_TIME_BUDGET_MS` and the `delay(500)` poll it belongs on the real clock.
Using the global `setTimeout` gives `stopped (1039ms)` and leaves the suite at 1223/1223.

Same rule, lower stakes, two more places: the slow-query and `#statsElapsedSeconds` stopwatches in
`boss.ts` bracket real I/O with `clock.now()`, so they read zero under a fake clock — and a `tick()`
landing between the two reads can persist a real `version.monitor_backoff_on`.

### 3. `clock: undefined` throws instead of defaulting

```
{ connectionString: … }                     -> ok
{ connectionString: …, clock: undefined }   -> configuration assert: clock must implement now()
```

`attorney.ts` gates on `'clock' in config`. That's the shape any conditional harness reaches for
(`clock: process.env.FAKE_TIME ? testClock : undefined`), and it's inconsistent with every other
optional option in the file. `if (config.clock)` fixes it.

---

### Worth doing, not blocking

**A killed test leaves the schema on frozen time, and nothing repairs it.** After a killed run, a plain
`new PgBoss()` with no clock option starts clean and silently runs on the leaked clock — every job
stamped 2020, retention computed from 2020. Blast radius is any database a suite has ever pointed at.
A warning doesn't help, since the instance still runs on 2020. What does is gating the override in the
body so a default instance is immune by construction:

```sql
SELECT COALESCE(
  CASE WHEN current_setting('pgboss.test_clock', true) = 'on'
       THEN (SELECT c.now FROM ${schema}.clock c LIMIT 1) END,
  pg_catalog.now());
```

with a `__test__enableClockOverride` flag (matching `__test__enableSpies`) that both gates `attach()` and
makes `Db` set `pgboss.test_clock = 'on'` on the connections it opens. Verified on PG18 and CRDB: an
opted-in pool reads 2020 on all five sessions while a plain pool on the same schema reads 2026, for
direct calls and column defaults alike, and `RESET ALL` fails back to real time. It's an AND with the
table, not a replacement — the row only exists while a clock is attached — and the only vector that
reaches another session is `ALTER DATABASE`/`ALTER SYSTEM`, both superuser-only. A leak then costs
~5× on the clock read instead of silently corrupting timestamps.

**CockroachDB, clock suites:** 3 failures. Two are `count(*)::int` coming back as a string in
`countJobs` (one `Number()`, same class as the #679 work). The third asserts drift is clean while
attached, which can't hold there — CRDB rewrites the override body, so the `clockOverride` substitution
no longer matches. Worth skipping on that backend.

**`delay()` has dead parameters now.** The Coveralls dip is entirely four lines in `tools.ts`: every
caller that passed `error` or an `abortController` moved to `clockDelay`, and the nine remaining
`delay()` call sites in `src/` all pass one argument. Dropping the two parameters removes the coverage
regression and a duplicate of `clockDelay`'s abort semantics.

**`tick()` fires an interval once per period crossed**, so `tick(3600_000)` on a 2s poll runs 1800
fetch cycles. Nothing here does that, but a sentence in the `tick` docs pointing at `setTime()` would
save someone an afternoon.

**First commit message says "Schema v41"** — it's 42.

bhamiltoncx:
Thanks, will address the schema rollback issue with CockroachDB.

> The test suite isn't served by it

It's true that I didn't add any tests which fail without `TestClock`; this PR just makes the existing tests much faster (down from about 42 s to about 19 s of wall-clock time), since they no longer need to sleep to make the physical DB clock tick. It also reduces flakiness when the physical system clock doesn't behave the way we'd like it to (e.g., moving backwards).

I'm happy to add tests that require `TestClock` to function (e.g., moving time backwards, or moving time a large amount).

timgit:
Thanks. I think the cockroach issue goes away with the default change request. Sorry for the verbose agent output.

bhamiltoncx:
No worries.

Just to make sure I fully understand, the issue is that schema defaults can't be redirected anyway since they're resolved too early for the test environment to substitute the implementation, correct?

timgit:
schema defaults are overkill for the test suite imo

bhamiltoncx:
OK, removed the schema defaults, added a few new tests that can only pass with `TestClock`, and addressed the code review comments.

In specific:

> 1. Drop the v42 repointing of column defaults

Column defaults dropped. There were a few queries implicitly relying on the default `start_after` and/or `created_on` and thus were untestable:

1) Dead-letter and redrive inserts were missing both `start_after` and `created_on` parameters
2) Stats snapshots omitted `captured_on`
3) Schedule and warning inserts omitted `created_on`

Queue and subscription rows were left alone and still use the defaults.

> 2. stop() never returns under a TestClock while a handler is running

Fixed, test added.

> 3. clock: undefined throws instead of defaulting

Fixed, test added.

> A killed test leaves the schema on frozen time, and nothing repairs it

Fixed by setting `pgboss.test_clock = 'on'` for sessions and checking it before using the test clock.

Killing a test and invoking `RESET ALL` now returns a session to the real clock.

> delay() has dead parameters now

Removed dead parameters. The one caller passing an AbortController now uses the returned promise's own abort.

> CockroachDB, clock suites: 3 failures

Fixed.

> tick() fires an interval once per period crossed

Updated docs. `tick()` fires once per period crossed so long jumps belong to `setTime()`; the stop caveat is replaced.

New tests added for clock moving backwards as well as a year-long jump expiring retention and files stats.

> First commit message says "Schema v41" — it's 42.

Fixed and force-pushed to rewrite commit message.

timgit:
Re-reviewed at `4e3e204` against PG 18, CockroachDB v26.2 and PGlite. You took all three, and the
first one the better way — deleting the repointing rather than patching around CockroachDB's
dependency. That fixes the blocker and makes v42 a single `CREATE FUNCTION` with a single
`DROP FUNCTION` rollback, which is about as small as a schema version gets.

Everything from the last pass is closed:

| | |
|---|---|
| v42 rollback on CockroachDB | **fixed** — `migrationTest.ts` is now **41 passed / 16 skipped**, was 8 failed |
| `stop()` hanging under a `TestClock` | **fixed**, with a test that asserts it returns on real time |
| `clock: undefined` throwing | **fixed** |
| leaked fake clock reaching a plain instance | **fixed** — see below |
| stopwatches on the fake clock | **fixed** |
| the three CockroachDB clock-suite failures | **fixed** — clock suites are **26 passed / 2 skipped** there |
| dead `delay()` parameters, `tick` catch-up docs | **fixed** |

PG 18 is **1230/1230 in 58s**. eslint, tsc and the manifest check are clean. (PGlite fails ~53 files
here, but the merge base fails 55 the same way, so that is the local environment, not this branch.)

Deriving the opt-in from `isAttachable(clock)` instead of adding a constructor flag is better than
what the last pass suggested — same protection, one fewer public option. The leak case re-ran end to
end: a process attaches a clock at 2020 and exits without `stop()`, a second process starts a plain
instance on that schema, the override is still installed, and it reads 2026 and stamps its jobs in
2026. That was the whole point of the finding and it holds.

Four new things, and only the second is worth holding a squash for: it is the price of the gate,
then `plans.ts` and `migrationStore.ts` describing the same v42 objects two different ways — once
invisibly to every check we have, once visibly but masked by a normalizer.

### 1. (minor) A `TestClock` on a pooled custom `db` adapter now silently runs on real time

`Db.open()` opts in every connection pg-boss's own pool opens. A custom `db` adapter never goes
through `Db`, so its only opt-in is the single `SET` `attach()` issues, which lands on whichever
connection the pool handed out. Plain `pg.Pool` adapter, `max: 5`, same `TestClock(2020-01-01)`:

```
before the gate   5 concurrent reads through the pool: 2020,2020,2020,2020,2020
                  job created_on: 2020-01-01T00:00:00.000Z

4e3e204           5 concurrent reads through the pool: 2020,2026,2026,2026,2020
                  job created_on: 2026-09-11T18:41:34.891Z    <- real time, under an attached TestClock
```

Nothing throws. The test just stops testing what it says it tests, and which statements get fake time
depends on connection checkout order. To be clear about the scope: this is test ergonomics, not
correctness — production is covered by the gate itself, and nobody wants an override there. The whole
population is someone writing clock tests on a pooled adapter, and what they get is flaky clock tests
with nothing pointing at why. It is also a capability the gate removes rather than one it fails to
add: those adapters worked before it. Of the six adapters in `src/adapters/`, only `pglite` is
single-connection — `drizzle`, `knex`, `kysely`, `prisma` and `bun` are all pooled. The sentence in
`docs/api/testing.md` is accurate, but it can't do the work here, because the failure is silence.

The clean fix is to refuse the configuration rather than detect it after the fact. `Db` already
carries `_pgbdb`, so pg-boss knows whether it owns the pool; an adapter that does its own session
setup — or has only one session — says so:

```ts
// index.ts, before attach()
assert(this.#db._pgbdb || this.#db.clockSessionSetup,
  `clock assert: this db adapter does not declare clockSessionSetup, so a TestClock cannot reach every session. Run "${plans.enableClockOverride()}" on every connection it opens, then set clockSessionSetup: true on the adapter.`)
```

with `clockSessionSetup?: boolean` on `IDatabase` and `fromPglite` setting it (single connection, so
`attach()`'s `SET` covers everything). Measured:

```
pooled adapter, no flag                 -> guard fires, with the SET in the message
pooled adapter + connect hook + flag    -> 5 concurrent reads: 2020,2020,2020,2020,2020
                                           job created_on: 2020-01-01T00:00:00.000Z
```

Three files, +17 lines, suite **1230/1230**, PGlite clock suites green, eslint and tsc clean; the
diff is ready if that helps.

If that is more than it is worth, the other two landings are: leave the sentence in
`docs/api/testing.md` and let whoever hits it debug once, or restore the capability instead of
forbidding it by leasing the clock row (`WHERE c.expires_on > pg_catalog.now()`) rather than gating on
a session setting — no session state, so every adapter works again and a killed test's clock
self-heals, at the price of a renewal timer in `TestClock`. The guard is simply the smallest of the
three.

Two small things alongside it — `dispose()` doesn't `RESET pgboss.test_clock` (moot when pg-boss owns
the pool, not when the application's connections outlive the test), and `CLOCK_OVERRIDE_SETTING` /
`enableClockOverride()` aren't exported, so an adapter author has to copy the string out of the docs.

### 2. `plans.ts` changes 11 column defaults, and nothing in v42 matches them

`7ef042a` rewrites every table DDL default from `now()` to `pg_catalog.now()`, but v42 has no
`ALTER … SET DEFAULT` to go with it. Fresh installs from this branch get one spelling, every install
from an earlier release keeps the other, and nothing moves them.

The values are the same, which is what makes it easy to miss: a default is stored as a parse tree, so
both bind to the same OID and both render as `now()`. On the real upgrade path — a v41 schema built
by master's plans, then migrated by this branch — the result is byte-identical to a fresh v42 install
(16 clock defaults, 0 rows differ, drift clean on both), and the manifest is unchanged either way. Which also
means no test can catch this: the round-trip and drift checks compare catalogs, and the catalog has
already collapsed the difference. The plans/migrationStore invariant is the only thing holding here.

Putting `plans.ts` back to `now()` closes it, with a scoped `eslint-disable` on the six `createTable*` functions
— `createTableQueue`, `createTableSchedule`, `createTableSubscription`, `createTableWarning`,
`createTableJob`, `createTableQueueStats` — same shape as the `bam.created_on` opt-out already in the
file, since the clock guard is what pushed it to the qualified spelling:

```diff
+/* eslint-disable no-restricted-syntax -- a table default is stored as a parse tree, so
+   pg_catalog.now() and now() are the same expression; keeping the existing spelling means this
+   DDL needs no migration to stay in step with installs from earlier releases */
 function createTableQueue (schema: string) {
   return `
     CREATE TABLE ${schema}.queue (
       ...
-      created_on timestamp with time zone not null default pg_catalog.now(),
-      updated_on timestamp with time zone not null default pg_catalog.now(),
+      created_on timestamp with time zone not null default now(),
+      updated_on timestamp with time zone not null default now(),
```

Then both install paths spell the same DDL, there's nothing to migrate, and v42 stays a lone
`CREATE FUNCTION`. Measured on `4e3e204`: manifest unchanged, eslint clean, **1230/1230**. The diff is
ready if that is easier — 11 defaults and six comment pairs, nothing else.

One loose end in the same commit either way: `docs/sql/job-table.md` went back to `DEFAULT now()`
but `docs/sql/warning-table.md` still documents `created_on … DEFAULT pgboss.now()`. No install has
that — fresh or upgraded, the stored default is `now()` — so it is wrong today regardless of which way
this finding lands. One line, folded into the same diff.

The alternative — keep the prefix and add the `ALTER`s — is the repointing you just deleted, and it
brings back the producer-blocking `ALTER TABLE job … SET DEFAULT` (40.6 ms, `send()` blocked for the
whole transaction) to change a value that is already correct everywhere.

Worth saying what the prefix was buying, since dropping it is not free either. Once `${schema}.now()`
exists, a bare `now()` in DDL parsed by a session whose `search_path` names `pg_catalog` *after* the
pg-boss schema binds to the schema function instead:

```
search_path = amb, public              (pg_catalog implicitly first)  default now() -> bound to pg_catalog
search_path = amb, pg_catalog, public  (pg_catalog explicitly last)   default now() -> bound to amb.now()
```

and drift is inverted there, since `pg_get_expr` qualifies whichever name is shadowed — the
wrongly-bound default renders bare `now()` and matches the manifest, while the correctly-bound one
renders `pg_catalog.now()` and is what gets flagged. Narrow, since the common
`search_path = pgboss, public` leaves `pg_catalog` implicitly first, and an upgraded install can't
have the protection anyway without the ALTER. If it's worth closing, it's worth closing for everyone
in its own change rather than as a side effect here.

### 3. (minor) The v42 function text doesn't match the one `plans.ts` installs

Same shape as the one above, one object over — except the catalog can see this one.
`plans.createClockFunction` emits a multi-line body; v42's `install` hardcodes a one-liner. Postgres
stores `prosrc` verbatim, so the two paths leave different function source behind:

```
migrated from v41   prosrc: " SELECT pg_catalog.now(); "
fresh v42 install   prosrc: "\n      SELECT pg_catalog.now();\n    "
prosrc identical?   false      normalized identical? true
```

Drift passes because `normalizeFunctionBody` collapses whitespace — right call for a body diff — but
`pg_get_functiondef`, and so `pg_dump`, differ between two installs that are meant to be the same
schema.

Simplest fix is to paste the manifest's own entry in as the frozen literal. `schema.json`'s
`functions.now.def` is the rendered definition of a fresh install, so copying it into the v42 `install`
with the schema token substituted makes the two paths byte-identical by construction rather than by
matching whitespace by eye:

```ts
`CREATE OR REPLACE FUNCTION ${schema}.now()
 RETURNS timestamp with time zone
 LANGUAGE sql
 STABLE
AS $function$
      SELECT pg_catalog.now();
    $function$
`
```

Pasted, not referenced — the migration shouldn't read `schema.json` (or call
`plans.createClockFunction()`) at runtime, since the manifest is regenerated every schema version and
that would let a later body change rewrite what the *v42* step installs, which `migrate(42)` and
`getMigrationPlans(schema, 42)` can stop on. The one semantic difference is `CREATE OR REPLACE` in
place of `CREATE`, which is just how Postgres renders a definition; at v42 nothing owns the name yet.

And `schema.json` belongs in a test for the same invariant. It is generated from a fresh install, so it is the third
party both paths have to match, and its `def` carries the body verbatim, with no normalizer in the
way:

```ts
const manifestNow = schemaManifest.partitioned.functions.find(fn => fn.name === 'now')
const freshSource = await clockSource()                       // prosrc, raw
expect(freshSource).toBe(extractFunctionBody(manifestNow.def))
…
await contractor.migrate(currentSchemaVersion - 1)
expect(await clockSource()).toBe(freshSource)                 // migrated == fresh, byte for byte
```

Dropped into the existing v42 migration test it fails on the current head with the divergence spelled
out —

```
AssertionError: expected ' SELECT pg_catalog.now(); ' to be '\n      SELECT pg_catalog.now();\n    '
```

— and passes with the pasted definition. Both together: 2 files, +29/-1, suite **1230/1230**, eslint
and the manifest check clean.

bhamiltoncx:
Updated:

> 1. (minor) A TestClock on a pooled custom db adapter now silently runs on real time

Fixed. `start()` now refuses a `TestClock` unless the adapter is pg-boss's own `Db` or declares `clockSessionSetup()` to manage attaching/detaching itself. `enableClockOverride` and `disableClockOverride` are now exported along with `CLOCK_OVERRIDE_SETTING` to allow pooled custom db adapters to use `TestClock`.

> 2. plans.ts changes 11 column defaults, and nothing in v42 matches them

Reverted change, added lint guard.

> 3. (minor) The v42 function text doesn't match the one plans.ts installs

v42 now installs the manifest's own rendering of the function, pasted as a literal.

timgit:
Thanks for the update.  I ran another review, and now Claude is suggesting to revert the change just requested about the pg_catalog.now() in the default DDL because of shadowing that will start happening because of the function being named `now()` with all of the fun that is search paths.  Instead, I think it would be easier going forward to stop shadowing entirely, and just name the new function `job_now()`.

bhamiltoncx:
OK, working on the rename now.

bhamiltoncx:
Renamed new function to `job_now()`. `eslint` will catch bare `now()` and tell callers to use `job_now()` instead.

docs/api/testing.md
```diff
@@ -104,6 +104,54 @@ afterEach(() => {
...
+
+Only sessions that set `pgboss.test_clock = 'on'` see the fake time. pg-boss's own pool sets it on every connection when a `TestClock` is configured, so an instance without one on the same schema keeps real time, as does anything started after a killed run left the override in place. A custom `db` adapter has to run the statement `enableClockOverride()` returns on every connection it opens and declare `clockSessionSetup: true`; `start()` refuses a `TestClock` on an adapter that has not, since a pooled adapter would otherwise read fake time on some connections and real time on others. `attach()` issues the `SET` once itself, which is all a single-session adapter such as PGlite needs, and the bundled PGlite adapter declares the flag. Releasing the clock issues `disableClockOverride()` on the same session.
```
timgit: paragraph is hard to read.  reword

src/adapters/pglite.ts
```diff
@@ -38,6 +38,8 @@ export function fromPglite (pglite: PGliteLike): IDatabase {
   }
 
   const db: IDatabase = {
+    // One session, so the SET attach() issues reaches everything.
+    clockSessionSetup: true,
```
timgit: clean this up.  seems a bit too on the nose for the core pglite adapter

docs/api/testing.md
```diff
@@ -104,6 +104,54 @@ afterEach(() => {
...
+import { PgBoss, TestClock } from 'pg-boss'
+
+const clock = new TestClock('2026-01-01T00:00:00Z')
+const boss = new PgBoss({ connectionString, clock, __test__enableSpies: true })
+await boss.start()
+
+await boss.work('q', async () => {})
+const id = await boss.send('q', {}, { startAfter: 60 })
+
+await clock.tick(60_000)   // Date, Postgres now(), and the worker's poll all move together
```
timgit:
```suggestion
await clock.tick(60_000)   // JS dates, Postgres timestamps, and the worker's poll all move together
```

src/clock.ts
```diff
@@ -0,0 +1,191 @@
...
+ * through ${schema}.job_now(), so JavaScript and Postgres agree on what "now" is.
+ */
+export class TestClock implements AttachableClock {
+  #now: number
+  #timers: Timer[] = []
+  #seq = 0
+  #ticking = false
+  // Every (db, schema) this clock currently pushes time to. More than one because a multi-instance
+  // test shares one clock across several PgBoss instances, each of which attaches on start(); the
+  // schema's now() is restored only when the last of them disposes its handle.
```
timgit:
```suggestion
  
```

timgit: this is no longer true

src/clock.ts
```diff
@@ -0,0 +1,191 @@
...
+    } finally {
+      this.#ticking = false
+    }
+  }
+
+  async attach (target: Target): Promise<AsyncDisposable> {
+    const { db, schema } = target
+
+    await db.executeSql(`
+      CREATE TABLE IF NOT EXISTS ${schema}.clock (now timestamp with time zone NOT NULL);
```
timgit: replace IF NOT EXISTS?  Could we stomp an existing table?

test/migrationTest.ts
```diff
@@ -1319,4 +1322,53 @@ describe('migration', function () {
       }
     })
   })
+
+  it('v42 adds the schema clock function and leaves the timestamp defaults on pg_catalog', async function () {
```
timgit: do we need to add this test?

eslint.config.js
```diff
@@ -8,8 +8,27 @@ const config = neostandard({
   noJsx: true,
 })
 
+// Ensure Postgres timestamp API calls in SQL statements go through
+// ${schema}.job_now() so a test clock can redirect them (pg-boss #689).
+const SQL_CLOCK_CALLS = String.raw`(^|[^.\w])(now|transaction_timestamp|statement_timestamp|clock_timestamp|timeofday|age)\(`
+const SQL_CLOCK_KEYWORDS = String.raw`\b(CURRENT_DATE|CURRENT_TIME|CURRENT_TIMESTAMP|LOCALTIME|LOCALTIMESTAMP)\b`
+const SQL_CLOCK_INPUT_STRINGS = String.raw`'(now|today|tomorrow|yesterday)'`
+const SQL_CLOCK_MESSAGE = 'SQL must read the clock through the schema-owned job_now() function (pg-boss #689)'
```
timgit:
```suggestion
const SQL_CLOCK_MESSAGE = 'SQL must read the clock through the schema-owned job_now() function (pg-boss #901)'
```

bhamiltoncx:
Rock on, thanks much!

timgit:
You're welcome, and thanks for the PR.  I plan on publishing this in a few hours.  Running some post-merge items through another pass.

PR #902, opened 2026-09-11
fix: close the connection pool on a stop after stop({ close: false })

### LINKED ISSUE #576: Two phase shutdown process
Our production application has 3 main _entrypoints_:

1. HTTP Routes (Koa)
2. Jobs/Crons (pg-boss)
3. Events (event-emitter2)

All within the same process in a Docker container.

When shutting down the application gracefully, we devised three steps:

1. Stop route handlers and job workers and drain them
2. Drain event listeners
3. Close connections to application DB, Redis, Sentry

Route handlers, job workers and event listeners may want to publish jobs or dispatch events at any time. As the event bus is in-memory, it needs to be available until all handlers/workers are drained.

Currently we use pg-boss `stop({ wait: true, close: true, graceful: true })`

This causes issues if, on step 2, an event listener wants to publish a job. Then, the pg-boss DB connection is already closed.

## Proposed solution

Expose a `drain(timeout)` method that stops workers (internally `offWork()` + wait with timeout) but keeps the DB open so new jobs may be published.

Allow `stop()` to stop the DB connection even if the worker is already stopped.

## Considered alternatives

We are going with alternative 3 for now, but would love to have your thoughts on this workflow!

### 1. Calling `stop` 2 times

```typescript
// Drain
    await this.boss.stop({
      close: false, // Keep DB connection open so we can still add jobs
      wait: true,
      graceful: true,
      timeout: timeoutMs,
    });
```

```typescript
// Close connections
    await this.boss.stop({
      close: true,
      wait: true,
      graceful: true,
      timeout: timeoutMs,
    });
```

This does not work as `stop` checks if the `this.stopped` flag is set and short-circuits

### 2. Stopping via `getDb().stop()`

```typescript
// Drain
    await this.boss.stop({
      close: false, // Keep DB connection open so we can still add jobs
      wait: true,
      graceful: true,
      timeout: timeoutMs,
    });
```

```typescript
await this.boss.getDb().stop();
```

The issue here is that the Db interface should not expose `stop()` as it is an implementation detail. We could patch it on our side but seems shady.

### 3. Implement custom Db

Copy the pg-boss implementation of `Db` to our code and manage the pool manually

### 4. `offWork()` + `stop()`

Call `offWork()` with each queue name (step 1) and then `stop()` after event listeners drain (step 3)

New jobs won't be processed during shutdown, but does not wait for current jobs to drain. This means these jobs may depend on the event listener which is shutdown.

### 5. Custom worker tracking

We could track running jobs on our side and offWork/wait for them.

toruiwasa:
## The problem

`stop({ close: false })` is the documented way to stop workers while keeping the pool open, so
`send()` and `fetch()` still work after the workers are gone. It also marks the instance stopped, so
every later `stop()` returns at the `#stopped` guard and the pool pg-boss opened is never closed.
`getDb()` returns `IDatabase`, which has no `close()`, so there is no supported way out.

That leaves `close: false` callers with the problem #318 fixed for everyone else: sockets stay open
for pg-pool's 10s idle timeout, so shutdown drags and test runners report leaked handles.

```js
await boss.start()
await boss.createQueue('q')
await boss.stop({ close: false })
await boss.send('q')            // the point of close: false
await boss.stop()               // asks to close
boss.getDb().pool.totalCount    // 2
```

## The change

`stop()` closes the pool when it is asked to close and one is still open, ahead of the `#stopped`
guard. The guard itself is untouched, and so is the first `stop()`: the new branch needs the
instance to be stopped already. The close is published as `#stoppingPromise` like the rest of the
teardown, so a concurrent `stop()` or `start()` waits for the pool to drain, and it shares
`#closeDb()` with the normal shutdown.

`#doStop()` does not run a second time, so `stopped` is not emitted again. It reports workers and
maintenance finishing, which happened on the first call. This is the part you may disagree with: if
the second call should emit it too, I will move the close into `#doStop()` instead.

The condition checks `opened` as well as `close`, because an instance that was never started is
already marked stopped, and `Db.close()` reads a pool that does not exist yet.

`graceful` and `timeout` are not used on the second call, as they are not today either: there is
nothing left to wait for.

This changes existing behavior in one case: a `stop()` after `stop({ close: false })` used to do
nothing, and now closes the pool. That is what the `close` default is documented to do, but
the no-op is what actually ran, so code that kept calling `send()` after a second `stop()`
would now get `Database not opened`. Saying so in case it matters for how this is versioned.

## Tests

The tests that end at the existing `#stopped` guard also count `stopped` events, so a second
`#doStop()` would show up as a second emit. Nothing asserted that before: with the guard removed,
every test around `stop()` still passed.

The second commit is separate on purpose. Nothing asserted the `Database not opened...` path, and
the `work()` rejection test never checked which message came back, so a database error replacing the
worker error would have gone unnoticed. That is worth having, but it is not needed for this fix, so
it is easy to drop if you would rather keep this PR to the change itself.

## Verification

On top of 6de0d90.

| leg | result |
|---|---|
| `npm run cover` | 1196 total, 1 to 5 failed per run over six runs |
| `npm run test:distributed` | 1196 total, 6 skipped, 1 to 2 failed per run over four runs |
| `npm run test:pglite` | 1031 passed, 165 skipped, 0 failed (1196) |

The failure that repeats is [`queueStatsHistoryTest > should filter by from/to range`](https://github.com/timgit/pg-boss/blob/6de0d9082d7650d348a3e6f07bc39ed6ddedf5d4/test/queueStatsHistoryTest.ts#L140-L151),
which compares a host `new Date()` with a `captured_on` the database stamped, and my database VM
runs about 70ms ahead of the host. It failed the same way before this change. The others are
time-based tests that change from run to run (`delayTest`, `deleteTest`, `expireTest`,
`failureTest`, `flowTest`, `monitoringTest`, `monitorVacuumTest`, `retryTest`, `speedTest`,
`updateTest`). They pass in isolation, and running those files three times each with the change
stashed and three times with it gave 4 failures without the change and 1 with it.

## Not in this PR

One neighboring path is unchanged. `stop()` called while another `stop()` is in flight returns the
in-flight promise without reading the new options, so `stop({ close: true })` after an un-awaited
`stop({ close: false })` leaves the pool open, before and after this change. What the second call
should do there, wait and re-evaluate or refuse conflicting options, is a question of its own.

No `drain()`, following your reply on the issue.

Closes #576


timgit:
Thanks for opening.  Pasting a review below.

-------------


This is the right shape, and the writeup makes it easy to check. Verified on Postgres 18: full suite 1197/1197, `DISTRIBUTED=true` 1191 passed / 6 skipped, and the PGlite leg green. `queueStatsHistoryTest > should filter by from/to range` passes here, so that one is the VM clock described in the PR, not this change.

The new tests are real regression tests — reverting only `src/index.ts` back to `6de0d90` and keeping them fails exactly the three expected.

The full #576 workflow also works end to end on this branch. With a worker whose handler is still running when shutdown starts:

```
before phase 1: handler start
phase 1 done after 501 ms: handler start | handler finish
phase 2 send: ok
phase 2 job states: completed,created
phase 3 totalCount: 0
phase 3 send after close: Database not opened. Call open() before executing SQL.
```

Phase 1 drains the in-flight job to `completed` rather than failing it through `failWip()`, phase 2 can still publish, and phase 3 is the part that only works with this PR.

On the `stopped` question: the choice made here is the correct one. `stopped` reports workers and maintenance shutting down, and that already happened on the first `stop()`. It should stay out of `#doStop()`.

One thing to fix before merge.

### The close needs to go through `#stoppingPromise`

The new branch is the only teardown path in `stop()` that does real work without publishing `#stoppingPromise`, so a concurrent `stop()` falls through to the `#stopped` guard and resolves while `pool.end()` is still draining:

```js
await boss.stop({ close: false })
boss.getDb().executeSql('select pg_sleep(0.5)')   // keep one connection busy

const first = boss.stop()
const second = boss.stop()

await second
// pool.ending: true, totalCount: 1   <-- socket still open
await first
// pool.ending: true, totalCount: 0
```

That hands a caller who awaited `stop()` the same open socket this PR exists to close. It also means a concurrent `start()` does not wait for the close, since `start()`'s first guard is `if (this.#stoppingPromise) await …`.

Publishing the promise fixes both, and it is a good excuse to stop repeating the close and the `delay(10)`:

```js
    let { close = true, graceful = true, timeout = 30000 } = options

    if (this.#stopped && close && this.#db._pgbdb && this.#db.opened) {
      this.#stoppingPromise = this.#closeDb()

      try {
        return await this.#stoppingPromise
      } finally {
        this.#stoppingPromise = null
      }
    }
```

```js
  async #closeDb (): Promise<void> {
    if (!this.#db._pgbdb || !this.#db.opened) return

    await this.#db.close()

    // Give event loop time to process socket closes
    await delay(10)
  }
```

with `#doStop()`'s `shutdown()` calling `#closeDb()` too. Hoisting the destructure above both guards also drops the second spelling of the `close` default — `(options.close ?? true)` and `close = true` six lines apart have to be changed together otherwise. Reading the options earlier costs nothing; `Math.max(timeout, 1000)` and `#stoppingOn` stay where they are.

This version was run against the probe above: the second stop now reports `totalCount: 0`, the suite stays at 1197/1197, and types and lint are clean.

### Smaller, non-blocking

- `docs/api/ops.md`: the sample is right, but the `close` bullet above it still reads as if `close: false` is terminal. Worth a sentence saying a later `stop()` closes the pool and that `send()`/`fetch()` reject with `Database not opened…` afterward. That is the behavior change the PR description calls out, and it is the only thing an upgrading `close: false` user can trip over. Same for the `stopped` event docs, since it now fires with the pool open and never again when it closes.
- `should not leave open handles…` counts every handle in the worker, not pg-boss's. It does fail without the fix, so it earns a place, but `pool.totalCount === 0` two tests above proves the same thing deterministically and will not start flaking when another file's timers land in the same fork.
- `should emit stopped once when a close: false stop is followed by a stop that closes` is contained in `should do nothing when stop() is called again after the pool was closed` — same calls, same assertion, one fewer `stop()`. Dropping it costs no coverage.

Keep the second commit. The `Database not opened` path being unasserted is exactly the kind of gap that lets a database error quietly replace a worker error.

Leaving the in-flight `stop()` case out is the right call — that one needs its own decision about whether the second caller waits and re-evaluates or refuses conflicting options, and it should not hold this up.

toruiwasa:
Thanks for the review, and for running the workflow end to end. The changes are pushed, details below.

-------------

**The close goes through `#stoppingPromise`**

That one I had missed: I had checked that a second `close()` was harmless, not that a second caller could get back before the pool drained. Took the shape you gave: the destructure sits above both guards, the close branch publishes `#stoppingPromise`, and `#closeDb()` is shared with `shutdown()`.

Added a regression test for the concurrent case. One note on the probe: with pg-pool 3.14 the query only takes a client on the next tick, so calling `stop()` right after `executeSql()` ended an all-idle pool here and the second `stop()` reported `totalCount: 0` even without the fix. The test waits for the pool's `acquire` event before stopping. With `src/index.ts` reverted it fails with `expected 1 to be +0`.

**Docs and tests**

Both doc sentences are in (`close` option and `stopped` event). The two tests are dropped, the open handles one as well, since the `totalCount` test shows the same thing. If you would rather keep it, I can add it back.

**Verification**

tsc and lint clean. The suite is at 1196 (two dropped, one added). PGlite: 1031 passed, 0 failed. The Postgres and distributed legs did not come out clean in a single run on this machine: between one and five time-based tests failed each run, different ones each time, plus the `queueStatsHistory` clock one. All of them pass in isolation, and running the failing files three times with the change stashed and three times with it gave 4 failures without the change and 1 with it, so I am reading that as host load rather than this change.

PR #905, opened 2026-09-17
fix: take the interval claims a second early so a tick that lands short does not skip a pass

toruiwasa:
## Summary

Each interval claim takes its row, the one on `version` or a queue's own on `queue`, when the row is more than the interval old. This moves that threshold one second below the interval, capped at a tenth of it so the shortest intervals keep a threshold above zero, through one function all six claims (cron, bam, flow, reindex, maintain, monitor) share.

The reason is that the same `seconds` sets both the tick and the threshold it is measured against, on two different clocks:

```js
// src/timekeeper.ts: the tick
this.config.clock.setInterval(async () => await this.onCron(), this.config.cronMonitorIntervalSeconds! * 1000)
```

```sql
-- src/plans.ts, trySetTimestamp: the claim the tick makes, before and after
WHERE EXTRACT( EPOCH FROM (${schema}.job_now() - ...) ) > ${seconds}
WHERE EXTRACT( EPOCH FROM (${schema}.job_now() - ...) ) > ${claimThreshold(seconds)}
```

```ts
const CLAIM_TOLERANCE_SECONDS = 1

function claimThreshold (seconds: number): number {
  return seconds - Math.min(CLAIM_TOLERANCE_SECONDS, seconds / 10)
}
```

(`...` is the `COALESCE` that lets the first claim through when the column is still null; it is unchanged.)

The tick runs on the client's monotonic clock; the threshold is the age of the row on the server's wall clock, to the microsecond. A tick that lands a few milliseconds short by the server's clock loses the claim outright, and the next tick claims a whole interval later. For the cron pass at the default interval that is a 30-second delay. At 31 to 45 seconds it is worse: the gap between passes is twice the interval, more than the 60-second window a pass sends, and the occurrences that fall outside it are not sent. 45 is the documented ceiling, and the scheduling docs say a deployment whose passes keep running has no gap.

## Why the ticks land short

Two things put the two clocks a few milliseconds apart.

Node schedules the next interval from the time it took just before running the callback, floored to a millisecond, and any delay before the callback runs shortens the next interval further. Out of 300 `setInterval(1000)` callbacks on this machine, 10 came less than a second after the one before.

The two clocks also do not run at quite the same rate. During the 45-second run below the wall clock ran 18 parts per million slower than the monotonic clock, 0.8 ms per interval, so 4 of the 7 refused ticks came at least 45 seconds after the previous one by the timer's clock and still found the row 44.999 seconds old.

`TestClock` shows the boundary without any of this: it fires each timer exactly one period after the last and moves the database clock with it, so every other `tick(N * 1000)` is refused with the row exactly `N` seconds old.

## Where the shape comes from

The claim took this shape in v10 (4b3d9f4, #425), when the debounce that re-queued a cron check after every job gave way to `setInterval` and an `UPDATE` that only goes through when the row is old enough, and the docs for `missed` (#898) were written with the interval as the spacing between passes. The other claims (`bam_on`, `flow_on`, the queue monitor) have the same shape, but nothing behind them is windowed, so they only run late.

## What it changes

A claim 29 seconds or less after the previous one (at the default) is still refused. With several instances, the claim goes to whichever tick arrives first once the row is old enough, which in practice is the same instance every interval; a tick from another instance that lands in the last second before it can now take the claim instead. Passes stay one interval apart, the one that moves the claim landing up to a second early, and a slot is never sent twice. The one visible effect of the claim moving is an extra row in `queue_stats` from the monitor.

A second covers the timer and the clock rate with room for a claim that waited a few hundred milliseconds for a pool connection. A tick more than a second late still loses, as it does today. If you would rather have the smaller number, 100 ms covers the timer and the clocks alone; the constant and the helper in `claimTest` that pins the same shape are the two places to change.

No schema change. During a rolling upgrade both releases run the same statement with different thresholds, the older one refuses a claim the newer one takes, and the claim moves to a newer instance within an interval or so. Rolling back is the same in reverse.

## Tests

`test/claimTest.ts` covers each claim on both sides of its threshold at several intervals, exactly one of several concurrent claims taken, a pass on a short tick with the time of the previous pass handed to it, and the previous release's statements running against the same rows in both directions. On a `TestClock`, a claim exactly one interval after the previous one is taken and one exactly at the threshold is refused, and an every-minute schedule at the 45-second ceiling sends every occurrence across four ticks, where the strict comparison sends two.

## Verification

Measured on this machine (macOS arm64, Postgres 18). A single instance on 6f8c8e1, whose claim statements are the ones on master today, running for 15 minutes with `* * * * *` and an RRULE at `:30` scheduled (a six-field cron was scheduled as well; its misses come from a separate slot issue and are left out):

| `cronMonitorIntervalSeconds` | claims refused | occurrences not sent |
|---|---|---|
| 45, before | 7 of 21 (row 44.994 to 44.999 s old) | 4 of 14 every-minute, 2 of 15 RRULE |
| 30, before | 9 of 31 | none, the affected ones 30 s late |
| 45, after | 0 of 21 (shortest row 44.990 s old) | none |

Two earlier 15-minute runs at 45 on 12.31.x lost 3 of 15 and 2 of 15. I have not measured the timer or the clock rates on Linux.

The suites, on top of f4233b2:

| leg | result |
|---|---|
| `npm run test:pglite` | 1157 passed, 212 skipped, 0 failed |
| `npm test` | 1369 passed, 0 failed |
| `npm run test:distributed` | 1363 passed, 6 skipped, 0 failed |

## Not in this PR

A gap of the same size opens when the instance holding the claim stops: the next instance's tick is anywhere up to an interval after the last pass, so at 45 the first pass after a deploy can be up to 90 seconds after the one before it. The tolerance does not cover that. It has its own fix, a single retry when the row is due, which I would put up separately.


timgit:
Thanks for opening.  I've reviewed, but triage on this indicated a pattern mistake instead of just a timing tolerance issue.  Basically, I'll abandon setInterval for setTimeout going forward for all of the important timestamp claims.  That shouldn't be on you, so I'm opening a new PR with a fix soon.

toruiwasa:
Thanks for the review, understood. Next time I'll consider opening an issue first for something in this area.

timgit:
No worries!  Here's the PR: #906

toruiwasa:
Ran my 15-minute reproduction against master with #906 and the lost claims are gone. Thanks a lot!

Issue #786, opened 2026-05-16
Feature: Support for URL base path (prefix) configuration in dashboard


company-myther:
## Background
The dashboard currently does not support deployment under a custom URL path prefix (base path). Attempting to proxy the dashboard at a subpath (such as `/dashboard`) results in broken static asset references (JS/CSS 404), since all routes and resource URLs are assumed to be rooted at `/`.

## Use Case
Many production environments require the dashboard (or any SPA) to be served behind a reverse proxy at a path prefix (e.g. `https://mydomain.com/pgboss-dashboard/`) for:
- Multi-app shared domains
- API Gateway routing
- Security or organizational requirements

However, currently all resource links and routing are absolute. Even with reverse proxy, the app breaks because of missing assets.

## Feature Request
Please add official support for configuring a URL base path (e.g., `BASEPATH` or `DASHBOARD_BASE_PATH` environment variable) for the dashboard front-end and back-end, so that:
- All asset/script/style URLs and API endpoints are generated with the prefix
- React Router and Hono server both honor the base path for all navigation
- The dashboard can be safely reverse proxied at a subpath

### Example
- Environment: `DASHBOARD_BASE_PATH=/pgboss-dashboard`
- Expect: All routes and assets served under `/pgboss-dashboard` (not `/`)

## References
- Related issues in other React/Remix/SPA projects: [React Router basename](https://reactrouter.com/en/main/routers/create-browser-router#basename), Remix base path discussions
- Many enterprise/self-hosted dashboards support this (Prometheus, Grafana etc.)

## Current workaround
Forking and patching asset references or using complex Nginx rewrites, which is brittle and not robust to upgrades.

## Impact
This feature would greatly improve dashboard deployability and flexibility for users with more complex infrastructure requirements.

Thanks!

TomasTokaMrazek:
@thyhjwb6 Does this really close the issue? I don't think that changing the base path through GH clone repo and manual build satisfies the `greatly improve dashboard deployability and flexibility for users`. The `PGBOSS_DASHBOARD_BASE_PATH` should be runtime variable.

Issue #418, opened 2023-08-21
High Database CPU when processing backlog


fenos:
Hello, firstly let me take a second to give my appreciation to pg-boss, great library!

We are noticing an increase in CPU consumption on the Database when a backlog of jobs is building up.
The chart looks like this:

<img width="513" alt="image" src="https://github.com/timgit/pg-boss/assets/4754064/b610c33b-a9e6-4383-835d-338a091d160b">

From the investigation, it seems that the workers pulling the jobs are putting too much strain on the Database CPU when it can't keep up.

You can see from the graph below that the CPU was quite stable, however during the raise a producer started queuing many more messages. The enqueue operation doesn't seem correlated to CPU usage. 
Subsequetially the Drop in CPU was caused by canceling all the jobs via a SQL query

<img width="2426" alt="image" src="https://github.com/timgit/pg-boss/assets/4754064/85e200d5-f03c-412f-87a6-cef854535a9b">

The workers are set up as follows:

```ts
 new PgBoss({
      connectionString: url,
      max: 4,
      application_name: 'pgboss',
      deleteAfterDays: 7,
      archiveCompletedAfterSeconds: 14_400,
      retentionDays: 7,
      retryBackoff: true,
      retryLimit: 20,
      expireInHours: 48,
      monitorStateIntervalSeconds: 30,
     newJobCheckInterval: 700,
    })
```

on the job handler, we use async / await to build up back pressure. Just to clarify the handler doesn't issue any SQL statement to this database, so the load is purely from pg-boss

We are currently running 1 worker in 4 separate containers, so each container should pull 1 message at a time.

I'm unsure why pg-boss would incrementally use more Database CPU even if the concurrency is set to 1 and the next message shouldn't be fetched until the previous one completes.

Currently, we are ok with processing these messages slower than they get produced, however, we can't justify the CPU increase.

Do you know what might cause the CPU to spike up internally in pg-boss?

timgit:
How many records were created in the job table when the util was high?

fenos:
@timgit After further investigation it seems that the issue is caused by the sorting on the `nextJob`
https://github.com/timgit/pg-boss/blob/master/src/plans.js#L379

Since there is no compound index on these columns when the table gets big, Postgres would start doing a disk merge sorting, causing high CPU usage as the table gets bigger.

I'm wondering if adding an index on the sorted columns could help speeding up things and reduce the load on Postgres when sorting

fenos:
[...]

PR #908, opened 2026-09-19
feat(dashboard): set the base path at runtime

FR073N:
## Summary

`PGBOSS_DASHBOARD_BASE_PATH` now works when you start the dashboard, not only when you build it. That makes sub-path deployments possible with the published npm package, which ships prebuilt at `/`. This is what @TomasTokaMrazek asked for at the end of #786.

While testing it I hit a second problem in the same code and fixed it here: the documented `node node_modules/@pg-boss/dashboard/build/server.js` serves no static assets unless you start it from the package directory, because the asset root is relative to the working directory.

## How it works

I expected to need Vite's `renderBuiltUrl`, but the build turned out to be almost portable already. Chunks import each other relatively, and no dynamic import carries preload dependencies. The only absolute URLs are the ones in the React Router route manifest, which exists twice: on the server as `build.assets`, and in the browser as the static `assets/manifest-*.js`.

So `withBasePath()` returns a copy of the server build with a new `basename`, a new `publicPath`, and the manifest URLs moved under the prefix. The static manifest file is byte for byte `window.__reactRouterManifest=${JSON.stringify(build.assets)};`, so under a runtime base path the server answers that one URL from the re-homed build instead of reading and patching the file.

Because this relies on a property of the build output, `scripts/check-build-portable.mjs` now runs at the end of `npm run build`. It fails if a client chunk embeds an absolute `/assets/` URL or if a dynamic import gains preload dependencies. I checked that it fails on both by hand-editing a chunk.

## Changes

- `app/lib/runtime-base-path.ts`: `withBasePath()`, `rehomeAssetUrls()`, `renderManifestSource()`. The base path is validated, since it ends up inside a Hono route pattern.
- `app/server.ts`: new `basePath` and `clientRoot` options on `createHonoApp()`.
- `app/server.node.ts`: reads `PGBOSS_DASHBOARD_BASE_PATH` at runtime and resolves the client assets from `import.meta.url`.
- `vite.config.ts`: removes the `define` that inlined the variable into the build. Nothing in `app/` reads it.
- `scripts/check-build-portable.mjs`, wired into `npm run build`.
- Docs: the sub-path section and the configuration table.

## Behaviour for existing users

| Built with | Started with | Before | After |
|---|---|---|---|
| unset | unset | `/` | `/` |
| `/a` | `/a` | `/a` | `/a` |
| `/a` | unset | `/a` | `/a` |
| unset | `/a` | `/` | `/a` |
| `/a` | `/b` | `/a` | `/b` |

The last two rows are the change. The old docs told people to set the variable for both `build` and `start`, so I would not expect anyone to be relying on a start-time value being ignored.

## Testing

- `npm run typecheck`
- `npm run test:frontend`: 236 passed
- `npm run test:server`: 288 passed. New: `runtime-base-path.test.ts` for the pure functions, and `server-base-path.test.ts`, which drives `createHonoApp()` with a stub build and a temp directory. It checks that the manifest comes back re-homed, that it stays behind basic auth, and that other assets are served from `clientRoot`.
- `npx eslint` on the touched files

Then against a packed tarball installed with `npm i --omit=dev`, started with `cd / && node <install>/node_modules/@pg-boss/dashboard/build/server.js`:

```
## this branch, PGBOSS_DASHBOARD_BASE_PATH=/pgboss
   log: base path: /pgboss
   GET /pgboss/schedules -> 200
   GET /pgboss/assets/entry.client-Bmgthb4R.js -> 200
   GET / -> 404   | un-prefixed asset URLs in HTML: 0
   manifest: 159 re-homed URLs, 0 left at the root
   POST /pgboss/queues/create -> 302 location: http://localhost:8092/pgboss/queues/proof14971

## this branch, no base path
   GET /schedules -> 200
   GET /assets/entry.client-Bmgthb4R.js -> 200

## published 1.8.0, no base path, same command
   GET /schedules -> 200
   GET /assets/entry.client-Bmgthb4R.js -> 404
```

In Chrome, under `/admin/queues`: 27 requests, all 200. That covers client-side navigation between pages, the lazily loaded route chunks, the `.data` loader calls, and `/__manifest` route discovery.

## Notes

- The console shows a React hydration error (#418) on `/queues/:name`. It is there on the published 1.8.0 build too, so I left it alone.
- No version bump, following what other external PRs did.
- #730 (mounting the dashboard in an existing app) builds on this. It is in #910, kept as a draft until this one has had a look, so each stays small.


FR073N:
Folded into #910. The runtime base path and the mountable handler touch the same few files and neither is much use alone, so one PR is easier to review than two stacked ones. Nothing was dropped.

PR #909, opened 2026-09-19
fix(dashboard): bundle UI libraries into the server build to shrink the install

FR073N:
## Summary

Installing `@pg-boss/dashboard` downloads 94 MB, and about two thirds of that is UI code that never runs. The client bundle tree-shakes the UI libraries down to 888 KB. The server bundle leaves them as external imports, so they have to be installed whole: `lucide-react` is 44 MB for the 20 or so icons the app renders, and `@base-ui/react` is another 19 MB.

This bundles them into the server build and moves them to `devDependencies`. The install drops to 27 MB.

It also fixes a crash I found while testing. The published build does not start under pnpm with `hoist=false`:

```
Cannot find package 'serialize-error' imported from .../build/server/index.js
```

`pg-boss` is aliased to `../../src` and inlined, so its own dependencies (`serialize-error`, `cron-parser`, `rrule-temporal`) end up as bare imports of a package that does not declare them. Under npm they resolve because of hoisting. With this change they are bundled too.

## Changes

`vite.config.ts` gets this, for builds only:

```ts
ssr: command === 'build'
  ? { noExternal: true, external: SERVER_RUNTIME_PACKAGES }
  : undefined,
```

`SERVER_RUNTIME_PACKAGES` is `@react-router/node`, `isbot`, `pg`, `react`, `react-dom`, `react-router`. React and React Router have to stay a single shared copy, and `pg` loads optional native bindings.

I first tried the opposite, a list of UI packages to bundle. It worked, but a dependency added later, or one brought in by the Pro overlay, would quietly become an import nothing installs. Bundling by default fails the safe way round. After the build, the only bare imports left in `build/server/index.js` and `build/server.js` are the six above plus `hono` and `@hono/node-server`, and all of them are declared.

It is limited to builds because the dev server evaluates inlined modules as ESM, and that breaks on the CommonJS-only `use-sync-external-store` that `@base-ui/react` depends on. My first version applied it to dev as well and every page returned 500 with `module is not defined`. Neither CI nor the test suites would have caught that, since they do not boot the dev server.

`package.json` moves `@base-ui/react`, `class-variance-authority`, `clsx`, `lucide-react`, `tailwind-merge` and `uplot` to `devDependencies`. The lockfile diff is only the resulting `dev` flags. My npm version also wanted to add some unrelated `oxide-wasm32-wasi` entries and drop `libc` fields, and I took those back out.

## Testing

- `npm run typecheck`
- `npm run test:frontend`: 236 passed
- `npm run test:server`: 270 passed
- Dev server with `node_modules/.vite` cleared: `/`, `/queues/alpha`, `/schedules/new` and `/queues/alpha/metrics` return 200.
- Built with the Pro overlay fixture copied into `app/pro` and `PGBOSS_PRO=1`: the build passes with the same external imports.

Packed tarballs, installed into empty directories:

| | published 1.8.0 | this branch |
|---|---|---|
| `npm i --omit=dev` | 94 MB | 27 MB |
| `pnpm add` with `hoist=false` | 94 MB, crashes at boot | 27 MB, starts |
| tarball | 399 KB | 664 KB |
| `build/server/index.js` | 745 KB | 1897 KB |
| `build/client` | 888 KB | 888 KB |

Nine pages return 200 under both npm and strict pnpm: `/`, `/jobs`, `/queues`, `/queues/:name`, `/queues/:name/metrics`, `/schedules`, `/send`, `/warnings`, `/migrations`. Response sizes match master byte for byte on every route compared.

The built bundle contains no second copy of React or React Router. `uplot` was already imported on the server on master, so bundling it changes nothing at render time.

## Notes

- `pg-boss` is still in `dependencies`, but nothing imports it at runtime any more. Dropping it would bring the install from 27 MB to 17 MB (luxon, type-fest and a few others). I did not touch it, because you may want it there as a compatibility signal. It is a one-line change if you agree.
- Licences: the published package now carries code that npm used to install alongside its own licence file. Everything is MIT or ISC except `class-variance-authority`, which is Apache-2.0 and asks for a copy of the licence with redistributions. The minified client bundle already redistributes the same code today, so this is not new, but it is more visible now. A build-time `THIRD-PARTY-NOTICES` file would settle it. Tell me if you would like that in this PR.
- No version bump.


timgit:
Thanks!

PR #910, opened 2026-09-19
feat(dashboard): mount the dashboard in an existing app, under a runtime base path

### LINKED ISSUE #730: dashboard - run as middleware

FR073N:
## Summary

Closes #730, and answers the runtime base path request at the end of #786.

Adds `@pg-boss/dashboard/handler`. `createDashboardHandler()` returns a plain Fetch handler, `(request: Request) => Promise<Response>`, that you mount in your own server behind your own authentication:

```ts
import { createDashboardHandler } from '@pg-boss/dashboard/handler'

const dashboard = createDashboardHandler({
  databases: [{ url: process.env.DATABASE_URL, schema: 'pgboss' }],
  basePath: '/admin/queues',
})

app.use('/admin/*', requireAdmin)
app.all('/admin/queues/*', (c) => dashboard(c.req.raw))
```

This is the shape pg-bossman has: its `createDashboard({ client, basePath })` also returns a Fetch handler, which is what the author of #730 uses today. We ran three Hono backends on it the same way. It has not been released in a year and is pinned to pg-boss 11, so we moved to this dashboard, first as a separate process. That is what the numbers below compare against.

Mounting under a prefix needs a base path the prebuilt package can honour, so `PGBOSS_DASHBOARD_BASE_PATH` now also works at start time for the standalone server. I first opened that part on its own as #908 and folded it in here: the two touch the same few files and neither is much use without the other.

## What it costs

Same Linux container (`node:24-slim`), same database, packed tarball installed with `npm i --omit=dev`.

| | Standalone process | Mounted with the handler |
|---|---|---|
| Resident memory while nobody uses it | 312 MB | 0 (nothing is imported) |
| Host process before the first visit | n/a | 73 MB, a bare `node:http` server |
| First request | n/a | 731 ms, loads the build once |
| Server-rendered page, median of 40 | 8.7 ms | 8.5 ms |
| `.data` call used for in-app navigation, median of 40 | 3.1 ms | 4.0 ms |
| Postgres connections, idle / browsing / cap | 0 / 4 to 7 / 10 | same pool, same numbers |
| Connections 14 s after the last request | 0 | 0 |
| Host memory after browsing | n/a | 324 MB resident, about +27 MB of live heap |

The last row deserves a note. Once visited, the host keeps the pages V8 reserved, as the standalone server does from its first second. The gain is that nothing is paid until someone opens the dashboard, and that there is no second process to run, authenticate and collect logs from.

## How the runtime base path works

The build turned out to be almost portable already. The only absolute URLs are in the React Router route manifest, which exists twice: `build.assets` on the server, and the static `assets/manifest-*.js`. `withBasePath()` returns a copy of the build with a new `basename`, a new `publicPath` and the manifest URLs moved under the prefix, and the server answers the manifest URL from that copy. `scripts/check-build-portable.mjs` runs at the end of `npm run build` and fails if a chunk starts embedding an absolute asset URL or a dynamic import gains preload dependencies. I checked it fails on both.

Existing deployments are unaffected: a start-time value equal to the build-time one, or no start-time value, behaves as before.

## Two things I only found by running it behind nginx

`db.server.ts` registers `SIGTERM`/`SIGINT` handlers on import that close the pools and `process.exit(0)`. Right for the standalone server, but inside a host it ended our API in the middle of its own shutdown. The handler sets a flag on `globalThis` before the build is imported, `db.server.ts` skips the registration when it sees it, and the handler exposes `close()`. Standalone behaviour is unchanged and `db.test.ts` covers both.

React Router rejects a form submission whose `Origin` does not match the request URL. Behind a proxy the host sees `http://127.0.0.1:3000`, so pages rendered and every action returned 400. That one is documentation: pass the handler a `Request` carrying the public URL.

## Also fixed

The documented `node node_modules/@pg-boss/dashboard/build/server.js` serves no assets unless started from the package directory, because the asset root was relative to the working directory. On the published 1.8.0, from `/`: page 200, `assets/entry.client-*.js` 404. It is now resolved from the file.

The package gains an `exports` map for `./handler`. It also lists `./build/*` so that `import '@pg-boss/dashboard/build/server.js'` keeps working for anyone already embedding it that way.

## Testing

- `npm run typecheck`, which now also fails if the hand-written `types/handler.d.ts` drifts from the implementation (checked by breaking it)
- `npm run test:frontend`: 236 passed. `npm run test:server`: 301 passed, 5 new files or cases.
- `npx eslint` on the touched files
- From a packed tarball in a separate project, started from `/`: the Hono and Express examples in the docs as written, host auth returning 403 in front of the dashboard, pages and assets 200 under the prefix, a queue created through the mounted dashboard (302 to the prefixed URL), and recovery after a failed first load without restarting the process (500, then 200).
- Standalone with `PGBOSS_DASHBOARD_BASE_PATH=/pgboss`: pages and assets 200 under `/pgboss`, `/` 404, 159 manifest URLs re-homed and none left at the root.
- Chrome, under a prefix: 27 requests, all 200, including client-side navigation, lazily loaded chunks and `/__manifest` route discovery.
- SIGTERM on an embedded host: its own 1.5 s shutdown ran to the end, and the only listener left was its own.

## Notes

- The docs say plainly that an unguarded mount exposes job payloads and the send, retry, cancel and delete actions, and mention CSRF for hosts using cookie sessions.
- `pgboss_db` and `sidebar:state` are set with `path=/`, so when embedded they reach the whole host application. That predates this PR. I can scope them to the base path in a follow-up.
- A React hydration error (#418) shows on `/queues/:name`. It is on the published build too, so I left it alone.
- No version bump.


timgit:
Thanks for both PRs

PR #913, opened 2026-09-20
fix(bam): complete a re-attempt whose index is already valid

FR073N:
Hi Tim,

We hit this in production on the day we moved an app from pg-boss 11 to 12.33.2. Nothing broke, but the API logged one error per minute, forever:

```
relation "job_common_i7" already exists
```

## What happens

`bam.ts` already handles "the build succeeded but the row was never marked" (the comment above the heal probe describes it exactly): on a re-attempt it leaves a valid index alone and lets the command's `IF NOT EXISTS` no-op, then marks the row completed.

That relies on the command having `IF NOT EXISTS`. The two oldest async commands don't:

```
CREATE INDEX CONCURRENTLY job_i7 ...          (group_concurency_index)
CREATE UNIQUE INDEX CONCURRENTLY job_i8 ...
```

Every later one (`job_i9`, `job_i5`, `job_i10`, `job_i11`) has it. So for these two, the re-attempt runs `CREATE INDEX` against its own valid index, fails, and a failed row is retried on every `bamIntervalSeconds`.

How we got there: our deploy runs a tiny script that starts pg-boss, registers the schedules and stops, all within a second. On the first boot after the upgrade it picked up `group_concurency_index`, built the index, and was stopped before `markCompleted` (`if (this.#stopped) return`). Five minutes later the API reclaimed the row and started failing. I reproduced it locally: a v11 schema, then `start()` immediately followed by `stop()` on 12.33.2 leaves the index built and the row `in_progress`.

## The change

- `bam.ts`: on a re-attempt, when the probe finds the index and it is valid, skip the command and mark the row completed. This is what fixes databases that already hold the old command text, since a `bam` row keeps the text it was queued with.
- `migrationStore.ts`: `IF NOT EXISTS` on `job_i7` and `job_i8`, like the others, for upgrades that have not happened yet and for the backends that do not take the probe path.
- `bamTest.ts`: the existing "valid leftover index" test, with a command that has no `IF NOT EXISTS`. It times out on master and passes here. `bamTest` and `migrationTest` pass (74 tests).

For anyone landing here with the same log line, the manual repair is safe once you have checked the index exists and is valid:

```sql
UPDATE pgboss.bam SET status = 'completed', completed_on = now(), error = NULL
WHERE name = 'group_concurency_index' AND status <> 'completed';
```

On our side we also stopped that short-lived script from running migrations (`migrate: false`), which is probably what it should have been doing all along.


timgit:
thanks!

PR #918, opened 2026-09-22
fix(vacuum): read lastVacuum as epoch millis, not timestamptz

### LINKED ISSUE #917: Vacuum monitor throws when the pool has a custom timestamptz type parser
In 12.30.0 (#885), `#checkVacuum()` started using `new Date(row.lastVacuum)` on a `timestamptz` column fetched from `pg_stat_user_tables`.

Because `pg-boss` uses the app's DB pool, any custom `timestamptz` parser can return non-`Date` values, like `Temporal.Instant`. `Temporal.Instant` intentionally throws an exception when passed to `new Date()`:

```
  pg-boss error: TypeError: Do not use Temporal.Instant.prototype.valueOf; use Temporal.Instant.prototype.compare for comparison.
      at Instant.valueOf (<anonymous>)
      at new Date (<anonymous>)
      at #checkVacuum (node_modules/pg-boss/dist/boss.js:481:49)
      at async #supervisePass (node_modules/pg-boss/dist/boss.js:268:9)
      at async #onSupervise (node_modules/pg-boss/dist/boss.js:187:32)
```

## Repro

pg-boss 12.33.5, pg 8.21.0, @js-temporal/polyfill 0.5.x, `types.setTypeParser(builtins.TIMESTAMPTZ, raw => Temporal.Instant.from(...))`, then start pg-boss with monitoring enabled.

Note: Other `timestamptz` reads coerced with `new Date()` will have the same problem:

* manager.ts #2446 (queue stats capturedOn)
* timekeeper.ts toTime()

bhamiltoncx:
Since 12.30.0 the vacuum monitor does `new Date(row.lastVacuum)` on a `timestamptz` column from `pg_stat_user_tables`. pg-boss runs on the application's pool, so an app-level `pg-types` parser that returns something other than a `Date` (`Temporal.Instant` in our case, whose `valueOf` throws by design) fails every supervise pass:

```
pg-boss error: TypeError: Do not use Temporal.Instant.prototype.valueOf; use Temporal.Instant.prototype.compare for comparison.
    at Instant.valueOf (<anonymous>)
    at new Date (<anonymous>)
    at #checkVacuum (node_modules/pg-boss/dist/boss.js:481:49)
    at async #supervisePass (node_modules/pg-boss/dist/boss.js:268:9)
```

The query now returns `lastVacuum` as epoch milliseconds computed in SQL, and `#checkVacuum` reads the number directly. The code only ever used the value as a number.

Adds a test that installs a throwing `timestamptz` parser on the pool and checks a supervise pass still completes and raises the xmin warning. It reproduced the stack above before the change.

`capturedOn` in `manager.ts` has the same coercion and is fixed separately in #919 .

Fixes #917.

I wrote this PR with the help of Claude Code.


bhamiltoncx:
Filed #920 on the flaky `bamTest` (passed 3/3 times locally).

timgit:
Thanks for the PRs!

bhamiltoncx:
Glad to help! Thanks for keeping the project rocking and rolling.

